Issue #10, opened 2018-07-16
Implement type casts


rainman110:
We need to support function astype 

Required at
 - https://github.com/NeuralEnsemble/elephant/blob/ed8ad8bf477e080c1b79c20a206b65131b85ffa0/elephant/asset.py#L280

rainman110:
Closed by #42

Issue #12, opened 2018-07-16
Add data type support in tensor data structure


rainman110:

rainman110:
Closed by #42

PR #41, opened 2018-07-31
Features/types

Markus-Goetz:
Realized HeAT's type system. Enabled type casting. Major refactoring of existing code to utilize new type system.

Resolves issues #10 and #12

Detailed overview of the changes:

* setup.py requires torch and numpy, h5py is optional dependency
* added io module
    - contains optional hdf5 loading function
    - enhanced hdf5 loading function to support arbitrary dimensions
    - added documentation
    - added tests
* added random module
    - copy over of the existing global seed initialization and uniform sampling
    - test cases and documentation missing
* enhanced communicator module
    - introduced abstract Communicator base class
    - communicators must now always be instantiated for a tensor
    - each communicator now has a chunk method, i.e. dividing a shape into slices according to the communication strategy and split axis, previously part of the load_h5 method
    - added tests
    - added documentation
* removed float16/half type, only rudimentary CPU support from PyTorch
* core module pulls io unqualified, random qualified, tensor qualified through to the parent module
* enhanced tensor module
    - introduced new constructor method
    - refactored existing code to utilize existing constructor
    - added a number of TODOs for missing documentation, tests, functionality, ...
    - added tested and documented astype cast function (see issue #10)
* refactored kmeans code
    - now utilizes the refactored random and tensor modules
    - refactored tests cases to utilize Python's unittest, need to be enhanced to better check KMeans functionality
    - added TODOs for missing documentation and functionality

### DIFF
--- a/heat/core/communicator.py
+++ b/heat/core/communicator.py
@@ -1,25 +1,91 @@
-import torch.distributed as dist
+import abc
+import torch.distributed as mpi
+
+from .stride_tricks import sanitize_axis
 
 
-def init_dist():
-    dist.init_process_group('mpi')
+def initialize_mpi():
+    mpi.init_process_group('mpi')
 
 
-init_dist()
+# initialize the MPI stack
+initialize_mpi()
 
 
-class NoneCommunicator:
+__all__ = [
+    'mpi'
+]
+
+
+class Communicator(metaclass=abc.ABCMeta):
+    @staticmethod
+    @abc.abstractstaticmethod
+    def is_distributed():
+        pass
+
+    @abc.abstractmethod
+    def __init__(self):
+        pass
+
+    @abc.abstractmethod
+    def chunk(self, shape, split):
+        """
+        Calculates the chunk of data that will be assigned to this compute node given a global data shape and a split
+        axis.
+
+        Parameters
+        ----------
+        shape : tuple of ints
+            the global shape of the data to be split
+        split : int
+            the axis along which to chunk the data
+
+        Returns
+        -------
+        out : tuple of slices
+            the chunk slices with respect to the given shape
+        """
+        pass
+
+
+class NoneCommunicator(Communicator):
     @staticmethod
     def is_distributed():
         return False
 
+    def __init__(self):
+        pass
 
-class MPICommunicator:
-    def __init__(self, group=dist.group.WORLD):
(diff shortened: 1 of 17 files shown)


heat/core/tests/test_types.py
```diff
@@ -0,0 +1,95 @@
...
+        self.assert_is_heat_type(heat_type)
+
+        # check a type constructor without any value
+        no_value = heat_type()
+        self.assertIsInstance(no_value, ht.tensor)
+        self.assertEqual(no_value.shape, (1,))
+        self.assertEqual((no_value._tensor__array == 0).all().item(), 1)
+        self.assertEqual(no_value._tensor__array.dtype, torch_type)
+
+        # check a type constructor with a complex value
```
d1saster: Complex value reminds me too much of a complex number. Wouldn't something like "elaborate input array" fit better, or just `elaborate_array` in order to have a short variable name?

heat/core/types.py
```diff
@@ -0,0 +1,257 @@
...
+        try:
+            array = torch.tensor(*value, dtype=torch_type)
+        except TypeError as exception:
+            # re-raise the exception to be consistent with numpy's exception interface
+            raise ValueError(str(exception))
+
+        return tensor.tensor(array, tuple(array.shape), split=None, comm=NoneCommunicator())
+
+    @classmethod
+    @abc.abstractclassmethod
```
d1saster: `abc.abstractclassmethod` has become deprecated in Python 3.3, according to the [documentation ](https://docs.python.org/3/library/abc.html#abc.abstractclassmethod). `abc.abstractmethod` should be sufficient.

heat/core/communicator.py
```diff
@@ -1,25 +1,91 @@
...
-class NoneCommunicator:
+
+__all__ = [
+    'mpi'
+]
+
+
+class Communicator(metaclass=abc.ABCMeta):
+    @staticmethod
+    @abc.abstractstaticmethod
```
d1saster: `abc.abstractstaticmethod` has become deprecated in Python 3.3, according to the [documentation ](https://docs.python.org/3/library/abc.html#abc.abstractstaticmethod). `abc.abstractmethod` should be sufficient. Btw. is there some special python 3 version we are targeting? Pytorch seems to need at least 3.5 if one chooses python3.

Issue #11, opened 2018-07-16
Support data type via dtype in factory functions


rainman110:
All factory function like asarray, zeros, ones etc must support a dtype argument as in
 - https://github.com/NeuralEnsemble/elephant/blob/ed8ad8bf477e080c1b79c20a206b65131b85ffa0/elephant/asset.py#L873

PR #43, opened 2018-08-02
Features/ones and zeros

Markus-Goetz:
Implemented ones, zeros, ones_like and zeros_like. Resolves issues #11, #13 and #14.


d1saster:
From what I onle the commit titles "Implemented ones-like and zeros-like function" is new, which looks great. Ship it!

PR #44, opened 2018-08-03
Features/ranges

Markus-Goetz:
Implements can_cast, promote_types, linspace. Resolves issue #23.


krajsek:
range operator is missing

PR #45, opened 2018-08-09
Features/ranges

krajsek:
Implemented arange function. Resolves issue #15

### DIFF
--- a/heat/core/tensor.py
+++ b/heat/core/tensor.py
@@ -234,7 +234,7 @@ def arange(*args, dtype=None, split=None):
     (in other words, the interval including `start` but excluding `stop`).
     For integer arguments the function is equivalent to the Python built-in
     `range <http://docs.python.org/lib/built-in-funcs.html>`_ function,
-    but returns an ndarray rather than a list.
+    but returns a tensor rather than a list.
 
     When using a non-integer step, such as 0.1, the results will often not
     be consistent.  It is better to use ``linspace`` for these cases.


krajsek:
Implemented arange function. Resolves issue #15

Markus-Goetz:
Generally, nice work! Builds and looks good, well documented and test. I have found two minor things, though, that I would suggest to be changed:

1. In the actual arange function, you define the default data types of the tensors in cases they are not given. The code makes use of torch dtypes, e.g. torch.int32. Is there a particular reason for this that I am not aware of? If not, it would be good to use the heat types instead, e.g. ht.int32.

2. This is an issue that I also faced and that requires a discussion with more people in my opinion. But as it currently stands, the following test (see link below) will not pass if you execute it with multiple processes/nodes. The internal Torch array will not always starts with 0.0, respectively end on 8.0.

[Test snippet](https://github.com/helmholtz-analytics/heat/blob/ac5ae700173327fbb123e3c8d432fcdd24928f56/heat/core/tests/test_tensor.py#L160-L169)

How do we deal with this? Shall we just assume that at some point the global __getitem__ operator will work and check this one instead?

heat/core/tensor.py
```diff
@@ -217,6 +226,108 @@ def __setitem__(self, key, value):
...
 
+def arange(*args, dtype=None, split=None):
+    """
+    Return evenly spaced values within a given interval.
+
+    Values are generated within the half-open interval ``[start, stop)``
+    (in other words, the interval including `start` but excluding `stop`).
+    For integer arguments the function is equivalent to the Python built-in
+    `range <http://docs.python.org/lib/built-in-funcs.html>`_ function,
+    but returns an ndarray rather than a list.
```
d1saster: s/ndarray/tensor/

heat/core/tests/test_tensor.py
```diff
@@ -46,7 +214,7 @@ def test_ones(self):
         self.assertEqual(simple_ones_uint.shape,  (5,))
         self.assertEqual(simple_ones_uint.lshape, (5,))
         self.assertEqual(simple_ones_uint.split,  None)
-        self.assertEqual(simple_ones_uint.dtype,  ht.uint8)
+        self.assertEqual(simple_ones_uint.dtype,  ht.bool)
```
d1saster: That looks strange?

heat/core/tests/test_tensor.py
```diff
@@ -92,7 +260,7 @@ def test_zeros(self):
         self.assertEqual(simple_zeros_uint.shape,  (5,))
         self.assertEqual(simple_zeros_uint.lshape, (5,))
         self.assertEqual(simple_zeros_uint.split,  None)
-        self.assertEqual(simple_zeros_uint.dtype,  ht.uint8)
+        self.assertEqual(simple_zeros_uint.dtype,  ht.bool)
```
d1saster: Same as above

heat/core/types.py
```diff
@@ -185,75 +190,277 @@ class flexible(generic):
...
     builtins.int:   int32,
     builtins.float: float32,
 }
 
 
-def as_torch_type(a_type):
+def canonical_heat_type(a_type):
     """
-    Infers the PyTorch type from a given HeAT type, type string or Python builtin type.
+    Canonizes the builtin Python type, type string or HeAT type into a canonical HeAT type.
```
d1saster: A really minor comment, but the literal translation of "canonizes" makes the sentence quite amusing

PR #48, opened 2018-08-29
Issue 38: Extend README.md

bhagemeier:
Markus,

if you think this is a reasonable start for the README, please merge it to master. I will add a documentation link once docs will be deployed automatically.

Cheers,
Björn


Markus-Goetz:
Hi Björn,

I think it looks good for a first draft, However, I have a few remarks:
1. The HeAT logo is not being displayed. Instead, it is a link that results in a 404
2. Would it be impossible to also include more of these status indicators, like the one that monitors the build status. In particular, I had some of the ones in mind that you can see here: https://github.com/MaineKuehn/cobald. Test coverage, license and, later, the docs build status would be good in my opinion. What is your opinion on that?
3. Do you think it is reasonable to have bullet-point list with the main features, like numpy has, see here: https://github.com/numpy/numpy?

PR #59, opened 2018-10-17
Issue #58: Make ht.randn check for sane input

### LINKED ISSUE #58: ht.randn(): unit test expects pytorch to detect malformed input
The unit test that checks whether `ht.randn` throws a `ValueError` depends on the fact that pytorch
shall detect malformed input. Unfortunately, this is not the case for my pytorch installation (trying to print the resulting tensor, however, results in an infinite loop, such that we can be sure that pytorch does not magically know how to handle negative dimensions ... )

For completeness the pytest output:
```
>           ht.randn(-1, 3, dtype=ht.float64)
E           AssertionError: ValueError not raised

heat/core/tests/test_tensor.py:278: AssertionError
```

There seems to be already some `all_ints` variable available in `ht.randn`, which could be augmented to also check for positive arguments?

d1saster:
So far ht.randn leaves the input checking mostly to pytorch.
This patch modifies ht.randn to at least check some of the
argument properties itself.

Fixes #58

### DIFF
--- a/heat/core/tensor.py
+++ b/heat/core/tensor.py
@@ -787,10 +787,10 @@ def randn(*args, dtype = torch.float32, split = None):
     num_of_param = len(args)
 
     # check if all positional arguments are integers and greater than zero
-    all_ints = all([isinstance(_, int) for _ in args])
+    all_ints = all(isinstance(_, int) for _ in args)
     if not all_ints:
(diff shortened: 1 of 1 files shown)


d1saster:
Thx, has been optimized :+1:

PR #65, opened 2018-10-23
Features/localops

ClaudiaComito:

### DIFF
--- a/heat/core/operations.py
+++ b/heat/core/operations.py
@@ -222,10 +222,9 @@ def max(x, axis=None):
     a : ht.tensor
     Input data.
         
-    axis : None or int  or #TODO: tuple of ints, optional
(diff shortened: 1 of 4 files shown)


Markus-Goetz:
Generally a good pull request, but I would suggest some changes:

1. We currently do not support tuples anywhere for reduction operations even though it should be (for all?) like this. This is an issue in itself. We should create another ticket for it and remove the TODOs. Instead a NotImplementedError could be raised on tuple input.

2. The out parameter is essential in my opinion and should be there. This might require to add it to other reduction operations as well in functionality and documentation

ClaudiaComito:
Thanks a lot Markus. I removed the TODOs and got stride_tricks to raise a NotImplementedError. Issues have been issued :) (#67 and #71)

PR #69, opened 2018-10-26
Features/sum

krajsek:
Revised tensor.sum() method such that always ht.tensors are returned. This differs from numpy where Python scalars are return when summed over all elements. I propose to differ from the numpy design here as it is IMO more consistent to return always the same data type if possible. A single scalar can always be obtained by a suitable method, e.g. a.value(). Also input checking and docstring have been added.


Markus-Goetz:
Question 1: why are the sqrt tests uncommented in a modification of the sum() code?
Suggestion 1: Move the sanitation code inside the __reduce_op method as it is needed for a number of reduction operations.
Suggestion 2: Most of the sanitation is already implemented in stride_tricks.sanitize_axis. Does anything speak again reusing the function?

krajsek:
Question 1: The tests did not run on my laptop and I was to lazy to look in the problem, so 
I removed them. I have reactivated them
Suggestion 1:  done 
Suggestion 2:  done

Markus-Goetz:
Leave it for now like this, Claudia faces a similar issue in her branch. Will think about a strategy there.

Issue #46, opened 2018-08-20
Add optional netCDF support


Markus-Goetz:
* Allow rudimentary netCDF support akin to what the HDF5 layer can already do
* Offer a possibility to write netCDF files

PR #75, opened 2018-11-04
Features/netcdf

Markus-Goetz:
Introduces APIs for:
* Storing tensors to HDF5
* Storing and loading tensors to netCDF4
* High-level APIs (save() and load()) for auto-detection of HDF5 and netCDF4 based on file extension

Modifications:
* Introduces mpi4py
* Removes the necessity for a manually built Torch

Resolves issue #46


Markus-Goetz:
Hi Claudia,
I do not have said test in my branch (or at least I cannot find it). Did you just double check whether your tests would run with the master pulled in?
Markus

ClaudiaComito:
OK, found the problem and it was on my end. All good.

heat/core/operations.py
```diff
@@ -105,7 +115,7 @@ def clip(a, a_min, a_max, out=None):
         raise ValueError('either a_min or a_max must be set')
 
     if out is None:
-        return tensor.tensor(a._tensor__array.clamp(a_min, a_max), a.shape, a.dtype, a.split, _copy(a._tensor__comm))
+        return tensor.tensor(a._tensor__array.clamp(a_min, a_max), a.shape, a.dtype, a.split, a.comm)
```
krajsek: I think we should spend setters for tensor attributes, i.e. allowing a.array.clamp(..
in this case and other accordingly

heat/core/communication.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @staticmethod
+    def as_buffer(obj):
+        if isinstance(obj, tensor.tensor):
+            obj = obj._tensor__array
+
+        if not isinstance(obj, torch.Tensor):
+            return obj
+
+        return MPI.memory.fromaddress(obj.cpu().data_ptr(), obj.element_size() * torch.numel(obj))
```
krajsek: "obj.cpu()", what is about cuda aware mpi? I think one can hack something with pycuda. We could postphone this for later but should not forget about it.

Issue #73, opened 2018-10-29
Fix broadcast of __binop 


krajsek:
broadcast of __binop does not work for certain cases in distributed mode, e.g. 
na = ht.ones((1,4), split=0)
nb = ht.ones((4,1), split=0)
na+nb
throws:
RuntimeError: The size of tensor a (0) must match the size of tensor b (2) at non-singleton dimension 0
for mpirun -np >1

PR #77, opened 2018-11-05
Features/convolve

krajsek:
Implementation of heat.convolve():
This is mainly a wrapper around the torch function conv1d in order to emulate the np.convolve() function. There are  differences compared to the numpy design :
-  The inputs are not swapped if v is larger than a. The reason is that v needs to be 
    non-splitted. This should not influence performance. If the filter weight is larger 
    than fitting into memory, using the FFT for convolution is recommended.
-  Only float, not integer,  data types are allowed. This follows from the design of the torch function conv1d which runs only for floats.

The heat.convolve() functions fetches halos is run in distributed mode if they are not a already stored
as attributes in the tensor class. Already stored halos are processed by local operations in the same way as the data array. This does not work for binary operations. First issue #73 needs to be resolved.

### DIFF
--- a/heat/core/communicator.py
+++ b/heat/core/communicator.py
@@ -56,7 +56,9 @@ class NoneCommunicator(Communicator):
         return False
 
     def __init__(self):
-        pass
+        self.group = None
+        self.rank = 0
+        self.size = 1
 
     def chunk(self, shape, split):
         # ensure the split axis is valid, we actually do not need it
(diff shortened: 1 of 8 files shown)


Markus-Goetz:
Quick remark, have not checked yet everything: I think it would be good to replace the newly introduced is_integer, is_bool, ... method with issubclass check on the respective HeAT types instead (e.g. ht.bool).

Motivation is twofold. First, this invariant should always be true even if we introduced specialized subtypes that potentially forget to override their respective methods and, second, we can later also accept numpy/torch types via mixins. What are your thoughts on this?

Will check the rest of the code soon.

PR #86, opened 2018-11-26
Install all packages, not just top level heat

bhagemeier:
Add heat.core, heat.ml, and heat.ml.cluster to the list of packages. Also,
install README.md and LICENSE file. README.md is required by setup.py for the
long_description field.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,8 @@ with open("README.md", "r") as fh:
 
 setup(
     name='heat',
-    packages=['heat'],
+    packages=['heat', 'heat.core', 'heat.ml', 'heat.ml.cluster'],
+    data_files=['README.md', 'LICENSE'],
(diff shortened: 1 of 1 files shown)


Markus-Goetz:
How is the pip policy for unittests? If they are copied over with the package, we also need to include /heat/datasets as to be able to execute the unittests

bhagemeier:
> How is the pip policy for unittests? If they are copied over with the package, we also need to include /heat/datasets as to be able to execute the unittests

My personal take on this is to omit test packages from a real installation. However, for lack of other code examples, one can also argue to include them. Howerver, what would be a good location for the data then: /share/heat/datasets?

Markus-Goetz:
Good point. I am not sure. So lets not include the datasets as of now and instead provide some meaningful examples later.

Approved by me

PR #113, opened 2019-02-13
Features/98 moments 2

coquelin77:
This pull request only includes the mean/var/std functions. there is no higher order moments here. Mean was tested for arrays up to 5D. Tests are all singular and multiple dimension combinations for the mean for a 5D array for all split axes. var was tested similarly but without the multiple dimension combinations (not implemented in pytorch).

This also includes a few other changes: 
- kwargs in the _local__operation
- distributed random.randn
- lnumel (local number of elements)
- size or gnumel (global number of elements)
- item() (for getting the sole element of a 1-D tensor)
- typo correction in stride axes

NOTE: the element-wise update functions are slow. this will need to be optimized. However, the functions for the cases where either the whole set is selected or the target dimension is equal to the split dimension are both much faster.


coquelin77:
i wanted to get this in before i left for the US. i wont be able to work on this for the next 7 days so if anything else is wrong with it please let me know and i will update it accordingly. My apologies for starting this and then leaving.

Markus-Goetz:
Does not build

coquelin77:
build fixed. also changed the lshape call here. see commit db8fba1

heat/core/tensor.py
```diff
@@ -29,17 +30,38 @@ def dtype(self):
...
+            # todo: throw a soft warning about taking the number of elements of a single element tensor
+            return 1
+
+    @property
+    def gnumel(self):
+        return self.size
+
+    @property
+    def lnumel(self):
+        return reduce(lambda i, j: i * j, self.lshape)
```
Markus-Goetz: code duplication with line 36. Proposal: introduce a prod(<iterable>) function implemented as you did or via

```
import operator
def prod(iterable)
    reduce(operator.mul, iterable, 1)
```

heat/core/tensor.py
```diff
@@ -29,17 +30,38 @@ def dtype(self):
...
+
     @property
     def lshape(self):
         if len(self.__array.shape) == len(self.__gshape):
             return tuple(self.__array.shape)
         # edge case when the local data tensor receives no elements after chunking
-        return self.__gshape[:self.__split] + (0,) + self.__gshape[self.split + 1:]
+        return self.__array.shape
+        # return self.__gshape[:self.__split] + (0,) + self.__gshape[self.split + 1:]
```
Markus-Goetz: please remove dead code in comments

in what way is the old implementation insufficient?

heat/core/tensor.py
```diff
@@ -29,17 +30,38 @@ def dtype(self):
...
 
     @property
     def shape(self):
         return self.__gshape
 
+    @property
+    def array_shape(self):
```
Markus-Goetz: how is this different from lshape?

heat/core/tensor.py
```diff
@@ -271,12 +293,84 @@ def max(self, axis=None, out=None):
 
         return operations.max(self, axis, out)
 
-    def mean(self, axis):
-        # TODO: document me
-        # TODO: test me
-        # TODO: sanitize input
-        # TODO: make me more numpy API complete
-        return self.sum(axis) / self.shape[axis]
+    def mean(self, dimen=None, all_procs=False):
```
Markus-Goetz: Should expose the numpy.mean() interface (see here: https://docs.scipy.org/doc/numpy/reference/generated/numpy.mean.html)

heat/core/tensor.py
```diff
@@ -271,12 +293,84 @@ def max(self, axis=None, out=None):
...
+                    If True: will split the result in the same direction as x
+                    Default: False (mean of the whole dataset still calculated but not on every node)
+
+        Returns
+        -------
+        ht.tensor containing the mean/s, if split, then split in the same direction as x.
+        """
+        return operations.mean(self, dimen, all_procs)
+
+    def var(self, dimen=None, all_procs=False, bessel=True):
```
Markus-Goetz: Should expose the numpy.var() interface (https://docs.scipy.org/doc/numpy/reference/generated/numpy.var.html)

heat/core/tensor.py
```diff
@@ -271,12 +293,84 @@ def max(self, axis=None, out=None):
...
+                 use the bessel correction when calculating the varaince/std
+                 toggle between unbiased and biased calculation of the var
+
+        Returns
+        -------
+        ht.tensor containing the var/s, if split, then split in the same direction as x.
+        """
+        return operations.var(self, dimen, all_procs, bessel=bessel)
+
+    def std(self, dimen=None, all_procs=False, bessel=True):
```
Markus-Goetz: Should expose the numpy.std() interface (https://docs.scipy.org/doc/numpy/reference/generated/numpy.std.html)

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+        mu1 = 0.0
+    if mu2.item() == np.nan:
+        mu2 = 0.0
+    delta = mu2.item() - mu1.item()
+    n1 = n1.item()
+    n2 = n2.item()
+    return mu1 + n2 * (delta / (n1+n2)), n1 + n2
+
+
+def mean(x, dimen=None, all_procs=False):
```
Markus-Goetz: Should expose the numpy.mean() interface

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+    number of elements in the combined set
+    """
+    if mu1.item() == np.nan:
+        mu1 = 0.0
+    if mu2.item() == np.nan:
+        mu2 = 0.0
+    delta = mu2.item() - mu1.item()
+    n1 = n1.item()
+    n2 = n2.item()
+    return mu1 + n2 * (delta / (n1+n2)), n1 + n2
```
Markus-Goetz: pep-8 spacing (n1+n2)

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+          Calculated mean
+    n2 : 1D ht.tensor or 1D torch.tensor
+         number of elements used to calculate mu2
+
+    Returns
+    -------
+    mean of combined set
+    number of elements in the combined set
+    """
+    if mu1.item() == np.nan:
```
Markus-Goetz: does comparison against np.nan work for GPUs?

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+                Flag to distribute the data to all processes
+                If True: will split the result in the same direction as x
+                Default: False (mean of the whole dataset still calculated but not on every node)
+
+    Returns
+    -------
+    ht.tensor containing the mean/s, if split, then split in the same direction as x.
+    """
+
+    def combine_all_means(target_dimens):
```
Markus-Goetz: target_dimens should be axis

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+    If a dimension is given, the mean will be taken in that direction.
+
+    Parameters
+    ----------
+    x : ht.tensor
+        Values for which the mean is calculated for
+    dimen : None, Int, iterable
+            Dimension which the mean is taken in.
+            Default: None -> mean of all data calculated
+    all_procs : Bool
```
Markus-Goetz: all_procs seems to be an unnecessary flag. Heat has rules for the calculation of reduction operations across split axis. There are two major cases:

a) the reduction is performed across the split axis, then the result tensor is not split and a copy on each node
b) the reduction is performed along the split axis, the result tensor has the same split axis, all the nodes have their own partial chunk

Should be written down somewhere, agreed!

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+                    mu_tot_hold[:, :, :, beg:end] = mu_proc
+                elif x.split - score == 4:
+                    mu_tot_hold[:, :, :, :, beg:end] = mu_proc
+                else:
+                    print('else', mu_tot_hold[beg:end].lshape, mu_proc.lshape)
+                    mu_tot_hold[beg:end] = mu_proc
+                beg = end
+            beg = x.comm.bcast(beg, r)
+        x.comm.Allreduce(mu_tot_hold, mu_tot, MPI.SUM)
+        x.comm.barrier()
```
Markus-Goetz: Allreduce has an implicit barrier

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+                elif x.split - score == 1:
+                    mu_tot_hold[:, beg:end] = mu_proc
+                elif x.split - score == 2:
+                    mu_tot_hold[:, :, beg:end] = mu_proc
+                elif x.split - score == 3:
+                    mu_tot_hold[:, :, :, beg:end] = mu_proc
+                elif x.split - score == 4:
+                    mu_tot_hold[:, :, :, :, beg:end] = mu_proc
+                else:
+                    print('else', mu_tot_hold[beg:end].lshape, mu_proc.lshape)
```
Markus-Goetz: remove print

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+            # multiple dimensions
+            score = [True for itt in dimen if itt < x.split].count(True)
+        elif isinstance(dimen, (int, float)):
+            # one dimension
+            score = 1 if dimen < x.split else 0
+        for r in range(x.comm.Get_size()):
+            # need to add them together in order
+            if x.comm.rank == r:
+                end = beg + x.lshape[x.split]
+                if x.split - score == 0:
```
Markus-Goetz: what about arbitrary dimensions? you can compress this by explicitly calling __getitem__() with the proper slices filled with None:

`mu_tot_hold.__getitem__(x.split - score * [None] + [slice(beg, end)])`

(needs to be adjusted slightly maybe, from brain to terminal)

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+        sz = x.comm.Get_size()
+        rem1 = rem2 = 0
+        while True:
+            if sz % 2 != 0:  # test if the length is an odd number
+                if rem1 != 0 and rem2 == 0:
+                    rem2 = sz - 1
+                elif rem1 == 0:
+                    rem1 = sz - 1
+            splt = int(sz / 2)
+            for sp_it in range(splt):  # todo: multithread perfect opp for GPU parrallelizm
```
Markus-Goetz: issue as ticket or fix in PR

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+                return mu_tot[0]
+    # ------------------------------------------------------------------------------------------------------------------
+    if dimen:
+        if isinstance(dimen, int):
+            if dimen >= len(x.shape):
+                raise ValueError("Dimension (dimen) must be < {}, currently is {}".format(len(x.shape), dimen))
+            dimen = dimen if dimen > 0 else dimen % len(x.shape)
+        else:
+            raise TypeError("Dimension (dimen) must be an int, currently is {}".format(type(dimen)))
+        # todo: travis is failing when a tuple is given for the dimension in the mean although it works in the newest pytorch.
```
Markus-Goetz: you can force an update of pytorch on travis

heat/core/operations.py
```diff
@@ -544,7 +549,620 @@ def sum(x, axis=None, out=None):
...
+        if isinstance(dimen, (list, tuple, np.ndarray, tensor.tensor)):
+            # multiple dimensions
+            score = [True for itt in dimen if itt < x.split].count(True)
+        elif isinstance(dimen, (int, float)):
+            # one dimension
+            score = 1 if dimen < x.split else 0
+        for r in range(x.comm.Get_size()):
+            # need to add them together in order
+            if x.comm.rank == r:
+                end = beg + x.lshape[x.split]
```
Markus-Goetz: see proposed slicing from mean()

Markus-Goetz:
First comments, need to look in detail

heat/core/tensor.py
coquelin77: the old way would return a 0 along the split axis and not the true local shape. 
i.e. when run with two processes
```python
z = ht.zeros([5, 4, 3], split=1)
z.lshape
>>> (5, 2, 3)
z.gshape[:z.split] + (0,) + z.gshape[z.split + 1:]
>>> (5, 0, 3)
```

I had left the old code so we could discuss this and i wouldn't have to look elsewhere for it.

heat/core/tensor.py
coquelin77: what about numpy.prod?

heat/core/tensor.py
coquelin77: artifact from development i missed when cleaning up. now removed

heat/core/operations.py
coquelin77: what i meant by this is that it is the dimensions of the output tensor and not the axis given to the function

heat/core/operations.py
coquelin77: okay my idea was that if at some time it was desired to have all of the means on one node then it could be done easily but if that is the case then i will simply remove the flag and the accompanying code

heat/core/tensor.py
Markus-Goetz: Sure would work as well

heat/core/tensor.py
Markus-Goetz: Okay, so the computation is actually more tricky than this even. Considers this:

`mpirun -np 5 python -c "import heat as ht; x = ht.zeros((5,4,3), split=1)"`

Rank 4 should correctly report that the lshape is (5,0,3), while all the others should report (5,1,3). While not always necessary sensible to have a such a high degree of parallelization with such a small amount of data, there are situations where this might happens.

heat/core/operations.py
Markus-Goetz: Alright, my bad. Could we call out output_shape to be consistent with numpy speak? Also, it there a need for the inline function except the parameter closure?

heat/core/tensor.py
coquelin77: testing with the new implementation does return that if it is the case:

```python
z = ht.zeros([5, 4, 3], split=1)
print(z.lshape)
>>>mpiexec -n 5 python testing.py 
>>>(5, 1, 3)
>>>(5, 1, 3)
>>>(5, 1, 3)
>>>(5, 1, 3)
>>>(5, 0, 3)

File: heat/core/operations.py (hunk 10 above)
Comment (coquelin77) [reply heat/core/operations.py]: There isnt a true need for the inline functions but i didnt want to have to put the same code twice if it's not needed. 

output_shape change is made

File: heat/core/tensor.py (hunk 2 above)
Comment (Markus-Goetz) [reply heat/core/tensor.py]: Alright, looks good

**Comment by coquelin77** [comment]
failing due to multi dimensional mean again. unsure about why. works locally with the same install as is said in the script. removing multi-dim tests for the moment

File: heat/core/operations.py (hunk 9 above)
Comment (coquelin77) [reply heat/core/operations.py]: removed the comparison. was only for handling edge cases. now handled elsewhere

File: heat/core/operations.py (hunk 14 above)
Comment (coquelin77) [reply heat/core/operations.py]: this was removed as it is not needed in the new implementation (creation of a distributed heat tensor)

**Comment by coquelin77** [comment]
Both @Cdebus and myself have seen some odd behavior or travis after updating to pytorch 1.0.1 (requirement: >= 1.0.0). I apologies for spamming you all with the commits but I could not replicate the problem locally. travis fails at the sqrt test 
```python
self.assertTrue((float32_sqrt._tensor__array == comparison.type(torch.float32)).all())
```
with some added prints to try to diagnose the problem i found that the reason for this stems from the comparison with the last value. 

``` python
print((float32_sqrt._tensor__array == comparison.type(torch.float32)).all())
>>> tensor(0, dtype=torch.uint8)
print(float32_sqrt._tensor__array == comparison.type(torch.float32))
>>> tensor([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0], dtype=torch.uint8)
print(float32_sqrt._tensor__array, comparison.type(torch.float32))
>>> tensor([0.0000, 1.0000, 1.4142, 1.7321, 2.0000, 2.2361, 2.4495, 2.6458, 2.8284,
        3.0000, 3.1623, 3.3166, 3.4641, 3.6056, 3.7417, 3.8730, 4.0000, 4.1231,
        4.2426, 4.3589]) 
>>> tensor([0.0000, 1.0000, 1.4142, 1.7321, 2.0000, 2.2361, 2.4495, 2.6458, 2.8284,
        3.0000, 3.1623, 3.3166, 3.4641, 3.6056, 3.7417, 3.8730, 4.0000, 4.1231,
        4.2426, 4.3589])
```
(spacing added to last entry to assist in comparison)

When these prints are left in the code then travis will often function. However I would greatly appreciate more eyes on this issue. I am very curious if others also see this issue

heat/core/operations.py
coquelin77: there is a GPU course at JSC coming up and I plan to use this as my problem to be solved

heat/core/operations.py
```diff
@@ -1,10 +1,13 @@
 import itertools
 import torch
+import warnings
+import numpy as np
 
 from .communication import MPI
 from . import stride_tricks
 from . import types
 from . import tensor
+import builtins
```
Markus-Goetz: Python packages and 3rd party packages should go to the top (if possible sorted alphabetically). Our import after that. builtins should be at the top

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
 
 
+def merge_means(mu1, n1, mu2, n2):
+    """
+    Function to merge two means by pairwise update
+
+    Parameters
+    ----------
+    mu1 : 1D ht.tensor or 1D torch.tensor
+          Calculated mean
```
Markus-Goetz: Indentation of explanation inconsistent

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+    """
+    Calculates and returns the mean of a tensor.
+    If a axis is given, the mean will be taken in that direction.
+
+    Parameters
+    ----------
+    x : ht.tensor
+        Values for which the mean is calculated for
+    axis : None, Int, iterable
+           axis which the mean is taken in.
```
Markus-Goetz: Indentation inconsistent

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+           Default: None -> mean of all data calculated
+
+    Returns
+    -------
+    ht.tensor containing the mean/s, if split, then split in the same direction as x.
```
Markus-Goetz: A number of example could be useful here. Maybe you can just use/adapt the ones on the numpy or torch mean() page.

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+    def reduce_means_elementwise(output_shape_i):
+        """
+        Function to combine the calculated means together.
+        This does an element-wise update of the calculated means to merge them together using the merge_means function.
+        This function operates using x from the mean function paramters
+
+        Parameters
+        ----------
+        output_shape_i : iterable
+                        iterable with the dimensions of the output of the mean function
```
Markus-Goetz: Indentation inconsistent with style template

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+        Returns
+        -------
+        ht.tensor of the calculated means
+        """
+        if x.lshape[x.split] != 0:
+            mu = __local_operation(torch.mean, x, out=None, dim=axis)
+        else:
+            mu = tensor.zeros(output_shape_i)
+
+        mu_to_combine = tensor.zeros((x.comm.Get_size(),) + tuple(output_shape_i))
```
Markus-Goetz: rank and size are members of x.comm

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+
+        sz = x.comm.Get_size()
+        rem1 = rem2 = 0
+        while True:
+            if sz % 2 != 0:
+                if rem1 != 0 and rem2 == 0:
+                    rem2 = sz - 1
+                elif rem1 == 0:
+                    rem1 = sz - 1
+            splt = int(sz / 2)
```
Markus-Goetz: you can use // here, does integer division but more efficient due to low-level implementation

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+
+        n_for_merge = tensor.zeros(x.comm.Get_size())
+        n2 = tensor.zeros(x.comm.Get_size())
+        mu_to_combine[x.comm.Get_rank()] = mu
+        n2[x.comm.Get_rank()] = x.lshape[x.split]
+        x.comm.Allreduce(mu_to_combine, mu_tot, MPI.SUM)
+        x.comm.Allreduce(n2, n_for_merge, MPI.SUM)
+
+        sz = x.comm.Get_size()
+        rem1 = rem2 = 0
```
Markus-Goetz: Better use vector assignment, i.e. `rem1, rem2 = 0, 0`

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+        while True:
+            if sz % 2 != 0:
+                if rem1 != 0 and rem2 == 0:
```
Markus-Goetz: you can directly check for "true-ness" of values by doing for example:

`if rem1 and not rem2:`

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+                    rem2 = sz - 1
+                elif rem1 == 0:
+                    rem1 = sz - 1
+            splt = int(sz / 2)
+            for sp_it in range(splt):  # todo: multithread for GPU parrallelizm
```
Markus-Goetz: Do you have an idea how?

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+                mu_reshape = __local_operation(torch.reshape, mu_tot[sp_it], out=None, shape=(1, int(mu_tot[sp_it].lnumel)))[0]
+                for en, (el1, el2) in enumerate(zip(mu_reshape,
+                                                    __local_operation(torch.reshape, mu_tot[sp_it+splt], out=None,
```
Markus-Goetz: This seems like there is a lot of code repetition going on here. I am not 100% sure (please help me here), but is it possible to prepare the parameters for the __local_operation call in the if-clauses and then have the call afterwards? Something along the lines of

```
if rem1 and not rem2:
    tensor, output_shape = mu_tot[...], (1, ...,)
elif rem1 and rem2:
    tensor, output_shape = ...

__local_operation(tensor, output_shape)
```

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+            if sz % 2 != 0:
+                if rem1 != 0 and rem2 == 0:
+                    rem2 = sz - 1
+                elif rem1 == 0:
+                    rem1 = sz - 1
+            splt = int(sz / 2)
+            for i in range(splt):  # todo: multithread for GPU
+                mu_reshape = __local_operation(torch.reshape, mu_tot[i], out=None, shape=(1, int(mu_tot[i].lnumel)))[0]
+                var_reshape = __local_operation(torch.reshape, var_tot[i], out=None, shape=(1, int(mu_tot[i].lnumel)))[0]
+                for en, (mu1, var1, mu2, var2) in enumerate(zip(mu_reshape, var_reshape,
```
Markus-Goetz: Same as above for the mean() implementation. Is it possible to avoid the code duplication here?

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+            elif axis == x.split:
+                return reduce_vars_elementwise(output_shape)
+            else:
+                try:
+                    return tensor.array(__local_operation(torch.var, x, out=None, dim=axis, unbiased=bessel), split=x.split, comm=x.comm)
+                except ValueError:
+                    return __local_operation(torch.var, x, out=None, dim=axis, unbiased=bessel)
+        else:
+            raise TypeError("Axis (axis) must be an int, currently is {}. Check if multidim var is available in pyTorch".format(type(axis)))
+            # TODO: when multi dim support is available from pytorch this can be uncommented and the raise above can be changed
```
Markus-Goetz: Should this be removed or documented in another ticket?

heat/core/tensor.py
```diff
@@ -300,12 +313,84 @@ def max(self, axis=None, out=None):
...
+                    If True: will split the result in the same direction as x
+                    Default: False (var of the whole dataset still calculated but not available on every node)
+        bessel : Bool
+                 Default: True
+                 use the bessel correction when calculating the varaince/std
+                 toggle between unbiased and biased calculation of the std
+
+        Returns
+        -------
+        ht.tensor containing the std/s, if split, then split in the same direction as x.
```
Markus-Goetz: Examples are needed

heat/core/tensor.py
```diff
@@ -300,12 +313,84 @@ def max(self, axis=None, out=None):
...
+                    Flag to distribute the data to all processes
+                    If True: will split the result in the same direction as x
+                    Default: False (var of the whole dataset still calculated but not available on every node)
+        bessel : Bool
+                 Default: True
+                 use the bessel correction when calculating the varaince/std
+                 toggle between unbiased and biased calculation of the var
+
+        Returns
+        -------
```
Markus-Goetz: Examples are needed

heat/core/tensor.py
```diff
@@ -300,12 +313,84 @@ def max(self, axis=None, out=None):
...
+                Dimension which the mean is taken in.
+                Default: None -> mean of all data calculated
+        all_procs : Bool
+                    Flag to distribute the data to all processes
+                    If True: will split the result in the same direction as x
+                    Default: False (mean of the whole dataset still calculated but not on every node)
+
+        Returns
+        -------
+        ht.tensor containing the mean/s, if split, then split in the same direction as x.
```
Markus-Goetz: Examples are needed

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+            Default: None -> std of all data calculated
+            NOTE -> if multidemensional var is implemented in pytorch, this can be an iterable. Only thing which muse be changed is the raise
+    bessel : Bool
+             Default: True
+             use the bessel correction when calculating the varaince/std
+             toggle between unbiased and biased calculation of the std
+
+    Returns
+    -------
+    ht.tensor containing the std/s, if split, then split in the same direction as x.
```
Markus-Goetz: Examples are needed

heat/core/operations.py
```diff
@@ -509,6 +515,494 @@ def sum(x, axis=None, out=None):
...
+           Default: None -> var of all data calculated
+           NOTE -> if multidemensional var is implemented in pytorch, this can be an iterable. Only thing which muse be changed is the raise
+    bessel : Bool
+             Default: True
+             use the bessel correction when calculating the varaince/std
+             toggle between unbiased and biased calculation of the std
+
+    Returns
+    -------
+    ht.tensor containing the var/s, if split, then split in the same direction as x.
```
Markus-Goetz: Examples are needed

heat/core/operations.py
coquelin77: i was saving this for a GPU class that is being taught at JSC at the beginning of next month

Markus-Goetz: Okay, sounds good

heat/core/operations.py
coquelin77: fixed for next commit

heat/core/operations.py
coquelin77: fixed for next commit

heat/core/operations.py
coquelin77: It could be either. the code is there when it's available in pytorch. the only changes required are to uncomment that code block and to change the raise to allow iterables.

coquelin77: see issue #156

coquelin77:
will need to fix the item function to allow for tensors with shape 1, as well (incoming with the global setters and getters) **reminder for myself

coquelin77:
got verbal confirmation from @Markus-Goetz for this merge. 

these functions are still not optimized. This is in the pipeline and will be coming in the future

Issue #54, opened 2018-09-15
ht.randn(): Seed-identical number generation not addressed in distributed case 


ClaudiaComito:

coquelin77:
this issue is also referenced in #110

Markus-Goetz:
Will have a working thing soon online

PR #114, opened 2019-02-13
features/110-split_output_randn: Adding the new code for random

coquelin77:
see issue 110.

heads up @ClaudiaComito i mentioned this in the issue as well but this will change the tests for argmin, min, and max. This is because the random numbers generated on separate processes are not equal but they are the same every run.

### DIFF
--- a/heat/core/random.py
+++ b/heat/core/random.py
@@ -30,7 +30,7 @@ def randn(*args, split=None, comm=MPI_WORLD):
 
     Parameters
     ----------
-    d0, d1, …, dn : int, optional
+    d0, d1, …, dn : ints, optional
         The dimensions of the returned array, should be all positive.
 
     Returns
@@ -55,6 +55,13 @@ def randn(*args, split=None, comm=MPI_WORLD):
             [-1.8548, -1.2574,  0.2391, -0.3302],
             [ 1.3365, -1.5212,  1.4159, -0.1671],
             [ 0.1260,  1.2126, -0.0804,  0.0907]])
+
+    >>> ht.randn(4, 4, split=0)
+    (two processes)
+    tensor([[-1.1261,  0.5971,  0.2851,  0.9998],
+            [-1.8548, -1.2574,  0.2391, -0.3302]])
+    tensor([[ 1.3365, -1.5212,  1.4159, -0.1671],
+            [ 0.1260,  1.2126, -0.0804,  0.0907]])
     """
     # check if all positional arguments are integers
     if not all(isinstance(_, int) for _ in args):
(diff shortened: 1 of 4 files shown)


heat/core/tests/test_operations.py
```diff
@@ -170,10 +171,10 @@ def test_argmin(self):
         random_data = ht.random.randn(3,3,3)
         torch.manual_seed(1)
         random_data_split = ht.random.randn(3,3,3,split=0)
-        
-        self.assertTrue((ht.argmin(random_data,axis=0)._tensor__array == random_data_split.argmin(axis=0)._tensor__array).all())
-        self.assertTrue((ht.argmin(random_data,axis=1)._tensor__array == random_data_split.argmin(axis=1)._tensor__array).all())
-        self.assertIsInstance(ht.argmin(random_data_split,axis=1),ht.tensor)    
+        # todo: these tests now fail if random_data is split.
```
Markus-Goetz: Remove or fix test

heat/core/tests/test_operations.py
```diff
@@ -387,9 +388,9 @@ def test_max(self):
         random_data = ht.random.randn(3,3,3)
         torch.manual_seed(1)
         random_data_split = ht.random.randn(3,3,3,split=0)
-        
-        self.assertTrue((ht.max(random_data,axis=0)._tensor__array[0] == random_data_split.max(axis=0)._tensor__array[0]).all())
-        self.assertTrue((ht.max(random_data,axis=1)._tensor__array[0] == random_data_split.max(axis=1)._tensor__array[0]).all())
+        # todo: these tests now fail if random_data is split.
```
Markus-Goetz: Remove or fix text

heat/core/random.py
```diff
@@ -30,7 +30,7 @@ def randn(*args, split=None, comm=MPI_WORLD):
 
     Parameters
     ----------
-    d0, d1, …, dn : int, optional
+    d0, d1, …, dn : ints, optional
```
Markus-Goetz: Implementation of the function is closely related to issue #54. What is the intended behaviour of this function? Output a number of random numbers where each of the split parts is random and not identical to the other split chunks of the other nodes? What will happen if I use the very same seed but a different node count? Will the random tensor differ?

heat/core/tests/test_operations.py
```diff
@@ -474,9 +475,9 @@ def test_min(self):
         random_data = ht.random.randn(3,3,3)
         torch.manual_seed(1)
         random_data_split = ht.random.randn(3,3,3,split=0)
-        
-        self.assertTrue((ht.min(random_data,axis=0)._tensor__array[0] == random_data_split.min(axis=0)._tensor__array[0]).all())
-        self.assertTrue((ht.min(random_data,axis=1)._tensor__array[0] == random_data_split.min(axis=1)._tensor__array[0]).all())
+        # todo: these tests now fail if random_data is split.
```
Markus-Goetz: Remove or fix test

heat/core/random.py
coquelin77: It looks like torch's randn will not generate the same values for arrays of different sizes. This means that even though the same seed is used, the split tensor will not be the same as a tensor generated only on one process. i.e.
```python
torch.randn((3, 3, 3))[0] != torch.randn((1, 3, 3))
```
One solution to this would be to generate the whole dataset then split it. But this will not scale. I do not see another way to do this. 

It must be noted that the matrix generated by the ht.randn with a fixed seed (torch.manual_seed) and a fixed size produces a reproducible result

coquelin77: if the sizes of the data are the same on multiple nodes, then the random data will be the same.

Markus-Goetz: In my opinion, we can leave it as proposed for now, however, for the future, I would want to actually have a different behaviour. Consider the following example:

```
ht.set_seed(1)
ht.randn(100, 5, 3, split=0)
```

Should always produce the same set of random numbers independent of the utilized nodes (for reproducibility reasons as you mentioned). This is exactly what is requested in issue #54. Obviously, this means that we would have to come up with a pseudo random generator allowing to skip to arbitrary/some fixed positions into the random sequence. 

In the proposed fix, we would have a simplified randn() call. It will only provide reproducible results for the exact same node count.

coquelin77: 100% agree. i am trying to figure a way to do this. i will move this discussion to the issue and close this request.

Issue #94, opened 2018-12-05
Add float16 support to types


Markus-Goetz:

Markus-Goetz:
PyTorch currently implements fp16 support for GPUs only. This issue will have to be resolved after HeAT has device support.

First implementation idea is to check the device the HeAT tensor shall be allocated on and not allow CPU backend.

mtar:
PyTorch has basic operations for float16 and bfloat16 by now. The function support, however, has as many holes as a Swiss cheese.

shahpratham:
> The function support, however, has as many holes as a Swiss cheese.

Yes, I tried to add float16 but later many test were failed as `torch.ceil`, `torch.floor` and many more function were not implemented for float16.

ClaudiaComito:
This issue is still open, and many torch functions still don't support `Half` data type. 

Proposal:
- introduce `ht.float16` datatype
- introduce a @SkipWithHalf (or similar) decorator for those tests that otherwise will fail
- whenever a new pytorch version is out, we test without the decorator and update.

PR #116, opened 2019-02-19
WIP: Included float16 into supported types

Cdebus:
- Concerns #94 
- Tests still needed

### DIFF
--- a/heat/core/tests/test_types.py
+++ b/heat/core/tests/test_types.py
@@ -83,6 +83,10 @@ class TestTypes(unittest.TestCase):
     def test_floating(self):
         self.assert_non_instantiable_heat_type(ht.floating)
 
+    def test_float16(self):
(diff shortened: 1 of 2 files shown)


Cdebus:
Issue in float26 implementation: Pytorch support

When implementing unit-tests for the float16 support in types, there was an issue with assert_is_instantiable_heat_type() :
apparently, pytorch does not support the "==" operation for torch.HalfTensor (exemplary also in https://github.com/pytorch/pytorch/issues/14344). This leads to an error in  self.assertEqual((no_value._tensor__array == 0).all().item(), 1)
--> E       RuntimeError: _th_eq is not implemented for type torch.HalfTensor

@Markus-Goetz: Under these circumstances, should there be a ht.float16 type at all, and we just make a query for that in the unit test? Or not support this datatype either?

PR #137, opened 2019-02-22
Implementation of finfo and iinfo

Cdebus:
Solves #36 
Minimal requirements for implementation of numpy.finfo analogy for Asset support. 
Returns attributes 'eps', 'min','max', 'bits' and 'tiny' for heat floating point values (from torch_finfo)
Future amendments could include more detailed info on the floating point representation (e.g. number of bits in mantissa, number of bits in exponent)


rainman110:
Unfortunately, our Pytorch on the CI has not a finfo function. Look here for a potential workaround:

https://github.com/pytorch/pytorch/issues/10742

Maybe it's time to update Pytorch in our CI.

Markus-Goetz:
Heat should build without any problem with Torch 1.0.1. Could you please change the setup.py so that it requires at least torch 1.0.0? That should also fix the failing CI build

heat/core/tests/test_types.py
```diff
@@ -189,3 +189,20 @@ def test_type_promotions(self):
             ht.promote_types(1, '?')
         with self.assertRaises(TypeError):
             ht.promote_types(ht.float32, 'hello world')
+
+class TestTypeConversion(unittest.TestCase):
+    def test_finfo(self):
+        info32 = ht.types.finfo(ht.float32)
```
Markus-Goetz: It would be good if finfo was accessible already on module level to be consistent with numpy, i.e.

`ht.finfo(ht.float32)`

Just needs a qualified import in one of the __init__.py. The implementation is in the correct file though.

Cdebus: Indeed a good idea, I added finfo as  import to the init.py to cover this.

coquelin77:
on the 110-moments_2 pull request I successfully built with pytorch 1.0.1. feel free to use my travis.yml and setup.py.

Cdebus:
Concerning #141 :
I have implemented iinfo() directly into this branch, rather than opening a new one, since it works in perfect analogy to iinfo.

Markus-Goetz:
Resolves issues #36 and #141

PR #154, opened 2019-03-11
separated operations.py/test_operations.py into multiple files. 

coquelin77:
created arithmetics and constants files even though there is nothing to fill them.

also added cos function as it is a simple change from the sin function


coquelin77:
i want to get this done before it requires more time and effort

Markus-Goetz:
Love it!

PR #158, opened 2019-03-12
added exp2 with np.exp2 functionality (not in pytorch)

coquelin77:


Markus-Goetz:
Using numpy only, we would not be able to utilize the GPU backend. Could this work instead?

```
In [94]: exponents = torch.tensor([0, 1, 1.5, 2])

In [95]: torch.pow(2, exponents)
Out[95]: tensor([1.0000, 2.0000, 2.8284, 4.0000])
```

coquelin77:
sorry about that. the short answer is yes but not with the current implementation of the local op. since the torch elements need to be second the code has to be duplicated and look like this:

``python
def exp2(x, out=None):
    """
    Calculate the exponential of all elements in the input array.

    Parameters
    ----------
    x : ht.tensor
        The value for which to compute the 2**p.
    out : ht.tensor or None, optional
        A location in which to store the results. If provided, it must have a broadcastable shape. If not provided
        or set to None, a fresh tensor is allocated.

    Returns
    -------
    exponentials : ht.tensor
        A tensor of the same shape as x, containing the 2**p of each element p in this tensor. If out
        was provided, logarithms is a reference to it.

    Examples
    --------
    >>> ht.exp2(ht.arange(5))
    tensor([ 1.,  2.,  4.,  8., 16.], dtype=torch.float64)
    """
    if not isinstance(x, tensor.tensor):
        raise TypeError(
            'expected x to be a ht.tensor, but was {}'.format(type(x)))
    if out is not None and not isinstance(out, tensor.tensor):
        raise TypeError(
            'expected out to be None or an ht.tensor, but was {}'.format(type(out)))

    # infer the output type of the tensor
    # we need floating point numbers here, due to PyTorch only providing sqrt() implementation for float32/64
    promoted_type = types.promote_types(x.dtype, types.float32)
    torch_type = promoted_type.torch_type()

    # no defined output tensor, return a freshly created one
    if out is None:
        result = torch.pow(2, x._tensor__array.type(torch_type))
        return tensor.tensor(result, x.gshape, promoted_type, x.split, x.device, x.comm)

    # output buffer writing requires a bit more work
    # we need to determine whether the operands are broadcastable and the multiple of the broadcasting
    # reason: manually repetition for each dimension as PyTorch does not conform to numpy's broadcast semantic
    # PyTorch always recreates the input shape and ignores broadcasting/too large buffers
    broadcast_shape = stride_tricks.broadcast_shape(x.lshape, out.lshape)
    padded_shape = (1,) * (len(broadcast_shape) - len(x.lshape)) + x.lshape
    multiples = [int(a / b) for a, b in zip(broadcast_shape, padded_shape)]
    needs_repetition = any(multiple > 1 for multiple in multiples)

    # do an inplace operation into a provided buffer
    casted = x._tensor__array.type(torch_type)
    torch.pow(2, casted.repeat(multiples)
              if needs_repetition else casted, out=out._tensor__array)
    return out
```
this runs smoothly but it is some extra code replication

Markus-Goetz:
How about the following:

def exp2(x, ...):
    def local_exp2(x, ...):
        return torch.pow(2, x, ...)

    return __local_op(local_exp2, x, ...)

You could compare here as well: https://github.com/helmholtz-analytics/heat/blob/52c81fbb7b79ef117808bfccd6e7ec56cd86982b/heat/core/operations.py#L436

coquelin77:
yup that works. thank you!

Issue #9, opened 2018-07-16
Implement data redistribution / distributed binary operations


rainman110:
inside ```Tensor.binop```:

Case requried: Both arrays distributed along different axes.

PR #146, opened 2019-02-27
operations.__reduce_op() now returns a tensor of type ht.bool...

ClaudiaComito:
if reduction_op is one of the 4 MPI boolean operations.

Closes #145

Issue #110, opened 2019-02-12
Fix random.randn to output split tensor


coquelin77:
essentially need to change gshape to lshape for the data creation

coquelin77:
hey Claudia, im seeing some interesting behavior with the argmin, min, and max tests for the split tensor. i just wanted to let you know that it isnt the same data even when the seed is the same @ClaudiaComito

ClaudiaComito:
Hey Daniel, interesting, thanks, I'll look into it. @coquelin77

ClaudiaComito:
OK I see the problem, it is the same data globally, but not necessarily when compared rank by rank, since no preferential split axis is specified for random_data. How to check for equality of the global result? @coquelin77

coquelin77:
when I run a simple test (see below) to just print the output of the randns locally and globally, the values are different. It looks to me like the pytorch randn will only be the same if the generated arrays are the same size. So if you wanted to test argmin you could make a heat tensor of zeros which is split whichever way you want then set the values to the same as those in the non-split tensor. @ClaudiaComito 

```python
torch.manual_seed(1)
random_data = ht.random.randn(3, 3, 3)
torch.manual_seed(1)
random_data_split = ht.random.randn(3, 3, 3, split=0)
print('r', random_data)
print('s', random_data_split)
```

coquelin77:
Open problem with this issue: pytorch's randn is dependent on the size of the tensor which is requested. This makes a problem for reproduce-ability as 
```python
ht.random.randn(3, 3, 3) != ht.random.randn(3, 3, 3, split=0)
```

PR #165, opened 2019-03-12
Features/9 redistribution

Markus-Goetz:
Resolves issue #9

Pull requests implements an (almost complete) MPI wrapper for all functions. Overview of the features:

* Wrapper allows to exchange torch tensors, contiguous and non-contiguous in memory, to be exchanges
* MPI wrapper will take core of creating MPI data types for strided multi-dimensional arrays
* Wrapper "thinks" in entire rows/columns/channels/... rather than the usual MPI bytes.
* Tensor can now be redistributed across nodes. Implementation can be found in tensor.py, method is called **resplit()**
* Working implementation of argmin() (by-product due to not working unittests after rewriting the MPI wrapper)


heat/core/communication.py
```diff
@@ -103,31 +119,420 @@ def chunk(self, shape, split):
...
+        counts = np.full((self.size,), shape[axis] // self.size)
+        counts[:shape[axis] % self.size] += 1
+
+        # the displacements into the buffer
+        displs = np.zeros((self.size,), dtype=counts.dtype)
+        np.cumsum(counts[:-1], out=displs[1:])
+
+        # helper that calculates the output shape for a receiving buffer under the assumption all nodes have an equally
+        # sized input compared to this node
+        output_shape = [ele for ele in shape]
```
coquelin77: list(shape)

Markus-Goetz: Curious I cannot reproduce this, for how many cores? mpirun -np 1 python -m unittest

coquelin77:
mpirun -np 4 pytest

and you just said the the add bug is a known issue.

heat/core/communication.py
ClaudiaComito: I'm late to this so maybe you have solved it already, but we have modified __eq__ to return a ht.bool instead of a torch.uint8, that's why the argmin test fails (#146)

Markus-Goetz: That might be it. I will have a look into it, thanks

coquelin77: i think it's a problem with the logic of the test and not the argmin or all functions: a print shows that the comparison is between 'tensor([[1, 1, 1, 1, 1, 1, 1, 1]])' and  '8' with -np 4.

heat/core/tensor.py
```diff
@@ -876,45 +851,124 @@ def __ne__(self, other):
...
-         """
+        >>> T2 = ht.float32([[3, 3], [2, 2]])
+        >>> T1.__pow__(T2)
+        tensor([[1., 8.],
+             [9., 16.]])
 
+        """
         return arithmetics.pow(self, other)
 
+    def resplit(self, axis=None):
```
Cdebus: This is not working for me locally.
I tried reproducing the example. The tensor.split does change, but lshape stays the same....

heat/core/operations.py
```diff
@@ -20,6 +20,28 @@
 ]
 
 
+def mpi_argmin(a, b, _):
```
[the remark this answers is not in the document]
Markus-Goetz: This actually shouldnt, but the resulting MPI operator probably should. Thanks

heat/core/operations.py
```diff
@@ -510,7 +533,7 @@ def __reduce_op(x, partial_op, reduction_op, axis, out):
     if x.split is not None and (axis is None or axis == x.split):
         split = None
         if x.comm.is_distributed():
-            x.comm.Allreduce(MPI.IN_PLACE, partial[0], reduction_op)
+            x.comm.Allreduce(MPI.IN_PLACE, partial, reduction_op)
```
ClaudiaComito: This doesn't seem to work when distributed and reduction_op is MPI_ARGMIN. Try this test with mpirun -n 2:
```
data = ht.array(ht.random.randn(4, 5), split=0)
result = ht.argmin(data, axis=0)
print('rank {} ht.argmin final result: {}'.format(data.comm.rank, result))
```
Do not set the seed!
The result from Allreduce will only contain min indices from the second half of the tensor (rank 1):

```
tensor([[-1.5692,  0.9294, -0.0859,  0.8452, -1.9167],
        [-0.9213,  0.8887,  0.8188,  1.1889,  0.1289],
        [ 0.5067, -0.6628,  0.7653, -0.6449,  2.5042],
        [ 0.6080, -1.7988,  0.5752,  0.2830,  1.1340]])
tensor([[ 1.7337, -2.8722,  0.5502,  1.9691, -0.3482],
        [ 0.2360,  1.5681, -1.3815,  2.6588, -0.7795],
        [ 0.7685,  0.7748, -1.2191, -0.1385, -0.6562],
        [-0.4852, -0.4101, -0.6436,  0.0545, -0.2059]])
rank 0 ht.argmin final result: tensor([[7, 4, 5, 6, 5]])
rank 1 ht.argmin final result: tensor([[7, 4, 5, 6, 5]])

```
Note that the actual MPI_ARGMIN result is correct:
```
MPI_ARGMIN: rhs.storage:   
-1.5691732168197632
 -2.872173547744751
 -1.3814921379089355
 -0.6449006795883179
 -1.9167150259017944
 0.0
 4.0
 5.0
 2.0
 0.0
```

The tests didn't catch it because  we've set torch.manual_seed(1), so the tensors on each rank are identical, and the result (all min indices from highest rank) isn't that suspicious.

coquelin77: Is this related to the problem with randn? (#110) since currently randn will simply create a tensor on each node that is the same size (the given size). 

what I'm concerned about is this: the tensors generated by randn are not created correctly.

ClaudiaComito: Hi Daniel,

but in this case the random tensor is "wrapped" in a ht.array with split=0 and that behaves as expected, as far as I can tell.

heat/core/communication.py
Markus-Goetz: Issue has been resolved by a fix, thanks for the remark!

heat/core/arithmetics.py
```diff
@@ -52,165 +53,158 @@ def add(t1, t2):
...
+    as argument.
 
-        Parameters
-        ----------
-        t1: tensor or scalar
-        The first operand whose values are divided
+    Parameters
+    ----------
+    t1: tensor or scalar
+    T   he first operand whose values are divided
```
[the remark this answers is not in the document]
Markus-Goetz: Thanks, fixed

heat/core/operations.py
```diff
@@ -165,21 +188,22 @@ def argmin(x, axis=None, out=None):
...
 
-        out = torch.reshape(torch.argmin(x._tensor__array), (1,))
-        return tensor.tensor(out, out.shape, types.canonical_heat_type(out.dtype), None, x.device, x.comm)
+    # perform the global reduction
+    reduced_result = __reduce_op(x, local_argmin, MPI_ARGMIN, axis, out)
 
-    out = __reduce_op(x, torch.min, MPI.MIN, axis, out=None)._tensor__array[1]
+    # correct the tensor
+    reduced_result._tensor__array = reduced_result._tensor__array.chunk(2)[-1].type(torch.int64)
+    reduced_result._tensor__dtype = types.int64
```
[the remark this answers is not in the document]
Markus-Goetz: Indeed you are correct

heat/core/operations.py
Markus-Goetz: I will have a look into it.

heat/core/operations.py
```diff
@@ -178,20 +202,23 @@ def argmin(x, axis=None, out=None):
...
 
-    return tensor.tensor(out, out.shape, types.canonical_heat_type(out.dtype), x.split, x.device, x.comm)
+    return reduced_result
```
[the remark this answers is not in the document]
Markus-Goetz: Let me have a look, good remark, thanks!

heat/core/operations.py
Markus-Goetz: In fact, the MPI_ARGMIN implementation seems to be incorrect. It has nothing to do with randn() in any way. Consider this:

`
# foo.py
import heat as ht

data = ht.array(ht.random.randn(4, 5), split=0)
print(data, data.shape, data.lshape)
print(ht.argmin(data, axis=0))
`

And:

`
mpirun --tag-output -np 2 python foo.py

[1,0]<stdout>:tensor([[ 0.3886,  0.7388,  1.7783,  0.6372, -1.3354],
[1,0]<stdout>:        [-0.3796,  1.6146, -1.5172,  1.7657,  0.0242],
[1,0]<stdout>:        [ 0.1005, -0.8079,  0.4624,  0.8972, -0.5272],
[1,0]<stdout>:        [ 0.2935,  0.3631,  2.3646,  1.1830, -0.0578]]) (8, 5) (4, 5)
[1,1]<stdout>:tensor([[-1.4479, -0.6387,  0.4102,  1.0344, -0.0191],
[1,1]<stdout>:        [-0.4462, -1.2580,  1.4736,  0.3634,  0.2449],
[1,1]<stdout>:        [-0.3439,  1.1679, -1.0671, -1.2921,  1.1171],
[1,1]<stdout>:        [-1.7884,  0.1019, -1.3534, -0.2997,  0.4900]]) (8, 5) (4, 5)
[1,0]<stdout>:tensor([[-1.7884, -1.2580, -1.3534, -1.2921, -0.0191],
[1,0]<stdout>:        [ 7.0000,  5.0000,  7.0000,  6.0000,  4.0000]], dtype=torch.float64)
[1,1]<stdout>:tensor([[-1.7884, -1.2580, -1.3534, -1.2921, -0.0191],
[1,1]<stdout>:        [ 7.0000,  5.0000,  7.0000,  6.0000,  4.0000]], dtype=torch.float64)
[1,0]<stdout>:tensor([[7, 5, 7, 6, 4]])
[1,1]<stdout>:tensor([[7, 5, 7, 6, 4]])
`

In the lines 9-12 you the output of the reduction result from MPI_ARGMIN. For the third column (index=2) you already see that the reduction result is wrong. -1.3534 is obviously not smaller than -1.5172.

ClaudiaComito: Hi Markus, glad you can reproduce it.

I find that rhs.storage within MPI_ARGMIN contains the correct result (see above). So I'm not sure if it's a problem with MPI_ARGMIN or with the Allreduce call.

Markus-Goetz: The issue is the write back into the buffer

`
rhs.storage().copy_(result.storage())
`

This needs to be fixed

Markus-Goetz: Provided fix, to be found in MPI_ARGMIN

heat/core/operations.py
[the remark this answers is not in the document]
Markus-Goetz: Was indeed an error in the implementation. Provided a fix

heat/core/operations.py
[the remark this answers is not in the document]
Markus-Goetz: Provided fix for this

Issue #24, opened 2018-07-16
Implement INF, NaN


bhagemeier:

Cdebus:
I was planning to implement Inf and NaN as tunnels for float('inf') and float('NaN'), the IEEE 754 implementations of those values. I could not find anything on the distinction between CPU and GPU for this standard, but I'm not an expert on this.
@Markus-Goetz what do you think?

PR #186, opened 2019-03-19
24 implementations of constants

Cdebus:
Solves #24 , #123 and #132 
- Implementation of Constants inf, nan, pi and e
- No distinction between GPU and CPU:
   - pi and e are "hard-coded" as floating precision (in analogy to numpy)
   - inf an nan are tunneled from float('inf') and float('nan'), the IEEE 754 of infinity and not a number (should be supported by NVIDIA ABD CUDA9

### DIFF
--- a/heat/core/__init__.py
+++ b/heat/core/__init__.py
@@ -1,4 +1,5 @@
 from .communication import *
+from .constants import *
 from .types import *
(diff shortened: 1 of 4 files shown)


Markus-Goetz:
Superb

Issue #153, opened 2019-03-07
Implement local item getters/setters


Markus-Goetz:
Implementation strategy could be to use a proxy object for this, i.e.:

a = ht.zeros(100, split=0)
a.loc[0:20]

Issue #152, opened 2019-03-07
Implement feature complete global item setter


Markus-Goetz:

coquelin77:
i propose that the global setter/getter functions should be the __getitem__ and __setitem__ functions to have the functionality of x[a : b]. 

also this is tied to issue #107 in that the __setitem__ can currently only accept heat tensors or scalars

PR #188, opened 2019-03-20
Features/global setters+getters

coquelin77:
added global setters/getters as well as lloc for local setting/getting

issues #151 #152 and #153


coquelin77:
i had issues getting the code coverage to work with multiple processes. it gave me MPI errors so i changed it back. but that causes the coverage of the set/getitem functions to decrease. However, these cases **are** covered during the pytest call

ClaudiaComito:
Hi Daniel,

it's great to have these available! 

Just a general question: how do we want to deal with those get/setitem calls that would run fine on a single process, but throw an IndexError when distributed? Silly example:

```
a = ht.zeros((7, 5), split=1)
print(a.comm.rank, a[1:4, 4])

```

As a user I'd like the library to figure out if the calculation can be distributed, and if not, that it should be run locally instead of refusing to proceed. I'm sure at the moment we have this issue with most other functions though.

coquelin77:
@ClaudiaComito you are totally correct. I had missed some cases during my initial writing/testing. With the latest commit all of the standard commands should work with any split axis

coquelin77:
travis fixes (number of processes issue) and I also discovered a bug for tensors with >2 dimensions

Markus-Goetz:
Bump, requires merge of master

Markus-Goetz:
Awesome stuff, works really well. I have only two minor things I would like to have seen changed

1. Allow arbitrary iterables to be used in setitem. Consider this:

```
import heat as ht

# before
a = ht.ones((4,5,), split=0).tril()
print(a)
# after
a[0] = ht.full((1,5,), 6)
print(a)
```

prints 

```
[1,0]<stdout>:tensor([[1., 0., 0., 0., 0.],
[1,0]<stdout>:        [1., 1., 0., 0., 0.]])
[1,1]<stdout>:tensor([[1., 1., 1., 0., 0.],
[1,1]<stdout>:        [1., 1., 1., 1., 0.]])

[1,0]<stdout>:tensor([[6., 6., 6., 6., 6.],
[1,0]<stdout>:        [1., 1., 0., 0., 0.]])
[1,1]<stdout>:tensor([[1., 1., 1., 0., 0.],
[1,1]<stdout>:        [1., 1., 1., 1., 0.]])
```

What I would expect from the function. However, if I do this:

```
import heat as ht

# before
a = ht.ones((4,5,), split=0).tril()
print(a)
# after
a[0] = [6, 6, 6, 6, 6]
print(a)
```

I will get a `NotImplementedError`. This should not be much of a change.


2. I am not sure the global indexing behaviour works as I would expect it (up for debate). Consider this:

```
import heat as ht

a = ht.ones((4,5,), split=0).tril()
print(a[0], a.split)
```

results in 

```
[1,0]<stdout>:tensor([1., 0., 0., 0., 0.]) 0
[1,1]<stdout>:None 0
```

My expectation would rather be that the result of global indexing operation is a copy of said item on each node (i.e. the same tensor without split):


```
[1,0]<stdout>:tensor([1., 0., 0., 0., 0.]) None
[1,1]<stdout>:tensor([1., 0., 0., 0., 0.]) None
```

What are your thoughts>

coquelin77:
1 is solved for lists, tuples and ndarrays. more can be added easily.

2 it is also a quick fix (an allreduce if the result is only on one node). But im not sure if we should do it. if it is implemented that way it would only happen if a slice it only on one node. I can see it becoming a problem for users because it would not be a common case for the extremely large data sets which we are targeting. 

I see slicing being used before a rebalance/redistribute function by the user. I did add a warning so that the user will know that the slice created is not on all processes. My concern is that it can get dicey if we do too much rebalancing.

We will need a rebalancing function but im not sure if we should include that here or if we should recommend using it after the slice is returned.

Markus-Goetz:
1. Thanks!

2. I see your point and would agree with you, especially for large data amounts. However, we will often just delay the data redistribution/balancing to the very next operation. Consider this

```
import heat as ht

a = ht.random.randn(15, 3, split=0)
b = a[0::2]

print(b.lshape)
```

should print (it currently does not there is a bug I annotated in the review):

```
[3,0]<stdout>:(3, 3,)
[3,1]<stdout>:(2, 3,)
[3,2]<stdout>:(3, 3,)
```

Meaning that the innermost data chunk is smaller in shape than one that is assigned to a higher rank. This means that all the assumptions made in the code about the chunk distribution are not true. Say you would want to add a 0-split tensor (e.g. ht.zeros(..., split=0)) to the above sliced, called `b`, you will get mismatching shape errors, as the local chunk sizes are imbalanced. Ideas on how to deal with this?

heat/core/tensor.py
```diff
@@ -1426,28 +1482,230 @@ def __repr__(self, *args):
...
+                # if the given axes are not splits (must be ints for python)
+                elif key[self.split] in range(chunk_start, chunk_end):
+                    key = list(key)
+                    key[self.split] = key[self.split] - chunk_start
+                    try:
+                        return tensor(self.__array[tuple(key)], tuple(self.__array[tuple(key)].shape), self.dtype, self.split, self.device, self.comm)
+                    except ValueError:  # case of returning just one value
+                        return self.__array[tuple(key)]
+                else:
+
```
Markus-Goetz: Stepped slicing is currently bugged. Try this example:

```
import heat as ht

a = ht.zeros((15,3,), split=0)
b = a[::2]

print(b.lshape)
```

will result in (mpirun -np 3):

```
[1,2]<stderr>:Traceback (most recent call last):
[1,2]<stderr>:  File "foo.py", line 4, in <module>
[1,2]<stderr>:    b = a[::2]
[1,2]<stderr>:  File "/home/mgoetz/Documents/heat/heat/core/tensor.py", line 1590, in __getitem__
[1,2]<stderr>:    key_set = set(range(key.start, key.stop, key.step if key.step else 1))
[1,2]<stderr>:TypeError: 'NoneType' object cannot be interpreted as an integer
[1,1]<stderr>:Traceback (most recent call last):
[1,1]<stderr>:  File "foo.py", line 4, in <module>
[1,1]<stderr>:    b = a[::2]
[1,1]<stderr>:  File "/home/mgoetz/Documents/heat/heat/core/tensor.py", line 1590, in __getitem__
[1,1]<stderr>:    key_set = set(range(key.start, key.stop, key.step if key.step else 1))
[1,1]<stderr>:TypeError: 'NoneType' object cannot be interpreted as an integer
[1,0]<stderr>:Traceback (most recent call last):
[1,0]<stderr>:  File "foo.py", line 4, in <module>
[1,0]<stderr>:    b = a[::2]
[1,0]<stderr>:  File "/home/mgoetz/Documents/heat/heat/core/tensor.py", line 1590, in __getitem__
[1,0]<stderr>:    key_set = set(range(key.start, key.stop, key.step if key.step else 1))
[1,0]<stderr>:TypeError: 'NoneType' object cannot be interpreted as an integer
```

coquelin77: fixed in latest commit. i thought i hit all the test cases but i guess i missed that one

coquelin77:
>     1. Thanks!
> 
>     2. I see your point and would agree with you, especially for large data amounts. However, we will often just delay the data redistribution/balancing to the very next operation. Consider this
> 
> 
> ```
> import heat as ht
> 
> a = ht.random.randn(15, 3, split=0)
> b = a[0::2]
> 
> print(b.lshape)
> ```
> 
> should print (it currently does not there is a bug I annotated in the review):
> 
> ```
> [3,0]<stdout>:(3, 3,)
> [3,1]<stdout>:(2, 3,)
> [3,2]<stdout>:(3, 3,)
> ```
> 
> Meaning that the innermost data chunk is smaller in shape than one that is assigned to a higher rank. This means that all the assumptions made in the code about the chunk distribution are not true. Say you would want to add a 0-split tensor (e.g. ht.zeros(..., split=0)) to the above sliced, called `b`, you will get mismatching shape errors, as the local chunk sizes are imbalanced. Ideas on how to deal with this?

yes this is a problem. My idea would be to have a balance function. However, this function would *not* do a full redistribute. It would instead "push" the data from a node to its neighbors so as to keep all of the data in the proper indices. This strategy would only be really beneficial if only a couple axes need to be communicated. In the cases where the data is more lopsided it would be better to do something like a full redistribute. thoughts?

Markus-Goetz:
Yeah I think this is something that we can agree on. So plan of action would be:

1. In this PR, do not assign None to the local tensor slice, but rather an empty (shape=(,)) tensor as __array of correct torch type.

2. Add an issue that balance() needs to be put into the getters after the balance() issue has been tackled.

coquelin77:
> Yeah I think this is something that we can agree on. So plan of action would be:
> 
>     1. In this PR, do not assign None to the local tensor slice, but rather an empty (shape=(,)) tensor as __array of correct torch type.
> 
>     2. Add an issue that balance() needs to be put into the getters after the balance() issue has been tackled.

1. are you opposed to tuple([0 for _ in self.gshape]) for the lshape? 
2. will do

Markus-Goetz:
I am opposed to this

```
import heat as ht

a = ht.ones((4,5,), split=0).tril()
print(a[0], a.split)
```

results in

```
[1,0]<stdout>:tensor([1., 0., 0., 0., 0.]) 0
[1,1]<stdout>:None 0
```

I would expect that rank 1 would have an empty tensor, i.e.

```
[1,0]<stdout>:tensor([1., 0., 0., 0., 0.]) 0
[1,1]<stdout>:tensor() 0
```

coquelin77:
yeah i get what you are saying. i meant that the lshape of the returned tensor on the nodes without data will not be (, ) but it will be (0, ). this was what i was referring to.

i.e.:
```python
a = ht.ones((15, 3,), split=0)
print(a.comm.rank, a[0:8:2], a[0:8:2].lshape)
>>> (3, 2) tensor([]) (0,)
>>> (3, 1) tensor([[1., 1., 1.]]) (1, 3)
>>> (3, 0) tensor([[1., 1., 1.],
                   [1., 1., 1.],
                   [1., 1., 1.]]) (3, 3)
```

heat/core/tensor.py
```diff
@@ -55,6 +72,45 @@ def split(self):
...
+    def lloc(self):
+        """
+        Local item setter and getter. i.e. this function operates on a local level and only on the pytorch tesnors composing the ht tensor
+        This function uses the LocalIndex class
+
+        Parameters
+        ----------
+        key : int, slice, list, tuple
+            indices of the desired data
+
```
Markus-Goetz: Remove line break

heat/core/tensor.py
```diff
@@ -1513,24 +1569,254 @@ def __repr__(self, *args):
...
+        Parametes
+        ---------
+        key : int, slice, tuple, list
+            indices to get from the tensor.
+
+        Returns
+        -------
+        result : ht.tensor
+            getter returns a new ht.tensor composed of the elements of the original tensor selected by the indices given.
+            this does *NOT* redistribute or rebalance the resulting tensor. If the selection of values is unbalanced then
```
Markus-Goetz: Could you please add this to the balance() issue as a note, that getters and setters need to be rebalanced OR open a separate issue referencing balance()

heat/core/tensor.py
```diff
@@ -1513,24 +1569,254 @@ def __repr__(self, *args):
...
+                            start = key[s].start if key[s].start is not None else 0
+                            stop = key[s].stop if key[s].stop is not None else self.gshape[s]
+                            if start == stop - 1:
+                                gout = gout[:-1]
+
+                if self.split >= len(gout):
+                    new_split = len(gout) - 1 if len(gout) - 1 > 0 else 0
+                else:
+                    new_split = self.split
+                # all_slices = True if all(slices) else False
```
Markus-Goetz: please remove this line

heat/core/tensor.py
```diff
@@ -13,6 +14,22 @@
 from . import rounding
 from . import reductions
 
+warnings.simplefilter('always', ResourceWarning)
```
Markus-Goetz: This seems like a very strong filter.

Why is this here? Can we filter this warning precisely where it happens, e.g. as a decorator around the function that produces the warning?

heat/core/tensor.py
```diff
@@ -1513,24 +1569,254 @@ def __repr__(self, *args):
...
+                arr = self.__array[key]
+
+            for e, _ in enumerate(gout):
+                if e == new_split:
+                    gout[e] = self.comm.allreduce(gout[e], MPI.SUM)
+                else:
+                    gout[e] = self.comm.allreduce(gout[e], MPI.MAX)
+
+            return tensor(arr, gout if isinstance(gout, tuple) else tuple(gout), self.dtype, new_split, self.device, self.comm)
+        else:
```
Markus-Goetz: can we have this as early-out up top?

heat/core/tests/test_tensor.py
```diff
@@ -1,6 +1,7 @@
 import torch
 import unittest
 
+import numpy as np
```
Markus-Goetz: please have this up to before `import torch`

PR #189, opened 2019-03-20
Bug/73 broadcast of   binop

krajsek:
This pull request will solve several issues in situations where there are more MPI ranks than the range of the spit dimension.

In particular:

Commit will fix the case tensor.comm.size>shape[tensor.split]  in function  __binary_op
in file operations.py

The original   __binary_op  will fail in case of broadcasing 
when there are ranks with empty tensors. In this case tensors are 
fetched from the the rank 0 process if needed. In case a rank contains an empty
tensor the operation will be skipped.
 
Commit will fix load_hdf5 for  tensor.comm.size>shape[tensor.split] in io.py

### DIFF
--- a/heat/core/io.py
+++ b/heat/core/io.py
@@ -1,7 +1,7 @@
 import os.path
 import torch
 import warnings
-
+from .stride_tricks import sanitize_axis
 from .communication import MPI, MPI_WORLD
 from . import devices
 from . import types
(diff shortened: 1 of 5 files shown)


Cdebus:
Hey, I also ran across this problem (in kmeans), thanks for the fix!

Markus-Goetz:
Bump, requires merge of master

Markus-Goetz:
Looks good to me.

Do we want to keep the redistribution warning where it is right now, or move it to resplit()?

krajsek:
> 
> 
> Looks good to me.
> 
> Do we want to keep the redistribution warning where it is right now, or move it to resplit()?

I would keep it there.

Issue #155, opened 2019-03-11
Fix the output shape of reduction operations, discard singular reduction axis


Markus-Goetz:
At the moment, the generic __reduce_op implementation maintains the number of dimensions in the output shape of the resulting tensor, e.g.:

```
In [16]: a = ht.zeros((3,4,5))

In [17]: a.sum(axis=1).shape
Out[17]: (3, 1, 5)
```
With the (here: second) axis set to 1. The proper behaviour should be that the axis disappear during the reduction, e.g.:

```
In [18]: a = np.zeros((3,4,5))

In [19]: a.sum(axis=1).shape
Out[19]: (3, 5)
```

PR #190, opened 2019-03-20
Introduced kwarg keepdim=False for reduction operations. Solves issue  #155

ClaudiaComito:
The kmeans implementation seems to rely on keeping the reduction axis and I don't want to open that can of worms for the time being, so I opted for introducing a keepdim argument (default = False) for operations.__reduce_op().

All relevant methods, functions and tests adjusted accordingly.


heat/core/operations.py
```diff
@@ -146,15 +147,17 @@ def allclose(x, y, rtol=1e-05, atol=1e-08, equal_nan=False):
...
     if not isinstance(y, tensor.tensor):
-        raise TypeError('Expected y to be a ht.tensor, but was {}'.format(type(y)))
+        raise TypeError(
+            'Expected y to be a ht.tensor, but was {}'.format(type(y)))
 
     return torch.allclose(x._tensor__array, y._tensor__array, rtol, atol, equal_nan)
 
 
-def argmin(x, axis=None, out=None):
+def argmin(x, axis=None, keepdim=False, out=None):
```
Markus-Goetz: keepdim is assumed to be auto-false if not set, so the parameter can be removed

heat/core/operations.py
```diff
@@ -197,10 +200,12 @@ def local_argmin(*args, **kwargs):
         return torch.cat([minimums.double(), indices.double()])
 
     # perform the global reduction
-    reduced_result = __reduce_op(x, local_argmin, MPI_ARGMIN, axis, out)
+    reduced_result = __reduce_op(
+        x, local_argmin, MPI_ARGMIN, axis, keepdim, out)
```
Markus-Goetz: We should configure your IDE to not break these lines :D

heat/core/operations.py
```diff
@@ -519,10 +527,11 @@ def __reduce_op(x, partial_op, reduction_op, axis, out):
 
     if axis is None:
         partial = partial_op(x._tensor__array).reshape((1,))
-        output_shape = (1,)
+        output_shape = (1,) if keepdim else ()
```
Markus-Goetz: I do not understand this line. Could you please explain it to me? I assume it is for scalars. Should the output shape not remain to be (1,) as it is exactly one element in one dimension?

heat/core/operations.py
coquelin77: its probably in the setting under the name auto wrap or auto break or something like that

heat/core/operations.py
```diff
@@ -519,10 +527,11 @@ def __reduce_op(x, partial_op, reduction_op, axis, out):
...
     else:
         partial = partial_op(x._tensor__array, dim=axis, keepdim=True)
```
coquelin77: should this be changed to keepdim=keepdim?

heat/core/operations.py
ClaudiaComito: :D @Markus-Goetz alright, We (pluralis majestatis) have configured now.

heat/core/operations.py
ClaudiaComito: Yes it's a scalar, that's the shape of a numpy reduction when axis=None, but of course numpy does output scalars. I suppose since a heat reduction outputs a tensor, we could stick to (1,). Is that your preference?

heat/core/operations.py
ClaudiaComito: No, my understanding is that we definitely want to keep the dimension until after the MPI collective has run.

coquelin77: okay sounds good

heat/core/operations.py
ClaudiaComito: Hey Markus, I don't get it. If I remove the parameter from the definition, I can't use it in the __reduce_op call later.

heat/core/operations.py
Markus-Goetz: Great! (I laughed my ass off here about the pluralis majestatis). Thanks Claudia!

heat/core/operations.py
Markus-Goetz: My point being: keepdim should not appear in the function signature of these calls here as they are also not part of the numpy API (see here: https://docs.scipy.org/doc/numpy/reference/generated/numpy.argmin.html)

Beyond that, most (all?) calls in PyTorch assume keepdim=False if keepdim is not explicitly set. Practically for implementation reasons you could do the following:

`
def argmin(x, axis=None):
    return __reduce_op(..., keepdim=False)

def __reduce_op(x, local_func, **kwargs):
    # access kwargs explicitly
    print(kwargs['keepdim'])
    # or even
    local_func(**kwargs) # pass whatever is in kwargs onward to local_func
`

Does that make sense or am I confused?

ClaudiaComito: OK fixed it.

Markus-Goetz:
Two things I noticed:

1. Your Royal Highnesses IDE is still doing this line break thing. :)

2. Most of the operations now only have signature of `<function>(operand, **kwargs)` instead of previously `<function>(operand, axis=None, out=None)`. Why exactly did the explicit argument have to disappear? I am okay with `**kwargs` for the implementation of `__reduce_op`, but the user function signatures should maintain their explicit parameters as much as possible so that the documentation makes sense and we get meaningful error messages.

ClaudiaComito:
Markus,
1. the line breaks were probably leftovers from previous implementations. I removed as many as I could find now, but there might still be some that I haven't noticed. Let me know if you find more.

2. you're right about the explicit arguments. I left the __reduce_op implementation with **kwargs, and left the **kwargs option open for argmin and argmax. These two are the only reduction operations that don't have a keepdim argument in numpy, but they do have it in torch and indeed we're using argmin with keepdim=True in kmeans.fit. All the other reduction operations have gone back to the original form with explicit arguments. 

Of course we can decide to never use keepdim for argmin or argmax because numpy doesn't offer it, and fix kmeans.fit accordingly.

PR #191, opened 2019-03-21
Fix for test_add

### LINKED ISSUE #164: test_add will fail with a process count >= 3

Cdebus:
Fixes #164 
- Error was result of uncomplete cleanup after restructuring of binary operator generic function
- removal of faulty tensor
- tests on NotImplementedError raise are performed with T_s

### DIFF
--- a/heat/core/tests/test_arithmetics.py
+++ b/heat/core/tests/test_arithmetics.py
@@ -16,8 +16,7 @@ T1 = ht.float32([
 ])
 v = ht.float32([2, 2])
 v2 = ht.float32([2, 2, 2])
-T_s = ht.tensor(T1._tensor__array, T1.shape, T1.dtype, 0, None, None)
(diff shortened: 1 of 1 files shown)


coquelin77:
everything passes here with up to 4 processes.

I was looking through the test though and I found a few stylistic things. Such as I didnt see why when you copy the T1 tensor to form T_s you say you want comm=None. I changed that to T1.comm and I also changed the device to T1.device. I also did some pep8 formating in the two lines following.

```python
T_s = ht.tensor(T1._tensor__array, T1.shape, T1.dtype, 0, T1.device, T1.comm)
Ts = ht.ones((2, 2), split=1)
otherType = (2, 2)
```

Cdebus:
Well, to be honest, there is no reason^^.... I didn't think about this at the time, I only cared for it to be a split tensor. But it does make a lot more sense to copy the device and communication. Thanks for catching it!

PR #196, opened 2019-04-01
Implemented mod()

TheSlimvReal:
Targets issue #124.

### DIFF
--- a/heat/core/arithmetics.py
+++ b/heat/core/arithmetics.py
@@ -9,7 +9,8 @@ __all__ = [
     'div',
     'mul',
     'pow',
-    'sub'
+    'sub',
+    'mod'
 ]
 
 def add(t1, t2):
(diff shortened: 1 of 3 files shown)


heat/core/arithmetics.py
```diff
@@ -9,7 +9,8 @@
     'div',
     'mul',
     'pow',
-    'sub'
+    'sub',
+    'mod'
```
Markus-Goetz: Please sort this alphabetically

heat/core/tests/test_arithmetics.py
```diff
@@ -125,6 +124,47 @@ def test_div(self):
         with self.assertRaises(TypeError):
             ht.div('T', 's')
 
+    def test_mod(self):
```
Markus-Goetz: Please provide a test case with a non-zero divisor (e.g. a % 2.4)

coquelin77:
should this be 'fmod' instead of 'mod'? I dont see a huge difference between the two but i think we should stick with the numpy naming scheme. also from their docs:

"This is the NumPy implementation of the C library function fmod, the remainder has the same sign as the dividend x1. It is equivalent to the Matlab(TM) rem function and should not be confused with the Python modulus operator x1 % x2."

TheSlimvReal:
> should this be 'fmod' instead of 'mod'? I dont see a huge difference between the two but i think we should stick with the numpy naming scheme. also from their docs:
> 
> "This is the NumPy implementation of the C library function fmod, the remainder has the same sign as the dividend x1. It is equivalent to the Matlab(TM) rem function and should not be confused with the Python modulus operator x1 % x2."
> 
> i guess my point is that i believe that if we implement a wrapper for a torch function in heat it should have the same name as the torch function

I understand the naming point. But would you then suggest to remove the __mod__ implementation because torch only offers the fmod functionality?

Markus-Goetz:
For all intents and purposes mod() and fmod() are currently aliases. fmod() converts the input tensor beforehand to a double-precision floating-point number, whereas mod not necessarily does that. Since we do not have an alternative implementation for mod() both are the same.

Issue #168, opened 2019-03-12
Provide implementation of right-hand side binary operations


Markus-Goetz:
At the moment it is possible to do the following for example:

```
a = ht.zeros((5, 3))
a + 1
```

However, it is impossible to do:

```
a = ht.zeros((5, 3))
1 + a
```
HeAT currently does not provide implementation for so-called right-hand side operations, e.g. `__radd__` etc.

Issue #167, opened 2019-03-12
Implement tensor redistribution for binary operations


Markus-Goetz:
After PR #165 is merged, implement binary operators for two tensor with set, but different split axis (e.g. split=0 and split=1).

Think of a rule to decide which tensor should be resplit. Proposal (might require modification): 

1. attempt to redistribute tensors with fewer elements
2. should both have the same amount of elements, redistribute tensor that are split along a minor axis, i.e. split >= 1.
3. should both have the same number of elements and both have a minor axis, redistribute the one with larger minor axis

--- 
**EDITED July 31st 2023 by @ClaudiaComito**
Here's the current status of binary operations `heat.core._operations.__binary_op()`:

- [x] ensure that operand types are promoted correctly
- [x] ensure that binary ops are possible between operands split along **the same axis** but with misaligned `lshape_maps` (e.g. as a result of indexing)
- [ ] ensure that operands are on the same MPI communicator
- [ ] ensure that binary ops are possible between operands split along **different axes**

---
Reviewed within #1109

Cdebus:
@Markus-Goetz I will try out these rules. However, what happens in 2) if none of the tensors is split among a minor axis?

Markus-Goetz:
In this case you may simply add them locally as already implemented and working

Cdebus:
No, I mean if they both have a split > 1

Markus-Goetz:
That would be rule 3.) if both have a minor axis, than the one with the larger minor axis will be redistributed

Markus-Goetz:
Couple more things that came to my mind that have to be considered:

* __binary_op should ensure that the comms of the two operands match. We cannot perform the operation if the data is distributed across different comms. Exception: the split of one of the operands is `None`. Local data can always be used

* Ensure that the operands types are properly promoted. See also `types.promote_types`.

Cdebus:
Why was that closed? Issue is not fixed, PR #268 bumped due to issues with MPI communication and resplit

coquelin77:
My apologies. i think I saw that the PR was closed (at the time it was a month after the bump, and i think that you had closed the PR). I am not sure but I think that I had thought that the PR was merged and then it was a stay issue which was not closed at the time of merging. Sorry @Cdebus

Cdebus:
No worries ^^
I was just surprised, because fixing the _binary_op was my original intention but didn't work due to resplit due to allgather, and that's how I ended up with that mess

[...]

PR #204, opened 2019-04-04
Implementing right hand side operations

TheSlimvReal:
Target #168.

Added  right-hand-side implementations for all implemented operations like "/", "%", ...


heat/core/tests/test_tensor.py
```diff
@@ -735,6 +737,40 @@ def test_empty_like(self):
...
+        operators = (
+            ('__add__', operator.add, True),
+            ('__sub__', operator.sub, False),
+            ('__mul__', operator.mul, True),
+            ('__truediv__', operator.truediv, False),
+            ('__floordiv__', operator.floordiv, False),
+            ('__mod__', operator.mod, False),
+            ('__pow__', operator.pow, False)
+        )
+        tensor = ht.float32([[1, 4], [2, 3]])
```
Markus-Goetz: We need test cases here for split tensors

heat/core/tensor.py
```diff
@@ -1484,6 +1589,25 @@ def __setitem__(self, key, value):
...
+    tensor([[2., 3.],
+            [4., 5.]])
+    >>> 1 + T
+    tensor([[2., 3.],
+        [4., 5.]])
+    """
+    __radd__ = __add__
+    __rmul__ = __mul__
+
+    # __rfloordiv__  //  TODO: Implement me when implementing __floordiv__
```
Markus-Goetz: Please provide both implementations (regular and right-hand side floor-div). Straight forward by adapting div (torch supports truediv as well)

heat/core/tests/test_tensor.py
Cdebus: This is related to #167, which needs to be fixed first
At the moment, binary operations do not work properly for split tensors.

heat/core/tensor.py
```diff
@@ -451,41 +451,93 @@ def __truediv__(self, other):
         """
         return arithmetics.div(self, other)
 
+    def __rtruediv__(self, other):
```
Cdebus: Just a technicality, but shouldn't the functions be sorted alphabetically? (at least for the operators before it is the case)
EDIT: This goes for all the operators below

Markus-Goetz:
Bump, please update to the current master, would want to merge :)

PR #215, opened 2019-04-09
Implement any()

TheSlimvReal:
Targets issue #140.


heat/core/operations.py
```diff
@@ -166,6 +168,51 @@ def allclose(x, y, rtol=1e-05, atol=1e-08, equal_nan=False):
...
+        The tensor on which any should be calculated
+    axis : int, optional
+        Defines along which axis any should be calculated. If not specified, any will be computed on the whole
+        tensor.
+    out : tensor, optional
+        An output tensor into which the result will be copied. Result is a tensor with dtype=uint8/bool.
+
+    Returns:
+    --------
+    boolean_tensor : tensor of type uint8/bool
```
Markus-Goetz: type is ht.bool (even if internally a uint8). Please update the doc string accordingly

heat/core/operations.py
```diff
@@ -166,6 +168,51 @@ def allclose(x, y, rtol=1e-05, atol=1e-08, equal_nan=False):
     return torch.allclose(x._tensor__array, y._tensor__array, rtol, atol, equal_nan)
 
 
+def any(x, axis=None, out=None):
+    """
+    Returns a boolean indicating if any elements are non-zero.
```
Markus-Goetz: Docstring is generally good, but I would personally prefer it to be closer to the numpy one. See here:

https://docs.scipy.org/doc/numpy/reference/generated/numpy.any.html

PR #219, opened 2019-04-09
Provides bug fixes for: 

Markus-Goetz:
* loading netcdf data files with empty chunks
* eye() unittests with non-iterable offsets


coquelin77:
why did you comment out all of the load netcdf tests? its great that it works with split=-1 but i think the other tests should still be in there

Markus-Goetz:
Good point, simple oversight. Will fix

PR #220, opened 2019-04-09
Features/67 tuple axis

ClaudiaComito:
Looking to address Issue #67


Markus-Goetz:
Small merge conflict with master. Bump

ClaudiaComito:
Merged latest master.

ClaudiaComito:
Oh yes, that looks bad! I'll look into it.

ClaudiaComito:
OK thanks for catching this Markus! Allreduce wasn't being triggered if axis was a tuple because the condition (axis == x.split) was never verified. I've fixed this now, fixed the tests, and added a check to test_sum to verify that Allreduce is triggered.

> I guess the expected solution would be:
> 
> ```
> mpirun --tag-output -np 2 python foo.py
> [1,0]<stdout>:tensor([[[12., 12., 12., 12., 12.]]]) (5,) (5,) None
> [1,1]<stdout>:tensor([[[12., 12., 12., 12., 12.]]]) (5,) (5,) None
> ```

This is what I get now:
```
tensor([[[12., 12., 12., 12., 12.]]]) (5,) (1, 1, 5) None
tensor([[[12., 12., 12., 12., 12.]]]) (5,) (1, 1, 5) None
```

We've got keepdim=True by default when calculating partial locally, so I do expect lshape to keep the reduced dimensions. Should that be changed?

Markus-Goetz:
This is awesome. Thanks for fixing it. 

Yes, I do think the local shape should be the same as the gshape in these cases. By default numpy does not do a keepdim (even though we wrongly did so initially). I think having a keepdim=False default should be generally the way to go, also for the other operations.

ClaudiaComito:
> 
> Yes, I do think the local shape should be the same as the gshape in these cases. By default numpy does not do a keepdim (even though we wrongly did so initially). I think having a keepdim=False default should be generally the way to go, also for the other operations.


I suggest we make this into a separate Issue if you don't mind. It'll break quite a few tests :)

Markus-Goetz:
Fair enough. Would you please create it and reference it here as soon as you did? Beyond that, I will start the review again :).

ClaudiaComito:
> Fair enough. Would you please create it and reference it here as soon as you did? Beyond that, I will start the review again :).

Done, see above. Thanks!

Issue #163, opened 2019-03-12
split reductions into sub-files


coquelin77:
numeric_reductions: mean, var, sum, etc.
logic_reduction: all, argmin, min, max, etc.

Markus-Goetz:
After further discussion in Mattermost, it was suggested to further enhance the above proposal as follow:

numeric_reductions: sum, prod, ....
statistics: mean, argmin, min, max, var, median, ...
logic_reductions: all, allclose, any, ...

PR #230, opened 2019-04-11
Features/163 split reductions

Markus-Goetz:
* General refactoring and code cleanup
* Logical functionality reorganization of code base (final for now)
* Resolves issue #163


coquelin77:
this is fantastic! everything looks great here to me. Thank you very much for doing this. 

I see nothing wrong here. However since it is a large amount of changes I would still like to have @Cdebus 's opinion as well.

Issue #250, opened 2019-05-06
bug in the setters and getters when using dndarrays as keys


coquelin77:
it doesnt work in that case

PR #251, opened 2019-05-06
Assorted bug fixes

coquelin77:


coquelin77:
fixes #250 + others in setters and getters

coquelin77:
fix for #261 also implemented in most recent commit

heat/core/factories.py
```diff
@@ -117,10 +117,11 @@ def arange(*args, dtype=None, split=None, device=None, comm=MPI_WORLD):
     device = devices.sanitize_device(device)
     data = torch.arange(
         start, stop, step,
-        dtype=types.canonical_heat_type(dtype).torch_type(),
         device=device.torch_device
     )
 
+    data = data.type(types.canonical_heat_type(dtype).torch_type())
```
Markus-Goetz: Double inference of heat type (two lines below once again). Since Python does not compile, but is interpreted, this function will actually be executed twice. Introduce a local variable instead and reuse it

heat/core/tests/test_exponential.py
```diff
@@ -38,7 +38,7 @@ def test_exp(self):
 
         # exponential of longs, automatic conversion to intermediate floats
         int64_tensor = ht.arange(elements, dtype=ht.int64)
-        int64_exp = ht.exp(int64_tensor)
+        int64_exp = int64_tensor.exp()
```
Markus-Goetz: Why did you not leave it as the module level function? Did you just want to additionally test whether the method call works as well?

heat/core/tests/test_statistics.py
```diff
@@ -301,14 +301,6 @@ def test_mean(self):
                             self.assertEqual(res.split, len(target_dims) - 1)
                         else:
                             self.assertEqual(res.split, z.split)
-
```
Markus-Goetz: Please reorder mean(), std() and var() to appear in alphabetical order in their respective files including test cases.

heat/core/tests/test_exponential.py
coquelin77: to test the method call and to increase the coverage on the dndarray file.

PR #253, opened 2019-05-07
Hotfix for failing test on docker build

TheSlimvReal:
The most recent builds of travis were all failing because of a failing docker run.
This was caused by a rather weird behavior:
Locally calling transpose with an out of bound dimension resulted in an RuntimeError, but on the Docker this resulted in an IndexError and was therefore unexpected.

Because the transpose function wants to reraise all torch errors as a ValueError anyways, I changed it to catch all Exceptions not just the RuntimeErrors.

### DIFF
--- a/heat/core/linalg.py
+++ b/heat/core/linalg.py
@@ -63,7 +63,7 @@ def transpose(a, axes=None):
 
         return dndarray.DNDarray(transposed_data, transposed_shape, a.dtype, transposed_split, a.device, a.comm)
     # if not possible re- raise any torch exception as ValueError
-    except RuntimeError as exception:
+    except Exception as exception:
         raise ValueError(str(exception))
 
 


heat/core/linalg.py
```diff
@@ -63,7 +63,7 @@ def transpose(a, axes=None):
 
         return dndarray.DNDarray(transposed_data, transposed_shape, a.dtype, transposed_split, a.device, a.comm)
     # if not possible re- raise any torch exception as ValueError
-    except RuntimeError as exception:
+    except Exception as exception:
```
coquelin77: This is a dangerous bit of code. what about `except (RuntimeError, IndexError) as exception` ? this is also dangerous but is a bit more narrow. I asked for @Markus-Goetz to review this PR as well to see if he can see the problem.

However, I am still not sure about why the docker is throwing an IndexError here in the first place. all of my local tests do not show this behavior.

Markus-Goetz: Catching Exception generally is not something we should do here. The proposed solution of Daniel is much better. However, I would like to know why there is a difference in exceptions to begin with? Is it simply a PyTorch version thing?

PR #256, opened 2019-05-09
features/98-moments_2 gshape fixes and coverage improvement

coquelin77:
jumped the gun on the previous merge, need to improve coverage first


coquelin77:
as mentioned before, this was approved by markus verbally. coverage is taken care of here a bit as well.

these functions are not optimized yet but this is in the pipeline

PR #258, opened 2019-05-10
expand_dims() implementation

Markus-Goetz:
Resolves issue #258 

* added module-level function
* added docstrings
* added unittests

### DIFF
--- a/heat/core/manipulation.py
+++ b/heat/core/manipulation.py
@@ -1,0 +1,64 @@
+from . import dndarray
+from . import stride_tricks
+
+__all__ = [
+    'expand_dims'
+]
+
+
+def expand_dims(a, axis):
+    """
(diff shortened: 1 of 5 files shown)


heat/core/manipulation.py
```diff
@@ -0,0 +1,64 @@
...
+    ValueError
+        If the axis is not in range of the axes.
+
+    Examples
+    --------
+    >>> x = ht.array([1,2])
+    >>> x.shape
+    (2,)
+
+    >>> y = np.expand_dims(x, axis=0)
```
TheSlimvReal: This should be changed to `ht.expand_dims(x, axis=0)`. Same goes for the other examples.

Markus-Goetz: Thanks for the remark

PR #262, opened 2019-05-14
changed mean/var merge functions to operate on tensors

coquelin77:
improvements for mean/var functions.

while/for loops removed and the merge functions now operate on the full tensors.
this does not close the issue as an MPI function may still be faster, however this is a great improvement. The var test shows a very strong improvement. This test is faster with the 3rd dimension in the for loop than it was previously with only 2 dimensions

PS this is still failing due to the needed fixes in the bug fix PR


Markus-Goetz:
Bump, needs conflict resolution. If fixed can be merged by you Daniel

PR #265, opened 2019-05-15
Interactive heat interpreter

Markus-Goetz:
Resolves issue #259


rainman110:
Can you please describe, how to use this?

Markus-Goetz:
An interactive Python interpreter that allows to steer multiple MPI processes at once. Start as follows:

mpirun -np <procs> -stdin all python script/interactive.py

It is important that the stdin forwarding is enabled

Markus-Goetz:
This means that one of the MPI barriers is not yet in the right place. I thought I prevented this with the runcode() one.

TheSlimvReal:
I had the same bug as before again, but I could not reproduce it in a new session.

Markus-Goetz:
To answer some of the questions:

* According to the MPICH manual here: https://www.mpich.org/static/downloads/1.3.1/mpich2-1.3.1-userguide.pdf the -s all flag should also exist for MPICH. Could some double check please?

* Kai is indeed right, commands do not work at the moment. I have the feeling that it has something to do with the way the MPI implementation forward stdin. Might not be something that we can fix easily, but should maybe open an issue on it.

Issue #233, opened 2019-04-12
Wrong element sorting in resplit(axis=None) 


Cdebus:
When unsplitting , depending on the previous split axis (i.e. when split axis = 1), the individual parts of the distributed tensor are not aligned correctly, i.e. "rolling out" of the tensor into a 1D array misaligns the individual rows and columns.

Minimal working example for two processes (n=2)
c = tensor([[3., 3., 3., 3.], [3., 3., 3., 3.], [4., 4., 4., 4.], [4., 4., 4., 4.]], split = 1)
c._tensor__array
>>(Rank 0):  tensor([[3., 3.],
        [3., 3.],
        [4., 4.],
        [4., 4.]])
>>(Rank 1):  tensor([[3., 3.],
        [3., 3.],
        [4., 4.],
        [4., 4.]])
c.resplit(axis=None)
 >> tensor([[3.0000e+00, 3.0000e+00, 3.0000e+00, 3.0000e+00],
        [4.0000e+00, 4.0000e+00, 4.0000e+00, 4.0000e+00],
        [2.3049e-41, 2.3049e-41, 2.3049e-41, 2.3049e-41],
        [2.3138e-41, 2.3138e-41, 2.3138e-41, 1.6115e-19]])

Cdebus:
This problem is induced by wrongful functioning of Allgatherv

PR #268, opened 2019-05-21
167 tensor redistribution for binary operations

Cdebus:
Addresses #167 
Option for two different split axes not implemented yet, as it first needs fixing of #233 for allgather.
However, should still be merged, since it fixes issues regarding one split and one non-split tensor.
Also, better handling of cases with scalar values is supported


Cdebus:
Bump, still needs improvement with respect to broadcasting

PR #270, opened 2019-05-22
Feature 177 unique()

TheSlimvReal:
See issue #177.

Implementation of the unique function. 

Our current package requirement (`v1.0.0`) for torch version doesn't support `return_counts` (available since `v1.1.0`) yet and therefore is also not available here.

One thing that needs to be discussed is what should be returned in the `return_indices`. Should this return all the indices according to the complete (not distributed) data or should this be some kind of local indices.
Because when `a.split != axis`, we create a new array that is split among the available processes. This leads to the problem that you might end up with completely different vectors than in the beginning.


Markus-Goetz:
Bump, needs a conflicted master merge

heat/core/dndarray.py
```diff
@@ -2178,3 +2178,44 @@ def __truediv__(self, other):
...
+            Axis along which unique elements should be found. Default to None, which will return a one dimensional list of
+            unique values.
+
+        Returns
+        -------
+        res : ht.DNDarray
+            Output array. The unique elements. Elements are distributed the same way as the input tensor.
+        inverse_indices : torch.tensor (optional)
+            If return_inverse is True, this tensor will hold the list of inverse indices
+
```
Markus-Goetz: Extra line

heat/core/manipulations.py
```diff
@@ -1,16 +1,17 @@
 import torch
 import numpy as np
 
+import heat
```
Markus-Goetz: Why is heat imported here?

heat/core/manipulations.py
```diff
@@ -179,3 +180,170 @@ def squeeze(x, axis=None):
...
+        Axis along which unique elements should be found. Default to None, which will return a one dimensional list of
+        unique values.
+
+    Returns
+    -------
+    res : ht.DNDarray
+        Output array. The unique elements. Elements are distributed the same way as the input tensor.
+    inverse_indices : torch.tensor (optional)
+        If return_inverse is True, this tensor will hold the list of inverse indices
+
```
Markus-Goetz: Extra line

heat/core/manipulations.py
```diff
@@ -179,3 +180,170 @@ def squeeze(x, axis=None):
...
+                begin = steps[i]
+                end = steps[i + 1]
+                for num, x in enumerate(inverse_buf[begin: end]):
+                    inverse_indices[begin + num] = g_inverse[begin + x]
+
+    else:
+        max_uniques, max_pos = uniques_buf.max(0)
+
+        # find indices of vectors
+        if a.comm.Get_rank() is max_pos.item():
```
Markus-Goetz: `is` is the wrong operator here. It checks for equality and identity, equality is sufficient and what you actually want

heat/core/manipulations.py
```diff
@@ -179,3 +180,170 @@ def squeeze(x, axis=None):
...
+        gres = local_data[indices.tolist()]
+
+        is_split = a.split
+        inverse_indices = indices
+
+    if axis is not None:
+        # transpose matrix back
+        gres = gres.transpose(0, axis)
+
+    result = heat.array(gres, dtype=a.dtype, device=a.device, comm=a.comm, is_split=is_split)
```
Markus-Goetz: import .factories here instead

heat/core/tests/test_manipulations.py
```diff
@@ -202,3 +201,53 @@ def test_squeeze(self):
...
+
+        exp_axis_zero = ht.array([rank], dtype=ht.int32).expand_dims(0)
+        res = ht.unique(split_one, sorted=True, axis=0)
+        self.assertTrue((res._DNDarray__array == exp_axis_zero._DNDarray__array).all())
+
+        exp_axis_one = ht.array([rank] * size, dtype=ht.int32).expand_dims(1)
+        res = ht.unique(split_one, sorted=True, axis=1)
+        self.assertTrue((res._DNDarray__array == exp_axis_one._DNDarray__array).all())
+
+        # torch_array = torch.tensor([
```
Markus-Goetz: These tests will be uncommented as soon as Allgatherv() works? Is this reason that the code coverage is also not hitting all the lines?

We should postpone the merge for right after the Allgatherv() fix or issue another ticket.

TheSlimvReal: I changed the code to only use Allgatherv() with axis=0, so everything works as expected

PR #276, opened 2019-05-26
add leading dimensions to array for ndim<0 if necessary

krajsek:
This feature allows to create a HeAT Tensor with array with adding leading dimensions by setting a negative ndmin parameter


Markus-Goetz:
Resolves issue #275

PR #295, opened 2019-06-03
implement __radd__

krajsek:
solves issue #168


Markus-Goetz:
Might be inside the master now, due to the merge of the generic right-hand side operator branch.

Issue #37, opened 2018-07-16
Implement minimum()


bhagemeier:

coquelin77:
@Markus-Goetz isnt this done?

PR #298, opened 2019-06-04
Features/37 minimum

ClaudiaComito:
Hi all,

here's a HeAt implementation of np.minimum(x1, x2), providing the element-wise minimum of two tensors (Issue #37). This is equivalent to torch.min(x1, x2).

Some open questions:
1) I'm not entirely sure how torch.min handles NaN propagation when both elements are NaN. np.minimum reports: 

> If one of the elements being compared is a NaN, then that element is returned. (Claudia: this works). If both elements are NaNs then the first is returned. The latter distinction is important for complex NaNs, which are defined as at least one of the real or imaginary parts being a NaN. 

https://docs.scipy.org/doc/numpy/reference/generated/numpy.minimum.html

2) Ideas on how to handle x1.split != x2.split? Right now the code is a bit brutal (x1.split rules).  

Thanks for your comments!

### DIFF
--- a/heat/core/operations.py
+++ b/heat/core/operations.py
@@ -11,6 +11,7 @@ from . import types
 
 __all__ = []
 
+
 def __binary_op(operation, t1, t2):
     """
     Generic wrapper for element-wise binary operations of two operands (either can be tensor or scalar).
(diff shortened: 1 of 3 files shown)


ClaudiaComito:
After chatting with Daniel @coquelin77 yesterday, I'm looking into a solution for the case x1.split != x2.split. I'll close this PR for now.

PR #304, opened 2019-06-12
Features/20 where

coquelin77:
reopening previous PR


coquelin77:
this is still failing because the diff hit is 96.36% instead of 96.38%. Can we let this one slide?

PR #308, opened 2019-06-17
Features/187 balance

coquelin77:
implemented a balance function to balance a DNDarray across all nodes


heat/core/dndarray.py
```diff
@@ -483,6 +484,131 @@ def astype(self, dtype, copy=True):
 
         return self
 
+    def balance_(self):
```
TheSlimvReal: Why is the function called `balance_` instead of just `balance`? The `b.balance()` call results in error because this method does not exist.

heat/core/dndarray.py
```diff
@@ -483,6 +484,131 @@ def astype(self, dtype, copy=True):
...
+        _, _, chk = self.comm.chunk(self.shape, self.split)
+        chunk_map[self.comm.rank, 0] = chk[0].stop - chk[0].start
+        chunk_map[self.comm.rank, 1] = chk[1].stop - chk[1].start
+        chunk_map_comm = self.comm.Iallreduce(MPI.IN_PLACE, chunk_map, MPI.SUM)
+
+        lshape_map_comm.wait()
+        chunk_map_comm.wait()
+
+        # create list of which processes need to send data to lower ranked nodes
+        send_list = [True if lshape_map[pr, self.split] != (chunk_map[pr, self.split]) and lshape_map[pr, self.split] != 0 else False
```
TheSlimvReal: If the dimension of the DNDarray is > 2 and split along an axis > 2 an index error is thrown. 
To reproduce:
```
t = ht.ones((2, 4, 5), split=2)
b = t[:, :, 2:]
b.balance_()
```

heat/core/dndarray.py
```diff
@@ -483,6 +484,131 @@ def astype(self, dtype, copy=True):
...
+        [0/2] tensor([[3., 0.],
+        [1/2] tensor([[4., 0.],
+                      [5., 0.],
+                      [6., 0.]])
+        [2/2] tensor([[7., 0.],
+                      [8., 0.],
+                      [9., 0.]])
+        >>> b.balance()
+        >>> print(b.gshape, b.lshape)
+        [0/2] (7, 2) (1, 2)
```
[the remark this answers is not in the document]
coquelin77: i am showing it with 3 processes. if you run it with -np 3 this is what you get. if you look at the print above it it shows what the data actually looks like

heat/core/dndarray.py
coquelin77: thank you. I have fixed this. it was an oversight in the chunk map

heat/core/dndarray.py
coquelin77: its because this modifies the DNDarray itself and does not create a copy of it.

heat/core/dndarray.py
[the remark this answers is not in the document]
TheSlimvReal: But you balanced it which resulted in the tensor at rank = 0 to have a dimension of (3, 2), which is shown in the print below? Otherwise the dimensions wont add up to the total of 7 of the global shape.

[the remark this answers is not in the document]
coquelin77: you are correct. i have fixed this

Issue #144, opened 2019-02-25
Implement sort


Markus-Goetz:
Proposed implementation strategy: parallel sorting by regular sampling [1] + subsequent rebalancing of the potentially unequally distributed data chunks

[1] http://csweb.cs.wfu.edu/bigiron/LittleFE-PSRS/build/html/PSRSalgorithm.html

PR #310, opened 2019-06-24
Implement feature sort()

TheSlimvReal:
Targets issue #144.

Sort is implemented using some parallel sorting algorithms with pivots and a balancing afterwards. Problem case is where `axis = a.split` because there the sorting is affected by the values of the other processes. In the end the returned tensor has the same shape as the local input tensor. The lower processes have the lower (higher) values, if sorting is performed ascending (descending) (`MPI_WORLD.rank == 0` has the lowest (highest)).

The function currently does not return the indices of the elements in the original tensor because it is not clear if this should be a tensor equal to `a.gshape` holding the information for all values or just for the local elements. The global shape one would mean a very high memory consumption for big matrices while the local one would mean more communication between the processes. Is this something that should be implemented and if yes, in which way?


Markus-Goetz:
Bump, build fail

PR #312, opened 2019-06-25
Implement sort()

TheSlimvReal:
Targets issue #144.

Sort is implemented using some parallel sorting algorithms with pivots and a balancing afterwards. Problem case is where axis = a.split because there the sorting is affected by the values of the other processes. In the end the returned tensor has the same shape as the local input tensor. The lower processes have the lower (higher) values, if sorting is performed ascending (descending) (MPI_WORLD.rank == 0 has the lowest (highest)).

The sorting is not stable, which means that equal elements might be in a different ordering than in the original array. This leads to the problem that the indices of the elements in the original data can be different and still produce the same sorting result.


heat/core/manipulations.py
```diff
@@ -71,6 +75,311 @@ def expand_dims(a, axis):
...
+    Sorting where `axis == a.split` needs a lot of communication between the processes of MPI.
+
+    Parameters
+    ----------
+    a : ht.DNDarray
+        Input array to be sorted.
+    axis : int, optional
+        The dimension to sort along.
+        Default is the last axis.
+    descending : bool, optiona
```
Markus-Goetz: typo

heat/core/manipulations.py
```diff
@@ -71,6 +75,311 @@ def expand_dims(a, axis):
...
+        partition_matrix = torch.zeros(zeroes_dim, dtype=torch.int64)
+
+        # Create matrix that holds information, which value is shipped to which process
+        index_matrix = torch.empty_like(local_sorted, dtype=torch.int64)
+
+        # Iterate along the split axis which is now 0 due to transpose
+        for i, x in enumerate(local_sorted):
+            # Enumerate over all elements with correct index
+            for idx, val in np.ndenumerate(x.numpy()):
+                op_func = operator.gt if descending else operator.lt
```
Markus-Goetz: this only needs to be evaluated once as far as I can tell, so it should be moved outside the loop

heat/core/manipulations.py
```diff
@@ -71,6 +75,311 @@ def expand_dims(a, axis):
...
+    stride_tricks.sanitize_axis(a.shape, axis)
+
+    if a.split is None or axis != a.split:
+        # sorting is not affected by split -> we can just sort along the axis
+        final_result, final_indices = torch.sort(a._DNDarray__array, dim=axis, descending=descending)
+
+    else:
+        # sorting is affected by split, processes need to communicate results
+        # transpose so we can work along the 0 axis
+        transposed = a._DNDarray__array.transpose(axis, 0)
```
Markus-Goetz: Can we transpose this in-place and later transpose it back? This would make the memory-footprint a lot smaller

heat/core/manipulations.py
```diff
@@ -71,6 +75,311 @@ def expand_dims(a, axis):
...
+
+        # Create matrix that holds information which process gets how many values at which position
+        zeroes_dim = (size, ) + transposed.size()[1:]
+        partition_matrix = torch.zeros(zeroes_dim, dtype=torch.int64)
+
+        # Create matrix that holds information, which value is shipped to which process
+        index_matrix = torch.empty_like(local_sorted, dtype=torch.int64)
+
+        # Iterate along the split axis which is now 0 due to transpose
+        for i, x in enumerate(local_sorted):
```
Markus-Goetz: Question without having thought it through entirely. Is this loop and numpy conversion necessary here? Generally, this will add quite some overhead due to cpu-gpu transfer and due to Python ineffcient loops

heat/core/manipulations.py
```diff
@@ -71,6 +75,311 @@ def expand_dims(a, axis):
...
+        fill_level = torch.zeros(shape[1:], dtype=torch.int32)
+
+        for i, x in enumerate(send_recv_matrix):
+            for idx, val in np.ndenumerate(x.numpy()):
+                for k in range(val):
+                    recv_indices[fill_level[idx]][idx] = i
+                    fill_level[idx] += 1
+
+        # Finally send and receive the values with the correct processes according to the global pivots
+        for idx, val in np.ndenumerate(index_matrix.numpy()):
```
Markus-Goetz: Same as above regarding cpu-gpu efficiency + transfer

heat/core/manipulations.py
```diff
@@ -71,6 +75,311 @@ def expand_dims(a, axis):
...
+            # Add tag to identify correct value we want to receive later
+            tag = int(''.join([str(el) for el in idx[1:]]))
+            a.comm.Send(send_buf, dest=val, tag=tag)
+
+        recv_amount = sum(send_recv_matrix)
+        fill_level = torch.zeros(shape[1:], dtype=torch.int32)
+        first_result = torch.empty(shape, dtype=local_sorted.dtype)
+        first_indices = torch.empty_like(first_result)
+
+        for idx, val in np.ndenumerate(recv_amount.numpy()):
```
Markus-Goetz: Same as above regarding cpu-gpu efficiency + transfer

TheSlimvReal:
The code is now made more efficient to use less MPI=calls. It still needs one call for each coordinate that is not along the split axis but I don't see a way to improve this because every coordinate needs to be balanced for itself. The speed-up is roughly at about 6 compared to the version where a MPI-call was made for every swap.

Issue #174, opened 2019-03-12
Implement cat/concatenate


Markus-Goetz:

coquelin77:
@krajsek have you made any progress on this? I have made a lot of this now. i am just taking care of one last case

PR #319, opened 2019-07-01
Features/174 concatenate

coquelin77:
closes issue #174


heat/core/manipulations.py
```diff
@@ -1,18 +1,372 @@
...
+                    lcl_slice[axis] = slice(ttl, sz.item() + ttl, 1)
+                    arr0._DNDarray__array = arr0._DNDarray__array[lcl_slice].clone().squeeze()
+                ttl += sz.item()
+
+            if len(arr0.lshape) < len(arr1.lshape):
+                arr0._DNDarray__array.unsqueeze_(axis)
+            out = factories.empty(out_shape, split=s1)
+            out._DNDarray__array = torch.cat((arr0._DNDarray__array, arr1._DNDarray__array), dim=axis)
+            return out
+    elif s1 is None:
```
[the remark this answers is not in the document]
coquelin77: i can look into a modification to handle the `s0/1 != axis` cases but the important thing to notice here is that if arr0 is split then the data need to be 'front-loaded' onto the first `size/2` processes and in the case that arr1 is split then the data needs to be 'back-loaded' onto the last `size/2` processes

heat/core/manipulations.py
```diff
@@ -1,18 +1,372 @@
...
+    [1/1]         [1., 1.],
+    [1/1]         [1., 1.]])
+    """
+
+    arr0, arr1 = arrays[0], arrays[1]
+    if len(arrays) != 2:
+        if len(arrays) < 2:
+            raise ValueError('concatenate requires 2 arrays')
+        else:
+            raise NotImplementedError('concatenate is not implemented for >2 DNDarrays')
```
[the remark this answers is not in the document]
coquelin77: thanks. that is done

heat/core/manipulations.py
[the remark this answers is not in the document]
coquelin77: so i looked at this again and it is possible to do it but it comes at the cost of speed, not a huge cost be it requires more if loops. it is done though. and in the latest commit

coquelin77:
@Markus-Goetz I made some changes to types and factories.array. these might interest you. I couldnt debug that reduction buffer anymore so i left it and call another allreduce to get the proper shape after the raise statement

heat/core/manipulations.py
```diff
@@ -1,18 +1,275 @@
...
+    out_dtype = types.promote_types(arr0.dtype, arr1.dtype)
+    if arr0.dtype != out_dtype:
+        arr0 = out_dtype(arr0)
+    if arr1.dtype != out_dtype:
+        arr1 = out_dtype(arr1)
+
+    if not isinstance(arr0, dndarray.DNDarray) and not isinstance(arr1, dndarray.DNDarray):
+        raise TypeError('Both arrays must be DNDarrays')
+    if not isinstance(axis, int):
+        raise TypeError('axis must be an integer, currently: {}'.format(type(axis)))
```
ClaudiaComito: I suppose you have a good reason not to call sanitize_axis? At the moment ht.concatenate fails if axis<0. 

Example axis>0, this one works fine:

```
a = ht.array(ht.random.randn(2, 4), split=1)
b = ht.array(ht.random.randn(2, 7), split=1)
c = ht.concatenate((a, b), axis=1)
```

Equivalent with axis<0, doesn't run, see below:

```
a = ht.array(ht.random.randn(2, 4), split=1)
b = ht.array(ht.random.randn(2, 7), split=1)
c = ht.concatenate((a, b), axis=-1)
```

> > mpirun --use-hwthread-cpus -n 2 -tag-output  python local_test.py
> [1,0]<stderr>:Traceback (most recent call last):
> [1,0]<stderr>:  File "local_test.py", line 8, in <module>
> [1,0]<stderr>:    c = ht.concatenate((a, b), axis=-1)
> [1,0]<stderr>:  File "/Users/c.comito/HAF/heat/heat/core/manipulations.py", line 108, in concatenate
> [1,0]<stderr>:    ' {}, {}'.format(arr0.gshape, arr1.gshape))
> [1,0]<stderr>:ValueError: Arrays cannot be concatenated, gshapes must be the same in every axis except the selected axis: (2, 4), (2, 7)
> [1,1]<stderr>:Traceback (most recent call last):
> [1,1]<stderr>:  File "local_test.py", line 8, in <module>
> [1,1]<stderr>:    c = ht.concatenate((a, b), axis=-1)
> [1,1]<stderr>:  File "/Users/c.comito/HAF/heat/heat/core/manipulations.py", line 108, in concatenate
> [1,1]<stderr>:    ' {}, {}'.format(arr0.gshape, arr1.gshape))
> [1,1]<stderr>:ValueError: Arrays cannot be concatenated, gshapes must be the same in every axis except the selected axis: (2, 4), (2, 7)

heat/core/tests/test_manipulations.py
```diff
@@ -5,6 +5,279 @@
 
 
 class TestManipulations(unittest.TestCase):
+    def test_cat(self):
```
ClaudiaComito: test_concatenate instead of test_cat?

heat/core/tests/test_manipulations.py
```diff
@@ -5,6 +5,279 @@
...
+        # 0 None 0
+        y = ht.ones((16,), split=None, dtype=ht.int64)
+        res = ht.concatenate((x, y), axis=0)
+        self.assertEqual(res.gshape, (32,))
+        self.assertEqual(res.dtype, ht.float64)
+        _, _, chk = res.comm.chunk((32,), res.split)
+        lshape = [0, ]
+        lshape[0] = chk[0].stop - chk[0].start
+        self.assertEqual(res.lshape, tuple(lshape))
+
```
ClaudiaComito: Test exceptions?

heat/core/manipulations.py
coquelin77: > Hi Daniel @coquelin77,
> 
> I ran a few tests and as far as I can tell it works fine (except for the negative axis case).
> 
> I see that the concatenated tensor inherits the split value from the input tensors and is distributed, it makes sense but it's not what e.g. maximum() and minimum() are doing (and probably other reduction operations as well, I've got to check), so if we agree that's what we want, I'll make the relevant changes.

The reason that I did this was to minimize communication. I feel that it makes intuitive sense to inherent the split axis in this case. I do not handle the separate split case here as it would require more overhead. instead i leave it for the user to decide if they want to resplit the data. and if so then they can do it themselves.

PR #322, opened 2019-07-01
Features/305 cov

### LINKED ISSUE #305: Implement covariance

### LINKED ISSUE #375: Introduce dot() call
**Feature functionality**
Implement the dot() function call. Should be as is a simple forward to matmul(), respectively the @ operator

coquelin77:
closes #305 by implementing cov in the same style as numpy
closes #375 by implementing dot


heat/core/tests/test_statistics.py
```diff
@@ -176,6 +178,64 @@ def test_argmin(self):
...
             ht.argmin(data, axis=-4)
 
+    def test_cov(self):
+        x = ht.array([[0, 2], [1, 1], [2, 0]], dtype=ht.float, split=1).T
+        if x.comm.size < 3:
+            cov = ht.cov(x)
+            actual = ht.array([[1, -1], [-1, 1]], split=0)
+            self.assertTrue(ht.equal(cov, actual))
+
+        f = h5py.File('heat/datasets/data/iris.h5', 'r')
```
Markus-Goetz: This is a bad idea. h5py is an optional feature, it might not necessarily be installed. We should fall back here to a iris.csv version that we should also have in the datasets.

it might as well be useful for testing purposes to use ht.load_csv instead

heat/core/tests/test_statistics.py
```diff
@@ -176,6 +178,64 @@ def test_argmin(self):
...
+
+        if 1 < x.comm.size < 5:
+            htdata = ht.load('heat/datasets/data/iris.h5', 'data', split=1)
+            np_cov = np.cov(data, rowvar=False)
+            ht_cov = ht.cov(htdata, rowvar=False)
+            self.assertTrue(ht.allclose(ht.array(np_cov, dtype=ht.float), ht_cov, atol=1e-4))
+
+            np_cov = np.cov(data, data, rowvar=True)
+
+            htdata = ht.load('heat/datasets/data/iris.h5', 'data', split=0)
```
Markus-Goetz: see h5py thing

heat/core/tests/test_statistics.py
```diff
@@ -176,6 +178,64 @@ def test_argmin(self):
...
+
+        np_cov = np.cov(data, rowvar=False, ddof=1)
+        ht_cov = ht.cov(htdata, rowvar=False, ddof=1)
+        self.assertTrue(ht.allclose(ht.array(np_cov, dtype=ht.float) - ht_cov, 0, atol=1e-4))
+
+        np_cov = np.cov(data, rowvar=False, bias=True)
+        ht_cov = ht.cov(htdata, rowvar=False, bias=True)
+        self.assertTrue(ht.allclose(ht.array(np_cov, dtype=ht.float) - ht_cov, 0, atol=1e-4))
+
+        if 1 < x.comm.size < 5:
```
Markus-Goetz: why is this condition here? the functions should work in principle for all node combinations

heat/core/tests/test_statistics.py
coquelin77: i found a bug in load_csv. i have made an issue (#368) at the moment this problem does not allow for the the usage of load_csv.

heat/core/tests/test_statistics.py
coquelin77: its because with the split=1 condition there is an empty tensor on the nodes > 4. this throws an error from the binary_op
```
[4]   File "testing.py", line 46, in <module>
[4]     ht_cov = ht.cov(htdata, rowvar=False)
[4]   File "/home/d.coquelin/.git/heat/heat/core/statistics.py", line 419, in cov
[4]     x -= avg[:, None]
[4]   File "/home/d.coquelin/.git/heat/heat/core/dndarray.py", line 2643, in __sub__
[4]     return arithmetics.sub(self, other)
[4]   File "/home/d.coquelin/.git/heat/heat/core/arithmetics.py", line 374, in sub
[4]     return operations.__binary_op(torch.sub, t1, t2)
[4]   File "/home/d.coquelin/.git/heat/heat/core/operations.py", line 85, in __binary_op
[4]     output_shape = stride_tricks.broadcast_shape(t1.shape, t2.shape)
[4]   File "/home/d.coquelin/.git/heat/heat/core/stride_tricks.py", line 48, in broadcast_shape
[4]     raise ValueError('operands could not be broadcast, input shapes {} {}'.format(shape_a, shape_b))
[4] ValueError: operands could not be broadcast, input shapes (4, 150) (4,)
```

heat/core/linalg.py
```diff
@@ -210,18 +267,27 @@ def matmul(a, b):
...
+
         index_map_comm.wait()
         for pr in range(a.comm.size):
             start0 = index_map[pr, 0, 0, 0].item()
             stop0 = index_map[pr, 0, 0, 1].item()
             start1 = index_map[pr, 0, 1, 0].item()
             stop1 = index_map[pr, 0, 1, 1].item()
 
-            for dim0 in range((stop0 - start0) // mB):
+            for dim0 in range((stop0 - start0) // mB // a.comm.size if a_d0_1s_flag else (stop0 - start0) // mB):
```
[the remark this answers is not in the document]
coquelin77: the problem with using mul vs div here is that the integer division is the crucial operation here and not the simple division. So unfortunately the order of the operations matters in this case as well as the others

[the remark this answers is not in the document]
Markus-Goetz: True resolved

heat/core/tests/test_statistics.py
Markus-Goetz: Yes, I have assigned Simon already. So we need to wait here for a proper load_csv()

heat/core/tests/test_statistics.py
Markus-Goetz: interesting, is this documented as an issue?

PR #324, opened 2019-07-04
Features/37 minimum

ClaudiaComito:
Now addressing the case where x1.split != x2.split.


heat/core/statistics.py
```diff
@@ -533,6 +536,139 @@ def local_min(*args, **kwargs):
...
+    if x1.split is not None or x2.split is not None:
+        if x1.comm.is_distributed():  # assuming x1.comm = x2.comm
+            output_gshape = stride_tricks.broadcast_shape(x1.gshape, x2.gshape)
+            result = factories.empty(output_gshape)
+            x1.comm.Allgather(lresult, result)
+            # TODO: adopt Allgatherv() as soon as it is fixed, Issue #233
+            result._DNDarray__dtype = lresult._DNDarray__dtype
+            result._DNDarray__split = split
+
+            if out is not None:
```
coquelin77: is it possible to combine this `if out` and the other one?

heat/core/statistics.py
```diff
@@ -569,6 +704,15 @@ def mpi_argmin(a, b, _):
 
 MPI_ARGMIN = MPI.Op.Create(mpi_argmin, commute=True)
 
+# # TODO: implement mpi_minimum
```
coquelin77: Does this need to be here? this could be put in an issue

ClaudiaComito: Sorry about this, it's a leftover from the attempted implementation through __reduce_op(), I removed it already.

heat/core/statistics.py
ClaudiaComito: I'm not sure how, as one of them only writes out in the distributed case, the other one in the non-distributed case. If you can think of something, let me know.

Issue #326, opened 2019-07-11
Implement maximum()


ClaudiaComito:
Analogue to minimum().

PR #328, opened 2019-07-12
Features/326 maximum

ClaudiaComito:
Analogue implementation to ht.minimum(). Solves Issue #326.


ClaudiaComito:
I'm going to go ahead and merge this as it's identical to ht.minimum()

Issue #329, opened 2019-07-15
balance_() with dtype != float32 not working


TheSlimvReal:
To reproduce (with 3 processes):

```
import heat as ht
a = ht.zeros((10, 2), split=0, dtype=ht.float64)
a[:, 0] = ht.arange(10, dtype=ht.float64)
b = a[3:]
b.balance_()
```

Behavior: MPI freezes.
Same thing happens with all data types that are not float32.

Issue #238, opened 2019-04-16
Provide load_csv()


Markus-Goetz:

TheSlimvReal:
should this be implemented in a general manner like  `numpy.genfromtxt()` or specific for the csv case?

PR #330, opened 2019-07-15
Feature 238 load_cvs()

TheSlimvReal:
Targets issue #238.

Loading a CSV works. The CSV is always distributed along the 0 axis because only this way the file can be read in parallel. 
It currently does not work with types other than `float32` because of the use of `balance_()` and issue #329.  

Question is, should the saving also be implemented or is this another issue?


heat/core/io.py
```diff
@@ -8,15 +9,18 @@
...
+__CSV_EXTENSION = frozenset(['.csv'])
 __HDF5_EXTENSIONS = frozenset(['.h5', '.hdf5'])
 __NETCDF_EXTENSIONS = frozenset(['.nc', '.nc4', 'netcdf'])
 __NETCDF_DIM_TEMPLATE = '{}_dim_{}'
 
 __all__ = [
     'load',
-    'save'
+    'save',
+    'load_csv'
```
Markus-Goetz: Please sort alphabetically

heat/core/io.py
```diff
@@ -406,14 +409,179 @@ def load(path, *args, **kwargs):
...
+        return load_csv(path, *args, **kwargs)
+    elif supports_hdf5() and extension in __HDF5_EXTENSIONS:
         return load_hdf5(path, *args, **kwargs)
     elif supports_netcdf() and extension in __NETCDF_EXTENSIONS:
         return load_netcdf(path, *args, **kwargs)
     else:
         raise ValueError('Unsupported file extension {}'.format(extension))
 
 
+def load_csv(path, header_lines=0, sep=';', dtype=types.float32, encoding='UTF-8', device=None, comm=MPI_WORLD):
```
Markus-Goetz: Shouldn't default sep be comma rather than semi-colon?

Markus-Goetz: Shouldn't there also be a split parameter as to request data partitioning?

heat/core/io.py
```diff
@@ -406,14 +409,179 @@ def load(path, *args, **kwargs):
...
+    size = comm.size
+    counts, displs, _ = comm.counts_displs_shape((file_size, 1), 0)
+
+    # Read a chunk of bytes and count the linebreaks
+    f = open(path, 'rb')
+    f.seek(displs[rank], 0)
+    line_starts = []
+    r = f.read(counts[rank])
+    for pos, l in enumerate(r):
+        if chr(l) == '\n':
```
Markus-Goetz: What about carriage return + line feed? Might want to use os.linesep instead here

heat/core/io.py
```diff
@@ -406,14 +409,179 @@ def load(path, *args, **kwargs):
...
+            line_starts.append(pos + 1)
+
+    if rank == 0:
+        line_starts = [0] + line_starts
```
Markus-Goetz: Do we need to store all line ending or rather only the number of lines locally and then research the buffer?

TheSlimvReal: This makes it more easy to read in the lines later. Do you think it costs to much memory to store it?

Markus-Goetz: Potentially. I am trying to determine right now, whether there will actually be so large CSV files that this will be an issue. For now we can leave it as is

Issue #331, opened 2019-07-16
resplit() of tensor with typer other than float 32 not working


TheSlimvReal:
To reproduce (with 2 processes):

```
>>> import heat as ht
>>> a = ht.ones((2, 2), dtype=ht.int32)
>>> a.resplit(None)
>>> a
tensor([[1.4013e-15, 1.4013e-15]
            [1.4013e-15, 1.4013e-15]])
```

For 64 bit data types the resplit function call completely breaks.
On a first look it seems like the buffer of the `allgatherv` call does not have the correct datatype.

PR #332, opened 2019-07-16
Fixed resplit(None) and added some test cases

TheSlimvReal:
Targets issues #331.

Resplit only worked for `dtype == ht.float32` because the buffer was not set up correctly.
Also there is the problem with `allgatherv` along a axis other than zero. As long as this problem exists we just transpose the data before and after the communication.


Markus-Goetz:
Good PR, but has been also fixed in the larger allgather(v) rework in PR #342. Your fixes are in there and will be pulled instead. Good work though!

Issue #313, opened 2019-06-25
Code of Conduct


Cdebus:
Create a code of conduct (from the contributors template), review it according to our needs, and provide email address as spokes person

PR #333, opened 2019-07-16
Create CODE_OF_CONDUCT.md

ClaudiaComito:
Issue #313 

Dear all,

please read the CoC and give your blessings or suggest changes/improvements. In this specific case, feel free to get in touch by email if you don't want to have a public discussion. 

Cheers,

Your CoC enforcer

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
+appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+ advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+ address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+ professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at c.comito@fz-juelich.de. All
+complaints will be reviewed and investigated and will result in a response that
+is deemed necessary and appropriate to the circumstances. The project team is
+obligated to maintain confidentiality with regard to the reporter of an incident.
+Further details of specific enforcement policies may be posted separately.
+
+Project maintainers who do not follow or enforce the Code of Conduct in good
+faith may face temporary or permanent repercussions as determined by other
+members of the project's leadership.
+
+## Attribution
+
+This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 1.4,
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
+appearance, race, religion, or sexual identity and orientation.
```
coquelin77: maybe `... ,sexual identity, or sexual orientation` this way it covers both and not just the two together.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
```
coquelin77: should it be `aligned with` instead of `aligned to`?

what about `... or to ban, temporarily or permanently, any contributor ...`

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at c.comito@fz-juelich.de. All
+complaints will be reviewed and investigated and will result in a response that
+is deemed necessary and appropriate to the circumstances. The project team is
+obligated to maintain confidentiality with regard to the reporter of an incident.
+Further details of specific enforcement policies may be posted separately.
```
coquelin77: Possible edits: `Instances of abuse, harassment, or any other unacceptable behavior may be reported by contacting the project team at c.comito@fz-juelich.de.`

ClaudiaComito:
Thanks for your comments. @coquelin77 this is the template github recommends/proposes for smaller projects, I certainly didn't write it myself...

CODE_OF_CONDUCT.md
Cdebus: What exactly is the difference between "sex characteristics, gender identity and expression", and  "sexual identity, or sexual orientation" ? To me they describe the same thing. However, it probably doesn't hurt to have more inclusions, I'm just curious...

Cdebus: maybe add something like "linguistic expression"? We don't want people to be harassed justed because they can't write proper english either, right?

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
```
Cdebus: Also, giving constructive criticism, rather than destructive ?

CODE_OF_CONDUCT.md
Cdebus: I'd leave it at abusive, harrassing ... behavious, since this is more loosley termed, and thus covers a broader range (at least according to my understanding). It's hard to directly classify something as abuse... what's the correct definition of that?

Cdebus:
I like it, just gave my 2cents to some of the points.
But it definitely doesn't restrict us in any way, it just says we can ban bullies ;)

CODE_OF_CONDUCT.md
coquelin77: from google about gender vs sex: In general terms, "sex" refers to the biological differences between males and females, such as the genitalia and genetic differences. "Gender" is more difficult to define, but it can refer to the role of a male or female in society, known as a gender role, or an individual's concept of themselves, or gender identity.

i think that listing everything is just a way to cover all the possible bases. 

with regards to linguistic expression, i dont think that we should add something like that. If required i think that it would fall under education. I am also wary about adding something like that as if someone was very sensitive about it they might feel attacked for some typos which had occurred in a doc string. Obviously, if the comments about the typos constitute harassment or abuse then this, in my belief, would be covered by the education section of this paragraph.

CODE_OF_CONDUCT.md
coquelin77: to be honest what i suggested was only because of how it hits my ear, but this is common with legal language for me.

abuse(v) is defined as to treat with cruelty or violence, especially regularly or repeatedly or to speak to (someone) in an insulting and offensive way.
whereas harassment(n) is defined as aggressive pressure or intimidation.

to me what makes the difference is the frequency, severity, and nature of the behavior. but this could still be inaccurate. the difference between these two is more clearly, and legally, defined for sexual behaviors. For non-sexual behaviors i think you could use the legal definitions of the two as a guide to differentiate between them.

Markus-Goetz:
Have all the comments been worked in?

CODE_OF_CONDUCT.md
ClaudiaComito: @cdebus, @coquelin77 here's my 2 cents:

This template was written by somebody who presumably knew what was doing, i.e. I guess a legal team at GitHub. Whatever change we introduce, we do it at our own risk - where the most likely "risk" in our case is to overthink/overcomplicate things. 

So maybe let's take a step back and ask ourselves, is there anything on this CoC that we  can't live with. 

We can always improve it later, or write our own. Here's an example of the numpy Code of Conduct:
https://numpy.org/devdocs/dev/conduct/code_of_conduct.html

pytorch doesn't have one.

Issue #346, opened 2019-07-26
Bug in setitem/getitem, single selection of multidimensional array wrong gshape


coquelin77:
**Description**
wrong gshape created for getting an array with >2 dimensions from a single integer key

2nd bug here as well: false call in the setitem for the case of an integer key with split > 0

**To Reproduce**
1. Which module/class/function is affected? - dndarray.py
2. What are the circumstances under which the bug appears? 
```python
out = ht.ones((4, 5, 5), split=1)
    print(out[0].gshape, out[0].lshape)
```
3. What is the exact error-message/errorous behavious?
```
[0] (2, 20) (2, 5)
[1] (2, 20) (1, 5)
[2] (2, 20) (1, 5)
[3] (2, 20) (1, 5)
```

**Expected behavior**
out[0].gshape should equal (5, 5), out[0].lshape is correct here

for second error: this is routed in the first problem
```[1] Traceback (most recent call last):
[1]   File "testing.py", line 60, in <module>
[1]     out[day] = mean
[1]   File "/home/d.coquelin/.git/heat/heat/core/dndarray.py", line 2294, in __setitem__
[1]     self[key, :] = value
[1]   File "/home/d.coquelin/.git/heat/heat/core/dndarray.py", line 2308, in __setitem__
[1]     self.__setter(tuple(key), value[overlap])
[1]   File "/home/d.coquelin/.git/heat/heat/core/dndarray.py", line 1302, in __getitem__
[1]     if isinstance(key[self.split], slice):  # if a slice is given in the split direction
[1] IndexError: list index out of range
```

**Version Info**
most current master (07.26.2019)

PR #348, opened 2019-07-29
Bug/346 getitem wrong gshape

coquelin77:
bug fix for #346 

issue was in the false setting of the new split axis. 
a secondary bug was found in the setter which did not properly call the setter function. this was also rectified.


coquelin77:
you are absolutely correct. this is added into the latest commit

PR #353, opened 2019-08-02
unique() Bugfix and improvement

TheSlimvReal:
## Description

Until now the `unique` function did not distribute the result if it was one dimensional. That changed with this PR. 
The code also got shrunk and optimized a little bit.

I also realized a problem for the sorting in the case where `axis != None and axis != a.split`.
For this case the process with the most unique tensors tells the others which of their local tensors they need to keep. This results in an unordered result. To sort this order we can't use the `sort` method, because this sorts the values independently. Instead we would need a vector-sort which keeps the actual vectors and only sorts them along `axis`. I could not think of an effective algorithm for this and therefore left it out for the moment

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)


heat/core/manipulations.py
```diff
@@ -609,41 +603,32 @@ def unique(a, sorted=False, return_inverse=False, axis=None):
...
-            indices = torch.empty((max_uniques.item(),), dtype=a.dtype.torch_type())
+            indices = torch.empty((max_uniques.item(),), dtype=inverse_pos.dtype)
 
         a.comm.Bcast(indices, root=max_pos)
+
         gres = local_data[indices.tolist()]
 
-        is_split = a.split
         inverse_indices = indices
+        if sorted:
```
coquelin77: should be an issue (if it isn't already) and then this can be removed or raise/return NotImplement/Error

Issue #354, opened 2019-08-02
__local_op returns tensor with wrong global shape


TheSlimvReal:
**Description**
The operation `__local_op` is used in methods where a local result can be computed without communicating with the other processes. The problem is it expects the result to have `gshape == lshape` which is mostly wrong.

**To Reproduce**
1. Which module/class/function is affected?
All using the `__local_op` function
2. What are the circumstances under which the bug appears? 
Accessing `gshape` on the result of a computation which internally used the `__local_op` (e.g. `exp`)
3. What is the exact error-message/errorous behavious?
The global shape does not reflect the actual global shape

**Expected behavior**
The `gshape` should represent the actual global shape

**Illustrative**

Example with 4 cores
```
>>> a = ht.ones((4, ), split=0)
>>> a = a.exp()
>>> a.gshape
shape (1,)
```

**Version Info**
Master-Branch

**Additional comments**
I would suggest one communication call to calculate the resulting global shape.

PR #355, opened 2019-08-02
__local_op bugfix

TheSlimvReal:
## Description

using the `__local_op` resulted in a wrong global shape

Fixes: #354 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### DIFF
--- a/heat/core/operations.py
+++ b/heat/core/operations.py
@@ -172,7 +172,12 @@ def __local_op(operation, x, out, **kwargs):
         if kwargs_dtype:
             del kwargs['dtype']
         result = operation(x._DNDarray__array.type(torch_type), **kwargs)
-        return dndarray.DNDarray(result, tuple(result.shape), promoted_type if kwargs_dtype is None else kwargs_dtype,
+        gshape = list(result.shape)
+        if x.split is not None:
+            split_shape = torch.tensor([result.shape[x.split]])
+            x.comm.Allreduce(MPI.IN_PLACE, split_shape, MPI.SUM)
+            gshape[x.split] = split_shape.item()
+        return dndarray.DNDarray(result, tuple(gshape), promoted_type if kwargs_dtype is None else kwargs_dtype,
                                  x.split, x.device, x.comm)
 
     # output buffer writing requires a bit more work
--- a/heat/core/tests/test_exponential.py
+++ b/heat/core/tests/test_exponential.py
@@ -45,6 +45,10 @@ class TestExponential(unittest.TestCase):
         with self.assertRaises(TypeError):
             ht.exp('hello world')
 
+    def test_exp_with_split(self):
+        expected = torch.arange(10, dtype=torch.float32).exp()
+        actual = ht.arange(10, split=0, dtype=ht.float32).exp()
+        self.assertEqual(actual.gshape, tuple(expected.shape))
 
     def test_expm1(self):
         elements = 10


heat/core/operations.py
```diff
@@ -172,8 +172,23 @@ def __local_op(operation, x, out, **kwargs):
         if kwargs_dtype:
             del kwargs['dtype']
         result = operation(x._DNDarray__array.type(torch_type), **kwargs)
-        return dndarray.DNDarray(result, tuple(result.shape), promoted_type if kwargs_dtype is None else kwargs_dtype,
-                                 x.split, x.device, x.comm)
+        split = x.split
+        gshape = list(result.shape)
+        if x.split is not None:
+            # The operation could have reduced the result by one axis
```
Markus-Goetz: How is this the case for a local_op? It is just element-wise application of a function. There should never be a reduction happening

heat/core/operations.py
```diff
@@ -172,8 +172,23 @@ def __local_op(operation, x, out, **kwargs):
...
+            axis = x.split if x.split < len(result.shape) else len(result.shape) - 1
+            if len(result.shape) > 0:
+                split_shape = torch.tensor([result.shape[axis]])
+            else:
+                # 0-Dimensional tensor in other words just a number on each process
+                split_shape = torch.tensor([1 if x.shape[x.split] > 0 else 0], dtype=result.dtype)
+                gshape = [0]
+            x.comm.Allreduce(MPI.IN_PLACE, split_shape, MPI.SUM)
```
Markus-Goetz: Why not simply use the gshape and lshape of x?

heat/core/operations.py
TheSlimvReal: Some functions (e.g. var()) use the local_op for some local temporary computations (e.g. calculation mean locally and sharing it later). Either we remove this "false" usage or we need to make this case differentiation.

heat/core/operations.py
TheSlimvReal: In the case where a reduction happened this is necessary.

Markus-Goetz:
@coquelin77 mentioned removing the local_ops from mean() and var(). Is this still necessary then?

coquelin77:
@Markus-Goetz i made a couple changes to the local_op function as well. essentially changing it to use array instead of the DNDarray constructor. i think its neater this way. it takes care of the gshape bug in the case of dimensional reductions caused by the local operation. The other feature i implemented is the flag to avoid casting.

Markus-Goetz:
Does this mean we need this PR still. Personally, I would want to avoid the proposed changes (albeit not incorrect), as it would introduce more communication for shape determination. If we could keep it as a plain local operator, we would not need to communicate.

coquelin77:
it does fix the dtype bug that simon had. however that fix only requires the no_cast flag and not the array call.

Markus-Goetz:
So I would propose that this PR needs to be changed insofar that we only set the no_cast flag and the rest should not be merged?!

coquelin77:
if we want to have the local_op which maintains the gshape we need some communication. otherwise we have to assume that the local operations do not change the lshapes. thoughts?

Markus-Goetz:
Well local_op should not. It is meant for point-wise operations like pow(), exp(), sqrt() etc. These will leave the global shape unchanged.

Markus-Goetz:
Should be perhaps called local_elementwise_op()?

coquelin77:
i think we could just make a note in the doc string. i dont think we need to refactor

PR #362, opened 2019-08-26
Features/54 distributed random numbers

TheSlimvReal:
## Description

Target issue #54.

This builds on @Markus-Goetz implementation of the `threefry64`  algorithm. The `__counter_sequenze` function was completed which returns initial values for the "random" number calculation. It is currently only tested for `torch.int64` because the `threefry32` is not implemented yet.


heat/core/random.py
```diff
@@ -1,44 +1,376 @@
...
+        diff -= signed_mask
+    x_1 += diff
+
+    # generate the x_0 counter sequence
+    x_0 = torch.empty_like(x_1)
+    diff = c_0 - signed_mask
+    if diff > 0:
+        # same problem as for x_1 with the overflow
+        x_0.fill_(signed_mask)
+        while diff > signed_mask:
```
Markus-Goetz: This is somewhat sub-optimal from a performance perspective. Would be better to have something like vectorized version of this without a while-loop. Potentially determine the split point and then do the minus and plus magic?

heat/core/tests/test_random.py
```diff
@@ -1,28 +1,88 @@
 import unittest
 
 import heat as ht
+import numpy as np
 
 
 class TestTensor(unittest.TestCase):
+    def test_rand(self):
```
Markus-Goetz: Need a lot more tests here for uniform and randint

heat/core/random.py
TheSlimvReal: With the check `total_elements <= 2 * 0xffffffff` this loop runs for a maximum of four times. However, adding a value larger than 0x7fffffff is not possible and this iterative process also takes care of to correct overflow after 0xffffffff is reached. I can not think of a direct way to get this expected overflow because at some point the sign needs to be flipped.

Markus-Goetz: Yes, sorry you explained that obviously already. Nevermind

Markus-Goetz:
What is the state regarding the 32-bit Threefry? Have you tried the upper bit removal?

Issue #341, opened 2019-07-22
Rework MPI communication Wrappers


Cdebus:
**Related**
Related to Issue #233 : Wrong sorting in dndarray resplit was due to errorous allgatherv

**Feature functionality**
Refactor MPI global functions: allgather*, alltoall*, scatter*, gather*.
So far algorithmic of all these functions is bundled in _scatter_like. However, some assumptions do not hold for all functions. Thus, each group gets own generic wrapper

**Additional context**
Also implement testing functions accordingly (to see if elements are sorted correctly)

ClaudiaComito:
What's the status here @Markus-Goetz ?

Markus-Goetz:
Still pending, this issue also requires to account for non-contiguous striding of the input tensors, e.g. if it is a view, to read out the correct elements from the buffers.

ClaudiaComito:
@Markus-Goetz 

> 

Is that right? @Markus-Goetz maybe I misunderstand the original issue. We've had `__allgather_like`, `__alltoall_like`, `__broadcast_like` for a very long time.

mrfh92:
Resolved by merging of #1204

PR #364, opened 2019-08-28
All to all mpi type buffer prep

Markus-Goetz:
Related to #341
Fixes Wrapper for MPI AllToAll

## Description
Before, alltoall was treated as a _scatter_like function, with the corresponding buffer preparation. However, due to the fact that AllToAll does send only specific blocks of data to the individual ranks, the strides in the send and receive buffer preparation via Create_Vector did lead to 'scrambled' tensors for non-continuous data, especially in the case when the split axis dimension does not distribute equally over all ranks. 
This PR proposes individual buffer preparations and receive buffer, that creates n=number of processes custom data types for the send buffer , according to the count distribution along the new split axis, that guaranties a specific stride vector wrapping for the receiving buffer.
If either the send or the receive buffer is not send continuous (i.e. alltoall involves axes other than only 0 and 1), MPI_AllToAllw is called with the custom data types and calculated counts and displacements of the custom datatypes.
Since this can be computationally rather expensive, a shortcut version via the old (continuous) buffer preparation is offered for the most common split axes 0 and 1.
The fix will have direct impact on the dndarray.resplit(axis) function

Changes proposed:
-  custom buffer preparations for alltoall sendbuffer and receive buffer
- alltoall functions with involved split axes >1 are all handled by mpi4py.Alltoallw
-  shortcut via old as_buffer --> mpi_type_and_elements for split axes <=1

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

Have you handled and tested all split configurations?
Yes, but dedicated unit tests still need to be implemented

### DIFF
--- a/heat/core/communication.py
+++ b/heat/core/communication.py
@@ -169,6 +169,9 @@ class MPICommunication(Communication):
             The data type object
         elements : int or tuple of ints
             The number of elements of the respective data type
+
+        # ToDo: The option to explicitely specify the counts and displacements to be send still needs propper implementation
+
         """
         mpi_type, elements = cls.__mpi_type_mappings[obj.dtype], torch.numel(obj)
 
@@ -240,6 +243,99 @@ class MPICommunication(Communication):
         mpi_type, elements = cls.mpi_type_and_elements_of(obj, counts, displs)
 
         return [cls.as_mpi_memory(obj), elements, mpi_type]
+
+    def alltoall_sendbuffer(self, obj):
+        """
+        Converts a passed torch tensor into a memory buffer object with associated number of elements and MPI data type.
+        XXX: might not work for all MPI stacks. Might require multiple type commits or so
+
+        Parameters
+        ----------
+        obj: torch.Tensor
+             The object to be transformed into a custom MPI datatype
+
+        """
+        mpi_type, elements = self.__mpi_type_mappings[obj.dtype], torch.numel(obj)
+
+        nproc = self.size
+        shape = obj.shape
+        strides = [1] * len(shape)
+        strides[-1] = obj.stride()[-1]
+        offsets = [0] * len(shape)
+        offsets[1:] = [obj.element_size() * stride for stride in obj.stride()[:-1]]
+
+        # Step 1: Wrap along axes > 1 (all axes except send_axis and recv_axis
+        for i in range(len(shape) - 1, 1, -1):
+            mpi_type = mpi_type.Create_vector(shape[i], 1, strides[i]).Create_resized(0, offsets[i])
+            mpi_type.Commit()
+
+        # Step 2: Create Custom sized vector datatypes, according to rank-specific size along send_axis
+        # send_elements has nproc entries, defining how many vectors of mpi_type are stacked together for each process to receive along the send_axis
+        send_elements = np.full((nproc,), obj.shape[1] // nproc)
+        send_elements[:obj.shape[1] % nproc] += 1
+
+        #Create short_Type from the last entry of send_elements
+        mpi_short_type = mpi_type.Create_vector(send_elements[-1], 1, strides[1]).Create_resized(0, offsets[1])
+        mpi_short_type.Commit()
+        #Create long_Type from the first entry of send_elements (wraps one more mpi_type vector than short_Type
+        mpi_long_type = mpi_type.Create_vector(send_elements[0], 1, strides[1]).Create_resized(0, offsets[1])
+        mpi_long_type.Commit()
+
+        #Step 3: Pack short_type and long_type along the recv_axis
+        mpi_short_type=mpi_short_type.Create_vector(shape[0], 1, strides[0]).Create_resized(0, send_elements[-1] * obj.stride()[1] * obj.element_size())
+        mpi_short_type.Commit()
+        mpi_long_type=mpi_long_type.Create_vector(shape[0], 1, strides[0]).Create_resized(0, send_elements[0] * obj.stride()[1] * obj.element_size())
+        mpi_long_type.Commit()
+
+        #Step 4: Prepare sencounts, senddispls and sendtypes for alltoallw
+        # to each process 1 element (=sendcount) of the custom prepared long or short type will be send
+        sendcount = [1] * nproc
+        tmp_displs = [0] * nproc
+        tmp_displs[1:] = np.cumsum(send_elements[:-1] )
+        element_size = obj.element_size()
+        senddispls = [element_size * obj.stride()[1] * d for d in tmp_displs]
+        sendtypes = [mpi_short_type] * nproc
+        for i in range(obj.shape[1] % nproc):
+            sendtypes[i] = mpi_long_type
+
+        return self.as_mpi_memory(obj), (sendcount, senddispls), sendtypes
+
+    def alltoall_recvbuffer(self, obj):
+        """
+        Converts a passed torch tensor into a memory buffer object with associated number of elements and MPI data type.
+        XXX: might not work for all MPI stacks. Might require multiple type commits or so
+
+        Parameters
+        ----------
+        obj: torch.Tensor
+             The object to be transformed into a custom MPI datatype
+        """
+
+        mpi_type, elements = self.__mpi_type_mappings[obj.dtype], torch.numel(obj)
+
+        nproc = self.size
+        shape = obj.shape[1:]
+        strides = [1] * len(shape)
+        strides[0] = obj.stride()[-1]
+        strides = strides[::-1]
+        offsets = [obj.element_size() * stride for stride in obj.stride()[:-1]]
+
+        # Step 1: Wrap along axes > 0 (all axes except recv_axis)
+        for i in range(len(shape) - 1, -1, -1):
+            mpi_type = mpi_type.Create_vector(shape[i], 1, strides[i]).Create_resized(0, offsets[i])
+            mpi_type.Commit()
+
+        # Step 2: Receive blocks along the recv axis
+        # Prepare recvcount, senddispls and sendtypes for alltoallw
+        recvcount = np.full((nproc,), obj.shape[0] // nproc)
+        recvcount[:obj.shape[0] % nproc]+=1
+        # size/extent of mpitype = offsets[0]
+        tmp_displs = [0] * nproc
+        tmp_displs[1:] = np.cumsum(recvcount[:-1])
+        recvdispls = [offsets[0] * d for d in tmp_displs]
+        recvtypes = [mpi_type] * nproc
+
+        return self.as_mpi_memory(obj), (recvcount, recvdispls), recvtypes
 
 
     def Irecv(self, buf, source=MPI.ANY_SOURCE, tag=MPI.ANY_TAG):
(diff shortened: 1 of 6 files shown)


heat/core/communication.py
```diff
@@ -199,6 +199,7 @@ def mpi_type_and_elements_of(cls, obj, counts, displs):
 
     @classmethod
     def as_mpi_memory(cls, obj):
+
```
Markus-Goetz: Remove

heat/core/communication.py
```diff
@@ -463,6 +464,161 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
         return self.__allgather_like(self.handle.Iallgatherv, sendbuf, recvbuf, send_axis, recv_axis)
     Iallgatherv.__doc__ = MPI.Comm.Iallgatherv.__doc__
 
+    @classmethod
+    def alltoall_sendbuffer(cls, obj, nproc, counts=None, displs=None): #This is a temporary soltion to pass number of arguments as variable. Better to access via cls itself, but how?
```
Markus-Goetz: I do not understand the comment given.
What would you like to access via the class directly? counts and displs? If so, they are perfectly fine as parameters. If not, what is it?

heat/core/communication.py
```diff
@@ -463,6 +464,161 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
...
+        mpi_short_type.Commit()
+        mpi_long_type=mpi_long_type.Create_vector(shape[0], 1, strides[0]).Create_resized(0, send_elements[0] * obj.stride()[1] * obj.element_size())
+        mpi_long_type.Commit()
+
+        #Step 4: Prepare sencounts, senddispls and sendtypes for alltoallw
+        # to each process 1 element (=sendcount) of the custom prepared long or short type will be send
+        sendcount = [1]*nproc
+        tmp_displs = [0] * nproc
+        tmp_displs[1:] = np.cumsum(send_elements[:-1] )
+        senddispls = [obj.element_size() * obj.stride()[1] * d for d in tmp_displs]
```
Markus-Goetz: slight optimization: call element_size() only once outside the loop

heat/core/communication.py
```diff
@@ -463,6 +464,161 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
...
+        offsets = [obj.element_size() * stride for stride in obj.stride()[:-1]]
+
+        # Step 1: Wrap along axes > 0 (all axes except recv_axis)
+        for i in range(len(shape) - 1, -1, -1):
+            mpi_type = mpi_type.Create_vector(shape[i], 1, strides[i]).Create_resized(0, offsets[i])
+            mpi_type.Commit()
+
+        # Step 2: Receive blocks along the recv axis
+        # Prepare recvcount, senddispls and sendtypes for alltoallw
+        recvcount = np.full((nproc,), obj.shape[0]//nproc)
```
Markus-Goetz: spaces around binary operators

heat/core/communication.py
```diff
@@ -463,6 +464,161 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
...
+        tmp_displs = [0] * nproc
+        tmp_displs[1:] = np.cumsum(recvcount[:-1] )
+        recvdispls = [offsets[0] * d for d in tmp_displs]
+        recvtypes = [mpi_type] * nproc
+
+        return cls.as_mpi_memory(obj), (recvcount, recvdispls), recvtypes
+
+
+    def __alltoall_w(self, sendbuf, recvbuf, send_axis, recv_axis, **kwargs):
+
```
Markus-Goetz: remove

heat/core/communication.py
```diff
@@ -463,6 +464,161 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
...
+            axis_permutation[0], axis_permutation[recv_axis] = recv_axis, 0
+            axis_permutation[send_axis] = axis_permutation[1]
+            axis_permutation[1] = send_axis
+
+        sendbuf = sendbuf.permute(*axis_permutation)
+        recvbuf = recvbuf.permute(*axis_permutation)
+
+        # prepare buffer objects
+        nproc = self.size
+        #if sendbuf is not MPI.IN_PLACE: --> Can we allow IN_Place opperation? (Buffer config might be wrong
```
Markus-Goetz: MPI_IN_PLACE should in principle be allowed

heat/core/communication.py
```diff
@@ -463,6 +464,161 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
...
+        if(((axis == 0) & (recv_axis == 1)) | ((axis == 1) & (recv_axis == 0))):
+            return self.__scatter_like(self.handle.Alltoall, sendbuf, recvbuf, axis, recv_axis, send_factor=self.size, recv_factor=self.size)
+        else:
+            return self.__alltoall_w(sendbuf, recvbuf, axis, recv_axis)
+    Alltoall.__doc__ = MPI.Comm.Alltoall.__doc__
+
+    def Alltoallv(self, sendbuf, recvbuf, axis, recv_axis):
+        raise NotImplementedError(
+            'AllToAll for same axes not supported. Please choose send_axis and recv_axis to be different.')
+        if(((axis == 0) & (recv_axis == 1)) | ((axis == 1) & (recv_axis == 0))):
```
Markus-Goetz: This is most likely not what you want to do here. You are looking for the logical operators 'and' and 'or' here instead of the bitwise & and |

heat/core/communication.py
```diff
@@ -463,6 +464,161 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
...
 
+    @classmethod
+    def alltoall_sendbuffer(cls, obj, nproc, counts=None, displs=None): #This is a temporary soltion to pass number of arguments as variable. Better to access via cls itself, but how?
+        mpi_type, elements = cls.__mpi_type_mappings[obj.dtype], torch.numel(obj)
+
+        shape = obj.shape
+        strides = [1] * len(shape)
+        strides[-1] = obj.stride()[-1]
+
+        offsets = [0]*len(shape)
```
Markus-Goetz: spaces around binary operators

heat/core/communication.py
```diff
@@ -463,6 +464,161 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
...
+
+        #Step 3: Pack short_type and long_type along the recv_axis
+        mpi_short_type=mpi_short_type.Create_vector(shape[0], 1, strides[0]).Create_resized(0, send_elements[-1] * obj.stride()[1] * obj.element_size())
+        mpi_short_type.Commit()
+        mpi_long_type=mpi_long_type.Create_vector(shape[0], 1, strides[0]).Create_resized(0, send_elements[0] * obj.stride()[1] * obj.element_size())
+        mpi_long_type.Commit()
+
+        #Step 4: Prepare sencounts, senddispls and sendtypes for alltoallw
+        # to each process 1 element (=sendcount) of the custom prepared long or short type will be send
+        sendcount = [1]*nproc
```
Markus-Goetz: spaces

heat/core/communication.py
```diff
@@ -463,6 +464,161 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
...
+
+        #Create short_Type from the last entry of send_elements
+        mpi_short_type = mpi_type.Create_vector(send_elements[-1], 1, strides[1]).Create_resized(0, offsets[1])
+        mpi_short_type.Commit()
+        #Create long_Type from the first entry of send_elements (wraps one more mpi_type vector than short_Type
+        mpi_long_type = mpi_type.Create_vector(send_elements[0], 1, strides[1]).Create_resized(0, offsets[1])
+        mpi_long_type.Commit()
+
+        #Step 3: Pack short_type and long_type along the recv_axis
+        mpi_short_type=mpi_short_type.Create_vector(shape[0], 1, strides[0]).Create_resized(0, send_elements[-1] * obj.stride()[1] * obj.element_size())
```
Markus-Goetz: I am certain that chaining type creation operations are fine in Fortran. I am not sure whether this is the case for C/C++ implementation. Since we bind against them that might be a problem for certain MPI stacks. Need to keep an eye on this. Could lead some bugs

heat/core/communication.py
Cdebus: The problem is nproc (the number of processes). For the other functions, you can just do self.size. But with the classmethod this doesn't work. Is there some way to do cls.size() ? otherwise it has to be passed as a parameter. I believe counts and displs are not even needed

heat/core/communication.py
Cdebus: yes, but does it WORK in this setting? when the buffer is not specifically prepared?

heat/core/communication.py
Cdebus: this is my C++ speaking :laughing:

heat/core/communication.py
Markus-Goetz: You can change it to be a simple instance method, aka

def alltoall_sendbuffer(self, obj, ...):

heat/core/communication.py
Cdebus: ? But this is what is done in all the other cases as well, you always did
mpi_type = mpi_type.Create_vector
Plus, so far it works ^^

heat/core/communication.py
Markus-Goetz: Well you could come up with weird edge cases of having singular items for example. Why one would do this is questionable, but it would also not be problematic to simply allow for MPI_IN_PLACE?

heat/core/communication.py
Cdebus: Will do, but just as a side-note:
This happend also in the as_buffer function for the offsets

Cdebus:
By the way, I put the alltoall-specific functions (Buffer preparation) dictly above the _alltoall_w function (where they are needed). However, this kind of breaks the design pattern for communication.py. Should I move the buffer preparation functions up to as_buffer?

heat/core/communication.py
Markus-Goetz: okay so we leave it, maybe add short comment of the sorts

# XXX: might not work for all MPI stacks. Might require multiple type commits or so

This will ease the search for other future developers

Markus-Goetz:
I think this would be a good idea. Modern IDEs support jumping to specific functions

heat/core/communication.py
Cdebus: oh ok, wasn't sure (I still don't get a lot of python concepts (like classmethod )). This might also make sense since these functions are only used for all_to_all (unlike the very versatile as_buffer)

Cdebus: argh, but can I then still access __mpi_type_mappings? 
   @classmethod
    def alltoall_sendbuffer(cls, obj, nproc, counts=None, displs=None): #This is a temporary soltion to pass number of arguments as variable. Better to access via cls itself, but how?
        mpi_type, elements = cls.__mpi_type_mappings[obj.dtype], torch.numel(obj)

heat/core/communication.py
Cdebus: Well, no. 
"Some intracommunicator collective operations do not support the ``in place'' option (e.g., MPI_ALLTOALLV)" (source: https://www.mcs.anl.gov/research/projects/mpi/mpi-standard/mpi-report-2.0/node145.htm)
I also came up with an examplary case where IN_PLACE would crash for alltoall:
If you have a [4,5] Tensor that is split among axis 0 across 4 processes, and you do alltoall from axis zero to axis 1, then every rank as as sendbuffer with 5 elements, but for the receivebuffer, rank 0 needs 8 elements, while ranks 1 to 3 need 4 elements.

heat/core/communication.py
Markus-Goetz: Yes, with self.__mpi_type_mappings

heat/core/communication.py
Markus-Goetz: Fine, I see the point, let's drop it

heat/core/communication.py
```diff
@@ -460,6 +556,92 @@ def Iallgatherv(self, sendbuf, recvbuf, send_axis=0, recv_axis=None):
...
+        -------
+        exit code: MPI4py alltoall_w
+        """
+        if (recv_axis == send_axis):
+            raise NotImplementedError('AllToAll for same axes not supported. Please choose send_axis and recv_axis to be different.')
+        if ( (send_axis == None) or (recv_axis == None) ):
+            raise NotImplementedError('AllToAll needs send_axis and recv_axis to be specified but was send_axis = {}, recv_axis = {}. Please set send_axis and recv_axis'.format(send_axis, recv_axis))
+
+        # ToDo: The option to explicitely specify the counts and displacements to be send still needs propper implementation
+        send_counts, send_displs, recv_counts, recv_displs = None, None, None, None,
```
Cdebus: This will be a seperate issue, since it concerns also the general scatter_like funcitons (and the mpi_type_and_elements function)

heat/core/communication.py
```diff
@@ -242,6 +245,99 @@ def as_buffer(cls, obj, counts=None, displs=None):
 
         return [cls.as_mpi_memory(obj), elements, mpi_type]
 
+    def alltoall_sendbuffer(self, obj):
```
Cdebus: @d1saster can you check if the comments on the workflow are understandable?

heat/core/communication.py
Cdebus: See Issue #371

Issue #323, opened 2019-07-03
Environment for better unit testing


TheSlimvReal:
Currently a lot of our test cases represent only some very basic input matrices (e.g. using `torch.ones` or `torch.arange`). This is good for asserting the result afterwards but many edge cases are undiscovered and we miss our chance to test with big input data for what heat is actually made for. 
Therefore it would be good to create a test environment, where we can create tests in a random manner. 
One way to achieve this would be to first choose a implementation of the algorithm we want to test against (in our case probably `numpy`).
Now we only have to specify a shape of the input tensor. This tensor will then be created with random values, split along one (random?) axis, the heat implementation of the algorithm to be tested is applied to the tensor and afterwards the tensor is resplit to `None`. This way we can assert the result to be close to the `numpy` result. This would not need much user input for creating a test and will eventually cover many different test cases. The downside is that we can't check if the tensor is correctly distributed after the calculation and before the resplit to `None` and we are dependent on the random number generation.

PR #365, opened 2019-08-29
Test Suite

TheSlimvReal:
## Description
Target issue #323 

This PR introduces a test suite for improved unit testing of our functions.
Most test cases are quite trivial or don't involve the splitting of tensors, because for different number of nodes the result looks differently. Furthermore many edge cases are left untested. 
This test suits helps in creating better tests by reducing boilerplate code for comparing an heat array with the expected numpy result.
For even more simplicity, heat functions that should perform equivalent to numpy functions can be tested in one single line on random numbers or a predefined array. 

The general idea is to compare a (split) heat tensor with a numpy tensor without having to worry about differing split axis or number of nodes.

Feel free to tell me which functionality is missing that would reduce your time writing unit tests.

If this PR gets merged the next step would be to apply this test suite to all existing functions.

To get an idea how unit tests with this test suite might look lake check out the test_basic_test.py file or have a look at the documentation examples.

## Type of change


heat/core/manipulations.py
```diff
@@ -785,28 +785,19 @@ def unique(a, sorted=False, return_inverse=False, axis=None):
...
-        # Local results can now just be added together
-        if axis is None:
-            # One dimensional vectors can't be distributed -> no split
-            output_dim = [uniques_buf.sum().item()]
-            recv_axis = 0
-        else:
-            output_dim = list(lres.shape)
-            output_dim[0] = uniques_buf.sum().item()
-            recv_axis = a.split
+        #
```
Markus-Goetz: remove

heat/core/manipulations.py
```diff
@@ -895,23 +890,16 @@ def unique(a, sorted=False, return_inverse=False, axis=None):
         a.comm.Bcast(indices, root=max_pos)
         gres = local_data[indices.tolist()]
 
-        is_split = a.split
+        # is_split = a.split
```
Markus-Goetz: remove

heat/core/operations.py
```diff
@@ -172,8 +172,23 @@ def __local_op(operation, x, out, **kwargs):
         if kwargs_dtype:
             del kwargs['dtype']
         result = operation(x._DNDarray__array.type(torch_type), **kwargs)
-        return dndarray.DNDarray(result, tuple(result.shape), promoted_type if kwargs_dtype is None else kwargs_dtype,
-                                 x.split, x.device, x.comm)
+        split = x.split
```
Markus-Goetz: For what local operations would this be the case? The ones that I can think of at the moment are just element-wise operations. Everything else is either a binary operation and/or a reduction.

heat/core/manipulations.py
coquelin77: remove dead code / is this ever set again? I see it used in the array call later

heat/core/statistics.py
```diff
@@ -622,10 +622,9 @@ def reduce_means_elementwise(output_shape_i):
             return dndarray.DNDarray(ret, tuple(ret.shape), types.canonical_heat_type(ret.dtype), None, x.device, x.comm)
         else:
             # if x is distributed and no axis is given: return mean of the whole set
-            if x.lshape[x.split] != 0:
-                mu_in = operations.__local_op(torch.mean, x, out=None)
-            else:
-                mu_in = 0
+            mu_in = operations.__local_op(torch.mean, x, out=None)
```
Markus-Goetz: this should be a local op

heat/core/tests/test_manipulations.py
```diff
@@ -630,6 +631,7 @@ def test_unique(self):
 
         exp_axis_one = ht.array([rank] * size, dtype=ht.int32).expand_dims(1)
         res = ht.unique(split_one, sorted=True, axis=1)
+        print('exp', exp_axis_one, 'res', res)
```
Markus-Goetz: remove

heat/core/statistics.py
```diff
@@ -1126,14 +1125,14 @@ def reduce_vars_elementwise(output_shape_i):
...
         else:  # case for full matrix calculation (axis is None)
-            if x.lshape[x.split] != 0:
-                mu_in = operations.__local_op(torch.mean, x, out=None)
-                var_in = operations.__local_op(torch.var, x, out=None, unbiased=bessel)
-                if torch.isnan(var_in._DNDarray__array):
-                    var_in = 0.0
-            else:
-                mu_in = 0
-                var_in = 0
+            mu_in = operations.__local_op(torch.mean, x, out=None)
```
Markus-Goetz: should also probably not be a local op

heat/core/tests/test_exponential.py
```diff
@@ -45,6 +45,11 @@ def test_exp(self):
         with self.assertRaises(TypeError):
             ht.exp('hello world')
 
+    def test_exp_with_split(self):
```
coquelin77: can you add a correctness test, a split test, and a dtype test?

heat/core/tests/test_manipulations.py
```diff
@@ -605,8 +605,9 @@ def test_unique(self):
         torch_array = torch.arange(size, dtype=torch.int32).expand(size, size)
         split_zero = ht.array(torch_array, split=0)
 
-        exp_axis_none = ht.arange(size, dtype=ht.int32)
+        exp_axis_none = ht.array([rank], dtype=ht.int32)
         res = split_zero.unique(sorted=True)
+        print('exp', exp_axis_none, 'res', res)
```
coquelin77: remove

heat/core/tests/test_manipulations.py
coquelin77: remove

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+    def device(self):
+        return BasicTest.__device
+
+    def get_rank(self):
+        return self.comm.rank
+
+    def get_size(self):
+        return self.comm.size
+
+    def assert_array_equal(self, heat_array: dndarray.DNDarray, numpy_array):
```
Markus-Goetz: if we start type annotations be consistent/complete with them

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+from heat.core import dndarray, MPICommunication, MPI, types, factories
+import heat as ht
+
+import numpy as np
+import torch
+
+
+class BasicTest(TestCase):
+
+    __comm = MPICommunication()
```
Markus-Goetz: This does not seem like a good idea. The comm might be dependent on the input variable

heat/core/tests/test_suites/basic_test.py
coquelin77: i prefer to use torch for many things, can you add a torch_tensor here. and can you make it an option to choose between the two of them?

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+        self.assertEqual(heat_array.lshape, numpy_array[slices].shape,
+                         'Local shapes do not match. '
+                         'Got {} expected {}'.format(heat_array.lshape, numpy_array[slices].shape))
+        local_numpy = heat_array._DNDarray__array.numpy()
+
+        # Array is distributed correctly
+        equal_res = np.array(compare_func(local_numpy, numpy_array[slices]))
+        self.comm.Allreduce(MPI.IN_PLACE, equal_res, MPI.LAND)
+        self.assertTrue(equal_res, 'Local tensors do not match the corresponding numpy slices.')
+        self.assertEqual(local_numpy.dtype, numpy_array.dtype,
```
Markus-Goetz: It does not need to necessarily equal but potential float_eq

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+            ht_array = factories.array(torch_tensor, split=i, dtype=dtype, device=self.device, comm=self.comm)
+            ht_res = heat_func(ht_array, **heat_args)
+            if distributed_result:
+                self.assert_array_equal(ht_res, np_res)
+            else:
+                self.assertTrue(np.array_equal(ht_res._DNDarray__array.numpy(), np_res))
+
+    def _create_random_array(self, shape):
+        # Ensure all processes have the same initial array
+        if self.get_rank() == 0:
```
Markus-Goetz: Should do this here with our random module by setting the seed twice

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+
+        self.assertIsInstance(heat_array, dndarray.DNDarray,
+                              'The array to test was not a instance of ht.DNDarray. '
+                              'Instead got {}.'.format(type(heat_array)))
+        self.assertIsInstance(numpy_array, np.ndarray,
+                              'The array to test against was not a instance of numpy.ndarray. '
+                              'Instead got {}.'.format(type(numpy_array)))
+        self.assertEqual(heat_array.shape, numpy_array.shape,
+                         'Global shapes do not match. Got {} expected {}'.format(heat_array.shape, numpy_array.shape))
+        split = heat_array.split
```
Markus-Goetz: device tests should happen here as well. for both:

a) our passed device value
b) pytorch's internal device value

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def get_size(self):
+        return self.comm.size
+
+    def assert_array_equal(self, heat_array: dndarray.DNDarray, numpy_array):
+        """
+        Check if the heat_array is equivalent to the numpy_array. Therefore first the split heat_array is compared to
+        the corresponding numpy_array slice locally and second the heat_array is combined and fully compared with the
+        numpy_array.
+        Note if the heat array is split it also needs to be balanced.
```
coquelin77: a secondary note should be added here that the standard numpy dtype is float64. this is different from torch/heat where the default if float32.  essentially, the numpy arrays used in the function should be created with the proper dtype

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+        self.assertIsInstance(heat_array, dndarray.DNDarray,
+                              'The array to test was not a instance of ht.DNDarray. '
+                              'Instead got {}.'.format(type(heat_array)))
+        self.assertIsInstance(numpy_array, np.ndarray,
+                              'The array to test against was not a instance of numpy.ndarray. '
+                              'Instead got {}.'.format(type(numpy_array)))
+        self.assertEqual(heat_array.shape, numpy_array.shape,
+                         'Global shapes do not match. Got {} expected {}'.format(heat_array.shape, numpy_array.shape))
+        split = heat_array.split
+        offset, local_shape, slices = heat_array.comm.chunk(heat_array.gshape, split)
```
Markus-Goetz: comm checks should happen here as well

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+            The function that is to be tested
+        numpy_func: function
+            The numpy implementation of an equivalent function to test against
+        distributed_result: bool, optional
+            Specify whether the result of the heat function is distributed across all nodes or all nodes have the full
+            result. Default is True.
+        heat_args: dictionary, optional
+            The keyword arguments that will be passed to the heat function. Array and split function don't need to be
+            specified. Default is {}.
+        numpy_args:
```
coquelin77: dictionary, optional

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+            else:
+                self.assertTrue(np.array_equal(ht_res._DNDarray__array.numpy(), np_res))
+
+    def _create_random_array(self, shape):
+        # Ensure all processes have the same initial array
+        if self.get_rank() == 0:
+            array = np.random.randn(*shape)
+        else:
+            array = np.empty(shape)
+        self.comm.Bcast(array, root=0)
```
coquelin77: couldnt this be avoided by setting a manual seed? then every array will be the same on all nodes anyway

heat/core/tests/test_suites/basic_test.py
```diff
@@ -0,0 +1,173 @@
...
+            torch_tensor = torch.from_numpy(tensor.copy())
+            np_array = tensor
+        elif isinstance(tensor, torch.Tensor):
+            torch_tensor = tensor
+            np_array = tensor.numpy().copy()
+        else:
+            raise TypeError('The input tensors type must be one of [tuple, list, ' +
+                            'numpy.ndarray, torch.tensor] but is {}'.format(type(tensor)))
+
+        np_res = numpy_func(np_array, **numpy_args)
```
coquelin77: dtype (float64 vs float32)

heat/core/operations.py
TheSlimvReal: Yes the target of this change are the element wise operations. It might be good to have them separated. This way no communication would be necessary because the shape stays the same

heat/core/statistics.py
TheSlimvReal: What do you mean?

heat/core/statistics.py
TheSlimvReal: Would you like to have this cases of "wrong" usage of the local op's removed?

heat/core/tests/test_suites/basic_test.py
TheSlimvReal: So the comm should rather be passed via the assert functions? For the array_equal function I understand the point but for the func_equal, the input values don't have a comm.

heat/core/tests/test_suites/basic_test.py
TheSlimvReal: Is this a general rule or can there be changes or single functions that differ in their behaviour? If the user wants to be certain, he can always pass args where the dtype is explicitly defined. I just don't want to make any general assumptions about the behaviour of the functions which might not be correct and result in unexpected results.

coquelin77: yes the dtype can change after a function. my concern is that the types will not match up because of the difference in what is the standard types of the packages. 
i think that the thing to do would be to create the numpy array with the dtype of the heat array, and if no dtype is available then it should be created with dtype=float32. that way they have the same starting point. and i believe that numpy will only increase the precision if it is necessary. i.e. type changes which occur in the course of the function will remain at the same precision.

heat/core/tests/test_suites/basic_test.py
TheSlimvReal: Do you mean the devices should be tested to be equal?

heat/core/tests/test_suites/basic_test.py
TheSlimvReal: For what should the comm be checked? That he is present and correct?

heat/core/tests/test_suites/basic_test.py
TheSlimvReal: So should this just be left out and just rely on the equality check of the tensor's values

TheSlimvReal:
Most of the requested changes got fixed/discussed. Would be great if you could have a new look.

heat/core/tests/test_suites/basic_test.py
TheSlimvReal: The user is always able to define this using the `numpy_args`. I want to keep the function general and therefore not make too many assumptions about what the user actually wants.

coquelin77:
bump, black error

TheSlimvReal:
Blake got fixed

Issue #226, opened 2019-04-09
Implement bitwise binary operations


Markus-Goetz:
Implement all binary bitwise operations, e.g. bitwise_and, bitwise_xor, ...

Provide operator overloads where necessary, i.e. &, | and ^

PR #366, opened 2019-08-29
Bit-wise AND OR XOR

mtar:
## Description

Fixes: #226 

Changes proposed:
- bitwise_and, bitwise_or, bitwise_xor in arithmetics
- operator overloading of the methods in DNDArray

## Type of change
The added functions call the new binary_bit_op function giving the method name.
Quite similar to binary_op, but calls the respective operator in torch.array and checks for same integer or boolean types in both tensors.

- [ ] This change requires a documentation update

Have you handled and tested all split configurations?
No


heat/core/operations.py
```diff
@@ -11,6 +11,127 @@
 
 __all__ = []
 
+def __binary_bit_op(method, t1, t2):
```
Markus-Goetz: I am not 100% sure it is necessary to have this as a separate function in contrast to __binary_op(). It also seems that it is almost a verbatim copy of the function below. Why would we needs this except the type checking and casting purposes? It would be more meaningful from a SE perspective to factor out the common parts from __binary_op and __bitwise_binary_op

heat/core/operations.py
```diff
@@ -11,6 +11,127 @@
...
+            except (ValueError, TypeError,):
+                raise TypeError('Data type not supported, input was {}'.format(type(t2)))
+
+        elif isinstance(t2, dndarray.DNDarray):
+            if t1.split is None:
+                t1 = factories.array(t1, split=t2.split, copy=False, comm=t1.comm, device=t1.device, ndmin=-t2.numdims)
+            elif t2.split is None:
+                t2 = factories.array(t2, split=t1.split, copy=False, comm=t2.comm, device=t2.device, ndmin=-t1.numdims)
+            elif t1.split != t2.split:
+                # It is NOT possible to perform binary operations on tensors with different splits, e.g. split=0
```
Markus-Goetz: Have an issue documenting the fact that we need to do something here still as soon as resplit is implemented (soon done)

heat/core/operations.py
mtar: You're right. It is mostly a copy of the function and I am reluctant because of that, too. However, I couldn't find a really satisfying solution at the moment.
What's different is the not desired type casting at the end of each of the if branches at the first half of the function. It still may be not be the best solution, but I'll try to factor each branch out for instance, scalar and mixed pairs to reduce some code redundancy. 
Also, the function calls at the end uses the overloaded operators of the torch.tensor class. I don't think I can do much there.

Markus-Goetz:
Minor remark for the future. Please have you branchname of the form:

feature/<issue_number>-short-description
or
bug/<issue_number>-short description

No worries right now. Will make it easier for us in the future to maybe get statistics across the branches. Thanks

Markus-Goetz:
So what is the status on this one? Do you think there is a way to refactor the respective methods or not?

mtar:
I rearranged it. The separate branches are put in respective functions now.

Issue #381, opened 2019-09-13
Get rid of UserWarning in manipulations.py


Markus-Goetz:
**Description**
The slicing in the sort() implementation currently produces errors regarding attached tensor copies. Ensure that the behavior is a) not faulty and b) is not issuing UserWarnings

**To Reproduce**
1. Execute unittests

**Expected behavior**
Attached but copyable slice or detached tensor copy without UserWarning

coquelin77:
this tracks back to the binary_op in the following lines. 

```python
if t1.split is not None:
    if t1.shape[t1.split] == 1 and t1.comm.is_distributed():
         warnings.warn('Broadcasting requires transferring data of first operator between MPI ranks!')
         if t1.comm.rank > 0:
             t1._DNDarray__array = torch.zeros(t1.shape, dtype=t1.dtype.torch_type())
         t1.comm.Bcast(t1)

if t2.split is not None:
    if t2.shape[t2.split] == 1 and t2.comm.is_distributed():
         warnings.warn('Broadcasting requires transferring data of second operator between MPI ranks!')
         if t2.comm.rank > 0:
             t2._DNDarray__array = torch.zeros(t2.shape, dtype=t2.dtype.torch_type())
         t2.comm.Bcast(t2)
```

PR #397, opened 2019-10-15
changed sort to not produce warnings

TheSlimvReal:
## Description

The sort function used a torch functionality that produces a warning. Torch proposed a different way which was now implemented.

Fixes: #381

Changes proposed:
- using clone and detach instead of creating a new tensor

## Type of change

Select relevant options.

[x] yes [ ] no
[ ] yes [x] no
[ ] yes [x] no
[ ] yes [x] no


coquelin77:
do we need to detach it as well? or is cloning enough? in other words, why do we need to remove the cloned tensor from the graph?

PR #412, opened 2019-11-14
diag() and diagonal() feature

TheSlimvReal:
## Description

This PR targets issue #242.

The returned array of `diag()` in case of creating a diagonal array, will be split the same as the input array (`split=None` or `split=0`).

## Type of change

Select relevant options.

[x] yes - [ ] no
[ ] yes [x] no
[ ] yes [x] no
[ ] yes [x] no

### DIFF
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -342,7 +342,7 @@ def diag(a, offset=0):
         If a is a n-dimensional array with n > 1 the diagonal entries will be returned in an n-1 dimensional array.
     offset: int, optional
         The offset from the main diagonal.
-        Offset greater zero means above the main diagonal, below zero is below the main diagonal.
+        Offset greater than zero means above the main diagonal, smaller than zero is below the main diagonal.
 
     Returns
     -------
@@ -354,11 +354,11 @@ def diag(a, offset=0):
     >>> import heat as ht
     >>> a = ht.array([1, 2])
     >>> ht.diag(a)
-    array([[1, 0],
+    tensor([[1, 0],
            [0, 2]])
 
     >>> ht.diag(a, offset=1)
-    array([[0, 1, 0],
+    tensor([[0, 1, 0],
            [0, 0, 2],
            [0, 0, 0]])
 
@@ -366,7 +366,7 @@ def diag(a, offset=0):
     True
     >>> a = ht.array([[1, 2], [3, 4]])
     >>> ht.diag(a)
-    array([1, 4])
+    tensor([1, 4])
     """
     if len(a.shape) > 1:
         return diagonal(a, offset=offset)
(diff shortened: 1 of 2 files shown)


TheSlimvReal:
I also made some small changes in the `BasicTest` file because this was the first time I used it in production. I hope it's okay to push this changes together.

heat/core/manipulations.py
```diff
@@ -316,6 +329,169 @@ def concatenate(arrays, axis=0):
...
+        padding = factories.empty(
+            (offset,), dtype=a.dtype, split=None, device=a.device, comm=a.comm
+        )
+        a = concatenate((a, padding))
+    elif offset < 0:
+        padding = factories.empty(
+            (abs(offset),), dtype=a.dtype, split=None, device=a.device, comm=a.comm
+        )
+        a = concatenate((padding, a))
+    if not a.is_balanced():
```
coquelin77: balance_ will call is_balanced before doing anything so this is optional. you can also just call balance_ here

heat/core/manipulations.py
```diff
@@ -316,6 +329,169 @@ def concatenate(arrays, axis=0):
...
+        a.balance_()
+
+    if offset > 0:
+        indices_x = np.arange(0, min(a.lshape[0], gshape[0] - off - offset))
```
coquelin77: please change to torch, numpy doesnt play well with GPUs (it should be the same call just with torch instead of np)

heat/core/manipulations.py
```diff
@@ -316,6 +329,169 @@ def concatenate(arrays, axis=0):
...
+    -------
+    res: ht.DNDarray
+        The extracted diagonal or the constructed diagonal array
+
+    Examples
+    --------
+    >>> import heat as ht
+    >>> a = ht.array([1, 2])
+    >>> ht.diag(a)
+    array([[1, 0],
```
coquelin77: why does these say array instead of tensor?

heat/core/manipulations.py
```diff
@@ -316,6 +329,169 @@ def concatenate(arrays, axis=0):
...
+    Extract a diagonal of an n-dimensional array with n > 1.
+    The returned array will be of dimension n-1.
+
+    Parameters
+    ----------
+    a: ht.DNDarray
+        The array of which the diagonal should be extracted.
+    offset: int, optional
+        The offset from the main diagonal.
+        Offset greater that zero means above the main diagonal, smaller zero is below the main diagonal.
```
coquelin77: smaller than zero*

heat/core/manipulations.py
```diff
@@ -316,6 +329,169 @@ def concatenate(arrays, axis=0):
...
+    else:
+        split = len(shape) - 1
+
+    if a.split is None or a.split not in (dim1, dim2):
+        result = torch.diagonal(a._DNDarray__array, offset=offset, dim1=dim1, dim2=dim2)
+    else:
+        vz = 1 if a.split == dim1 else -1
+        off, _, _ = a.comm.chunk(a.shape, a.split)
+        result = torch.diagonal(a._DNDarray__array, offset=offset + vz * off, dim1=dim1, dim2=dim2)
+    return dndarray.DNDarray(result, shape, a.dtype, split, a.device, a.comm)
```
coquelin77: should use `factories.array` instead of the DNDarray constructor. this is to handle more of the shapes and stuff like that.

heat/core/tests/test_manipulations.py
```diff
@@ -315,6 +316,273 @@ def test_concatenate(self):
...
+
+        if rank == 0:
+            data = torch.ones(size, dtype=torch.int32)
+        else:
+            data = torch.empty(0, dtype=torch.int32)
+        a = ht.array(data, is_split=0)
+        res = ht.diag(a)
+        self.assertTrue(
+            torch.equal(res[rank, rank]._DNDarray__array, torch.tensor(1, dtype=torch.int32))
+        )
```
coquelin77: please add a dtype test here to ensure that the dtype is not changed by the function, and that it handles different dtypes

heat/core/tests/test_manipulations.py
```diff
@@ -315,6 +316,273 @@ def test_concatenate(self):
...
+
+        self.assert_func_equal(
+            (5, 4, 3, 2),
+            heat_func=ht.diagonal,
+            numpy_func=np.diagonal,
+            heat_args={"dim1": 1, "dim2": 2},
+            numpy_args={"axis1": 1, "axis2": 2},
+        )
+
+        self.assert_func_equal(
```
coquelin77: do these functions test multiple types? i dont remember. if so then everything is all good, but if not then please add a couple tests with different dtypes

heat/core/manipulations.py
TheSlimvReal: I just realized that we are a little inconsistent in the documentation. Some examples use `array`, some use `tensor`. I will fix mine now but we probably should also make a PR that fixes all of the other functions to be in line with what actually will be displayed by the functions.

coquelin77:
tests failing, bump

heat/core/tests/test_manipulations.py
TheSlimvReal: I added that, see #420

TheSlimvReal: I added that, see #420

PR #415, opened 2019-11-15
improve GPU support

mtar:
## Description

This fix enables the usage of GPU-tensors in heat if ht.use_device is globally set to "gpu". The existing unit tests are modified to test the cpu or gpu only. Also, tensors on locally set devices can be tested.

Fixes: #

Changes proposed:
– unit tests depend on selected device (Environment variable DEVICE=gpu for global, lgpu for locally defined gpu tensors, cpu otherwise)
– additional copy operations GPU <-> CPU for not cuda-aware MPI
– smaller fixes on some core functions for accepting gpu tensors

## Type of change
• Setting device parameter in torch functions
• bugfixes
• test modifications

Select relevant options.

[ ] New feature (non-breaking change which adds functionality)
[ ] Documentation update

[x] yes [ ] no
[ ] yes [x] no
[ ] yes [x] no
[ ] yes [ ] no

### DIFF
--- a/heat/core/communication.py
+++ b/heat/core/communication.py
@@ -557,9 +557,6 @@ class MPICommunication(Communication):
         recvbuf: Input Receivebuffer
         axis: concatenation axis: The axis along which sendbuf is packed and along which recvbuf puts together individual chunks
         """
-
-        sbuf, rbuf, recv_axis_permutation = None, None, None
-
         # dummy allocation for *v calls
         # ToDO: Propper implementation of usage
         send_counts, send_displs, recv_counts, recv_displs = None, None, None, None
@@ -603,6 +600,8 @@ class MPICommunication(Communication):
             recv_axis_permutation = list(range(recvbuf.ndimension()))
             recv_axis_permutation[0], recv_axis_permutation[axis] = axis, 0
             recvbuf = recvbuf.permute(*recv_axis_permutation)
+        else:
+            recv_axis_permutation = None
 
         sbuf = sendbuf if CUDA_AWARE_MPI or not isinstance(sendbuf, torch.Tensor) else sendbuf.cpu()
         rbuf = recvbuf if CUDA_AWARE_MPI or not isinstance(recvbuf, torch.Tensor) else recvbuf.cpu()
@@ -624,11 +623,6 @@ class MPICommunication(Communication):
 
         # perform the scatter operation
         exit_code = func(mpi_sendbuf, mpi_recvbuf, **kwargs)
-
-        # undo the recvbuf permutation and assign the temporary buffer to the original recvbuf
-        # if axis != 0:
-        # rbuf = rbuf.permute(*recv_axis_permutation)
-        # original_recvbuf.set_(recvbuf.storage(), recvbuf.storage_offset(), recvbuf.shape, recvbuf.stride())
 
         return exit_code, sbuf, rbuf, original_recvbuf, recv_axis_permutation
 
@@ -788,11 +782,6 @@ class MPICommunication(Communication):
 
             # perform the scatter operation
             exit_code = func(mpi_sendbuf, mpi_recvbuf, **kwargs)
-            # undo the recvbuf permutation and assign the temporary buffer to the original recvbuf
-            # if recv_axis != 0:
-            #    recvbuf = recvbuf.permute(*recv_axis_permutation)
-            #    original_recvbuf.set_(recvbuf.storage(), recvbuf.storage_offset(), recvbuf.shape, recvbuf.stride())
-
         # slightly more difficult situation, send and receive buffer need custom datatype preparation;
         # operation is performed via alltoallw
         else:
@@ -1098,7 +1087,6 @@ class MPIRequest:
             if self.permutation is not None:
                 self.recvbuf = self.recvbuf.permute(self.permutation)
             self.tensor.copy_(self.recvbuf)
-        return
 
     def wait(self, status=None):
         self.handle.wait(status)
@@ -1111,7 +1099,6 @@ class MPIRequest:
             if self.permutation is not None:
                 self.recvbuf = self.recvbuf.permute(self.permutation)
             self.tensor.copy_(self.recvbuf)
-        return
 
     def __getattr__(self, name):
         """
--- a/heat/core/communication.py
+++ b/heat/core/communication.py
@@ -225,8 +225,8 @@ class MPICommunication(Communication):
         mpi_memory : MPI.memory
             The MPI memory objects of the passed tensor.
         """
-        # in case of GPUs, the memory has to be copied to host memory if CUDA-aware MPI is not supported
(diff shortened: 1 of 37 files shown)


coquelin77:
why is the coverage so low? is it possible to increase this?

mtar:
I added a "header" in the test files to select between CPU and GPU looking for the Environment variable DEVICE. On Travis, only the CPU is used and the GPU part is not covered.

heat/core/communication.py
```diff
@@ -570,32 +604,33 @@ def __allgather_like(self, func, sendbuf, recvbuf, axis, **kwargs):
...
         # perform the scatter operation
         exit_code = func(mpi_sendbuf, mpi_recvbuf, **kwargs)
 
         # undo the recvbuf permutation and assign the temporary buffer to the original recvbuf
-        if axis != 0:
-            recvbuf = recvbuf.permute(*recv_axis_permutation)
-            original_recvbuf.set_(
-                recvbuf.storage(), recvbuf.storage_offset(), recvbuf.shape, recvbuf.stride()
-            )
+        # if axis != 0:
```
coquelin77: remove dead code

heat/core/communication.py
```diff
@@ -713,24 +768,32 @@ def __alltoall_like(self, func, sendbuf, recvbuf, send_axis, recv_axis, **kwargs
...
 
             # perform the scatter operation
             exit_code = func(mpi_sendbuf, mpi_recvbuf, **kwargs)
             # undo the recvbuf permutation and assign the temporary buffer to the original recvbuf
-            if recv_axis != 0:
-                recvbuf = recvbuf.permute(*recv_axis_permutation)
-                original_recvbuf.set_(
-                    recvbuf.storage(), recvbuf.storage_offset(), recvbuf.shape, recvbuf.stride()
-                )
+            # if recv_axis != 0:
```
coquelin77: dead code

heat/core/communication.py
```diff
@@ -953,6 +1079,57 @@ def __getattr__(self, name):
...
+        if (
+            self.tensor is not None
+            and isinstance(self.tensor, torch.Tensor)
+            and self.tensor.is_cuda
+            and not CUDA_AWARE_MPI
+        ):
+            if self.permutation is not None:
+                self.recvbuf = self.recvbuf.permute(self.permutation)
+            self.tensor.copy_(self.recvbuf)
+        return
```
coquelin77: unnecessary, optional remove

heat/core/dndarray.py
```diff
@@ -51,6 +51,19 @@ def __init__(self, array, gshape, dtype, split, device, comm):
         self.__device = device
         self.__comm = comm
 
+        # handle inconsistencies between torch and heat devices
```
coquelin77: Can you add this to the `factories.array` function as well?

heat/core/dndarray.py
```diff
@@ -1414,7 +1427,10 @@ def __getitem__(self, key):
                 else:
                     gout = tuple(self.__array[key].shape)
                     if self.split is not None and self.split >= len(gout):
-                        new_split = len(gout) - 1 if len(gout) - 1 > 0 else 0
+                        if len(gout) > 0:
```
coquelin77: `new_split = len(gout) - 1 if len(gout) - 1 >= 0 else None`

heat/core/logical.py
```diff
@@ -106,7 +106,10 @@ def allclose(x, y, rtol=1e-05, atol=1e-08, equal_nan=False):
     """
     if np.isscalar(x):
         try:
-            x = factories.array([float(x)])
+            if isinstance(y, dndarray.DNDarray):
+                x = factories.array([float(x)], device=y.device)
+            else:
+                x = factories.array([float(x)])
```
coquelin77: this should still get a device. maybe make a device parameter above it to set the devices of both `x` and `y` to be the same

heat/core/logical.py
```diff
@@ -117,7 +120,10 @@ def allclose(x, y, rtol=1e-05, atol=1e-08, equal_nan=False):
 
     if np.isscalar(y):
         try:
-            y = factories.array([float(y)])
+            if isinstance(x, dndarray.DNDarray):
+                y = factories.array([float(y)], device=x.device)
+            else:
+                y = factories.array([float(y)])
```
coquelin77: see above for x

heat/core/manipulations.py
```diff
@@ -830,10 +842,10 @@ def sort(a, axis=None, descending=False, out=None):
         for idx in np.ndindex(local_sorted.shape[1:]):
             idx_slice = [slice(None)] + [slice(ind, ind + 1) for ind in idx]
 
-            send_count = send_vec[idx][rank]
+            send_count = send_vec.cpu()[idx][rank]
```
coquelin77: does this _need_ to be cpu? or can it get any device?

heat/core/communication.py
```diff
@@ -371,55 +372,62 @@ def Recv(self, buf, source=MPI.ANY_SOURCE, tag=MPI.ANY_TAG, status=None):
...
         if isinstance(buf, dndarray.DNDarray):
             buf = buf._DNDarray__array
         if not isinstance(buf, torch.Tensor):
-            return func(buf, dest, tag)
+            return func(buf, dest, tag), None
 
-        return func(self.as_buffer(buf), dest, tag)
+        # in case of GPUs, the memory has to be copied to host memory if CUDA-aware MPI is not supported
+        sbuf = buf if CUDA_AWARE_MPI else buf.cpu()
+        return func(self.as_buffer(sbuf), dest, tag), sbuf
```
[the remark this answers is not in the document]
mtar: Thats's right. For non-blocking communication, I want to keep the sendbuffer in my MPIRequest Wrapper class such that the pointer to the buffer is still valid after leaving the send_like function.

heat/core/manipulations.py
mtar: I don't know why it's there. send_vec is already on the cpu. cpu() can be removed

heat/core/manipulations.py
```diff
@@ -830,10 +842,10 @@ def sort(a, axis=None, descending=False, out=None):
...
             send_disp = [0] + list(np.cumsum(send_count[:-1]))
 
-            recv_count = send_vec[idx][:, rank]
+            recv_count = send_vec.cpu()[idx][:, rank]
```
mtar: Same as above

heat/core/logical.py
mtar: If no device is given (i.e. device=None), they have the same default device. In this case, I wanted to change the device implicitly, if a scalar is compared with a tensor.

heat/core/dndarray.py
coquelin77: this logic is already there, my bad

coquelin77:
The coverage of this branch is low because of the code which is added is tested on GPUs, something that travis does not do.

coquelin77:
The coverage of this branch is low because of the code which is added is tested on GPUs, something that travis does not do. after the manual check, the diff coverage is fine.

PR #419, opened 2019-11-22
Travis errno fix

TheSlimvReal:
## Description

After some searching around I found [this thread](https://github.com/open-mpi/ompi/issues/4948)  about the problem with the erros in the travis log:

```
Read -1, expected <someNumber>, errno =1
Read -1, expected <someNumber>, errno =1
Read -1, expected <someNumber>, errno =1
Read -1, expected <someNumber>, errno =1
Read -1, expected <someNumber>, errno =1
...
```

Trying out some of the proposed fixes, I finally found one that is actually working in our case.

Maybe someone of you knows more about what I am disabling by this option but as far as I understood it, docker has some security settings which result in this error messages. 
Another approach would be turning of these settings (as described [here](https://github.com/open-mpi/ompi/issues/4948#issuecomment-512987494)), which would maybe improve the performance but I did not try it out because I did not fully understand what I would have to do.

## Type of change

Select relevant options.

[ ] New feature (non-breaking change which adds functionality)
[ ] Documentation update

[ ] yes [x] no
[ ] yes [x] no
[ ] yes [x] no
[ ] yes [x] no

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ script:
   # Running multiple mpi process count, generate a unique coverage report for each one and merge into one report
   - |
     docker exec -t unittest /bin/bash -c '. /root/.bashrc && \
-      for i in {1..4}; do mpirun --mca btl_vader_single_copy_mechanism none --allow-run-as-root -np $i coverage run --source=heat --parallel-mode -m pytest heat/ || exit; done && \
-      mpirun --mca btl_vader_single_copy_mechanism none --allow-run-as-root -np 7 coverage run --source=heat --parallel-mode -m pytest heat/ && \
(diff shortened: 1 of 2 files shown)


TheSlimvReal:
There was another warnings message that looked something like the following:

```
[ef810dfd1a55:00427] 3 more processes have sent help message help-mpi-btl-base.txt / btl:no-nics

[ef810dfd1a55:00427] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

I found [this workaround](https://github.com/open-mpi/ompi/issues/1393#issuecomment-187980473). 

Also for this warning I am not quite sure what the real problem is and compared to the previous problem, this one did not spam the log with hundreds of messages. So I am not sure if this fix is necessary (even though I know implemented it).

PR #420, opened 2019-11-27
Different dtypes testing

TheSlimvReal:
## Description

After I used the TestSuite for the first time in production I found some useful functionalities to extend the module with.

As @coquelin77 requested, the TestSuite is now testing on multiple dtypes which can also be defined by the user.
Internally it used `np.random.randn` for float values and `np.random.randint` for integer values.
The latter one also accepts arguments for upper and lower bound of the random values.

Added to that, debugging messages were made more precise about when a exception is risen.

## Type of change

Select relevant options.

[ ] Bug fix (non-breaking change which fixes an issue)
[x] New feature (non-breaking change which adds functionality)
[ ] Documentation update

[x] yes [ ] no
[ ] yes [x] no
[ ] yes [x] no
[ ] yes [x] no


heat/core/tests/test_suites/basic_test.py
```diff
@@ -118,36 +128,111 @@ def assert_array_equal(self, heat_array, expected_array):
...
+        self,
+        tensor,
+        heat_func,
+        numpy_func,
+        heat_args=None,
+        numpy_args=None,
+        distributed_result=True,
+    ):
+        """
+        This function will create random tensors of the given shape with different data types.
```
coquelin77: these docs are the same as for assert_func_equal. are these functions the same?

heat/core/tests/test_suites/basic_test.py
```diff
@@ -209,9 +293,50 @@ def assert_func_equal(
             else:
                 self.assertTrue(np.array_equal(ht_res._DNDarray__array.numpy(), np_res))
 
-    def _create_random_array(self, shape):
+    def _create_random_array(self, shape, dtype=np.float64, low=-10000, high=10000):
```
coquelin77: can you note that this creates a numpy array? others might misjudge it by its name, maybe just change array to np_array? just a thought

coquelin77:
also can you trigger a new coverage run?

TheSlimvReal:
I removed the communicator tests in favour of #422. Once this discussion comes to a conclusion, the comm test might be added again.

heat/core/tests/test_suites/basic_test.py
TheSlimvReal: That was a mistake, I updated the docstring

heat/core/tests/test_suites/basic_test.py
TheSlimvReal: I changed the name to `__create_random_np_array`

PR #428, opened 2019-12-09
Features/lshape map+redistribute

coquelin77:
## Description

The balance function now uses a redistribute function. The logic is the same as it was before, but it can now be given an arbitrary shape to match.

Fixes: None, updates balance function in view of purposely unbalancing tensors for other operations

Changes proposed:
- added redistribute function to adjust lshapes
- added `create_lshape_map` function to generate the lshape of a function
- balance function now uses the redistribute function

## Type of change

Select relevant options.

- [x] Documentation update

[x] yes [ ] no
[ ] yes [x] no
[x] yes [ ] no
[ ] yes [x] no


heat/core/tests/test_dndarray.py
```diff
@@ -298,6 +306,62 @@ def test_numpy(self):
...
+
+            with self.assertRaises(TypeError):
+                st.redistribute_(target_map="sdfibn")
+            with self.assertRaises(TypeError):
+                st.redistribute_(lshape_map="sdfibn")
+            with self.assertRaises(ValueError):
+                st.redistribute_(lshape_map=torch.zeros(2))
+            with self.assertRaises(ValueError):
+                st.redistribute_(target_map=torch.zeros((2, 4)))
+
```
ClaudiaComito: OK I have a problem: when I run the unittests with n>=3 processes, I get this error from test_redistribute:

> Traceback (most recent call last):
>   File "/Users/c.comito/HAF/heat/heat/core/tests/test_dndarray.py", line 328, in test_redistribute
>     st.redistribute_(target_map=target_map)
>   File "/Users/c.comito/HAF/heat/heat/core/dndarray.py", line 2197, in redistribute_
>     self.redistribute_(lshape_map=lshape_map, target_map=target_map)
>   File "/Users/c.comito/HAF/heat/heat/core/dndarray.py", line 2197, in redistribute_
>     self.redistribute_(lshape_map=lshape_map, target_map=target_map)
>   File "/Users/c.comito/HAF/heat/heat/core/dndarray.py", line 2197, in redistribute_
>     self.redistribute_(lshape_map=lshape_map, target_map=target_map)
>   [Previous line repeated 968 more times]
>   File "/Users/c.comito/HAF/heat/heat/core/dndarray.py", line 2128, in redistribute_
>     if sum(target_map[..., self.split]) != self.shape[self.split]:
>   File "/anaconda3/envs/py37/lib/python3.7/site-packages/torch/tensor.py", line 411, in __iter__
>     return iter(imap(lambda i: self[i], range(self.size(0))))
> RecursionError: maximum recursion depth exceeded in comparison
>

coquelin77: with the unit tests? mine runs fine. can you provide a failing example?

ClaudiaComito: Well, the test is the failing example. However, I only have this problem with python 3.7. I've just repeated the whole exercise on a new virtualenv with python 3.6, and the tests run fine for me as well.

coquelin77: i just changed the sum to be an inline call. can you see if that changes it?

Issue #343, opened 2019-07-22
Implement QR decomposition


coquelin77:
**Related**
Feature request for QR decomposition, can be used in #337 and #336 hopefully. Has many other uses as well

**Feature functionality**
Computes the QR decomposition of a matrix input, and returns a namedtuple (Q, R) of matrices such that input = Q R with Q being an orthogonal matrix and R being an upper triangular matrix.

This returns the QR factorization.

**Additional context**
https://pdfs.semanticscholar.org/f55e/ec06f4ced19ae71ad5452f966e56b63da551.pdf

PR #429, opened 2019-12-09
Features/343 qr

coquelin77:
Description
=========
Please include a summary of the change and which issue/s is/are fixed.

Fixes: #343 

Changes proposed
------------------------
- added tiling class
- added `tiles` property to DNDarray
- implemented QR factorization
- abstracted the c_block_setter function from matmul
- reorder functions in DNDarray

Type of change
--------------------
-> New feature (non-breaking change which adds functionality)

[x] yes [ ] no
[ ] yes [x] no
[ ] yes [x] no
[x] yes [ ] no
-> There are multiple instances of >3 indentations here. Very often this occurs because of rank selection.


coquelin77:
The reason that coverage is low here is because the coverage module does not register the jitted function used in matmul (`c_block_setter`)

coquelin77:
I ran the tests successful with 1 to 8 GPUs and CPUs. I have also included an extended tests feature for QR. the loops for this equate to the following:

```python
st_whole = torch.randn(70, 70, device=device)
for sp in range(0, 1):
        for m in range(50, 71, 1):
            for n in range(50, 71, 1):
                for t in range(1, 3):
                    st = st_whole[:m, :n].clone()
                    a_comp = ht.array(st, split=0, device=ht_device)
                    a = ht.array(st, split=sp, device=ht_device)
                    qr = a.qr(tiles_per_proc=t)
                    self.assertTrue(ht.allclose(a_comp, qr.Q @ qr.R, rtol=1e-5, atol=1e-5))
                    self.assertTrue(
                        ht.allclose(
                            qr.Q.T @ qr.Q, ht.eye(m, device=ht_device), rtol=1e-5, atol=1e-5
                        )
                    )
                    self.assertTrue(
                        ht.allclose(
                            ht.eye(m, device=ht_device), qr.Q @ qr.Q.T, rtol=1e-5, atol=1e-5
                        )
                    )
```
there tests were successfully run on 4,7, and 8 GPUs and CPUs, however they do take about 100x as long. I have successfully run the extended tests for `range(30, 100)` as well; the change to `range(50, 71)` was made to reduce the time required. To activate these tests one must add `EXTENDED_TESTS=True` before the command starting the tests. It is the same syntax as the `DEVICE` parameter as implemented by @mtar for running the tests on GPUs. As these functions have been shown to work, I propose that they be merged in 2 weeks provided that there are no objections.

Issue #433, opened 2019-12-12
Tests fail without optional dependencies


mtar:
**Description**
Some tests do not check if optional dependencies are available.

**To Reproduce**
1. Which module/class/function is affected?
test_statistics, test_kmeans, test_lasso
2. What are the circumstances under which the bug appears?
Installation without optional dependencies
3. What is the exact error-message/errorous behavious?
AttributeError: module 'heat' has no attribute 'load_hdf5'
ValueError: Unsupported file extension .h5

**Expected behavior**
All tests pass

**Version Info**
 master branch

PR #434, opened 2019-12-12
check for existing dependency in tests

mtar:
## Description
Fixes: #433

Changes proposed:
- add `if ht.io.supports_hdf5():` before using function in unit tests

## Type of change

Select relevant options.

[✔] Bug fix (non-breaking change which fixes an issue)
[✘] New feature (non-breaking change which adds functionality)
[✘] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[✘] Documentation update

[ ] yes [✘] no
[ ] yes [✘] no
[ ] yes [✘] no
[ ] yes [✘] no

### DIFF
--- a/heat/core/tests/test_statistics.py
+++ b/heat/core/tests/test_statistics.py
@@ -590,7 +590,7 @@ class TestStatistics(unittest.TestCase):
                         else:
                             self.assertEqual(res.split, z.split)
 
-        if "load_hdf5" in dir(ht):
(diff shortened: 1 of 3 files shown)


coquelin77:
I changed the file types in the tests which used the iris dataset to use csv instead of hdf5 so that also solves the problem which this addressed, however in the case that the dataset was not available as as csv your changes remain

PR #440, opened 2019-12-18
Features/343 qr tiling

coquelin77:
## Description

This is a split of the QR PR which only contains the tiling class developed for QR.
This tiling scheme creates square tiles along the diagonal of the matrix and uses these dimensions for the columns and rows

*Note:* This should be merged before the QR PR

Changes proposed:
- added the tiling class in the tiling file
- added tiles property to DNDarray

## Type of change

- [x] yes
- [ ] no

- [ ] yes
- [x] no

- [ ] yes
- [x] no

- [ ] yes
- [x] no


heat/core/dndarray.py
```diff
@@ -867,6 +880,26 @@ def create_lshape_map(self):
...
+            Default = 2 (see :func:`SquareDiagTiles <tiling.SquareDaigTiles>`)
+            the number of divisions per process,
+
+        Returns
+        -------
+        None
+        """
+        self.__tiles = tiling.SquareDiagTiles(
+            self, tiles_per_proc=tiles_per_proc
+        )  # type: tiling.SquareDiagTiles
```
Markus-Goetz: Comment left over

heat/core/tiling.py
```diff
@@ -0,0 +1,756 @@
...
+
+    # adjust for small blocks on the last diag pr:
+    last_pr_minus1 = last_dia_pr - 1 if last_dia_pr > 0 else 0
+    rem_cols_last_pr = abs(
+        min(arr.gshape) - lshape_map[..., arr.split].cumsum(dim=0)[last_pr_minus1]
+    )
+    # this is the number of rows/columns after the last diagonal on the last diagonal pr
+
+    while 1 < rem_cols_last_pr / last_tile_cols < 2:
+        # todo: determine best value for this (prev at 2)
```
Markus-Goetz: Is this an issue?

heat/core/tiling.py
```diff
@@ -0,0 +1,756 @@
+import torch
+
+from . import dndarray
+
+__all__ = ["SquareDiagTiles"]
+
+
+def _create_cols_square_diag(arr, lshape_map, tiles_per_proc):
```
Markus-Goetz: Why is this not a static/classmethod of the class below?

heat/core/tiling.py
```diff
@@ -0,0 +1,756 @@
...
+
+        Parameters
+        ----------
+        arr : DNDarray
+            the array to be tiled
+        tiles_per_proc : int, optional
+            Default = 2
+            the number of divisions per process,
+
+        Returns
```
Markus-Goetz: Remove if nothing is returned

heat/core/tiling.py
```diff
@@ -0,0 +1,756 @@
...
+                row_per_proc_list[c] = i.item()
+            except AttributeError:
+                pass
+        for c, i in enumerate(col_per_proc_list):
+            try:
+                col_per_proc_list[c] = i.item()
+            except AttributeError:
+                pass
+
+        # =========================================================================================
```
Markus-Goetz: Left over?

heat/core/tiling.py
```diff
@@ -0,0 +1,756 @@
...
+            arr.redistribute_(lshape_map=lshape_map, target_map=target_map)
+
+        row_per_proc_list = [tiles_per_proc] * arr.comm.size
+
+        last_diag_pr, col_per_proc_list, col_inds, tile_columns = _create_cols_square_diag(
+            arr, lshape_map, tiles_per_proc
+        )
+
+        # need to adjust the lshape if the splits overlap...
+        if arr.split == 0 and tiles_per_proc == 1:
```
Markus-Goetz: The if conditions here seem highly nested. Can we break them up into sub functions/a function map?

heat/core/tiling.py
```diff
@@ -0,0 +1,756 @@
...
+    def col_indices(self):
+        """
+        Returns
+        -------
+        list : list containing the indices of the tile columns
+        """
+        return self.__col_inds
+
+    @property
+    def lshape_map(self):
```
Markus-Goetz: Do we want to over either the lmap and/or the tile map via the iterator interface, aka `__iter__()`?

heat/core/tiling.py
coquelin77: It is possible that it is not worth the time to change this at all as 2 is the minimum number of columns required for QR to work and datasets that we are targeting have much much more than this on each node. I think that it is very probable that any speedup which we find by modifying this value would be specific to that dataset/node configuration. That being said, I can create an issue for it as well

heat/core/dndarray.py
coquelin77: My IDE was have troubles inferring the type of tiles here so I have it explicitly defined it with a type tag to make things run more smoothly

heat/core/tiling.py
coquelin77: both of these are torch tensors, so shouldnt the iterator be simply the torch iterator? 

in what scenarios do you anticipate the need for the iterator interface?

coquelin77:
taken care of when QR PR was merged

Issue #156, opened 2019-03-12
Multi dimensional var


coquelin77:
once multi-dimensional variance is available in pytorch the following code can be added at the end of the var function. must also change the else statement to an elif to allow for iterables and then modify the statement associated with the raised TypeError

```python
# multiple dimensions
output_shape = [output_shape[it] for it in range(len(output_shape)) if it not in axis]
if x.split in axis:
    # merge in the direction of the split
    return reduce_vars_elementwise(output_shape)
else:
    # multiple dimensions which does *not* include the split axis
    # combine along the split axis
    try:
        return tensor.array(local_op(torch.var, x, out=None, dim=axis, unbiased=bessel), split=x.split, comm=x.comm)
    except ValueError:
        return local_op(torch.var, x, out=None, dim=axis, unbiased=bessel)
```

PR #445, opened 2019-12-30
Features/mean var merge cleanup

coquelin77:
## Description

This is a cleanup of the mean and var functions. The moment merging function is rewritten to be used by both mean and var (it is also possible to expand this function to be used for higher order moments).
The var function also supports multi-dimensional axis arguments (the algorithm is the same as it is for mean)

Fixes: #156 

Changes proposed:
- combination of moment merging functions into one function
- add multi-dimensional axis support to var

## Type of change

Select relevant options.

- Documentation update

- [x] yes
- [ ] no

- [ ] yes
- [x] no

- [ ] yes
- [x] no

- [ ] yes
- [x] no

### DIFF
--- a/heat/core/statistics.py
+++ b/heat/core/statistics.py
@@ -663,6 +663,7 @@ def mean(x, axis=None):
     ----------
     x : ht.DNDarray
         Values for which the mean is calculated for.
+        The dtype of x must be a float
     axis : None, Int, iterable, defaults to None
         Axis which the mean is taken in. Default None calculates mean of all data items.
 
@@ -739,7 +740,7 @@ def mean(x, axis=None):
         x.comm.Allreduce(MPI.IN_PLACE, mu_tot, MPI.SUM)
 
         for i in range(1, x.comm.size):
-            mu_tot[0, :], n_tot[0] = merge_moments(
+            mu_tot[0, :], n_tot[0] = __merge_moments(
                 (mu_tot[0, :], n_tot[0]), (mu_tot[i, :], n_tot[i])
             )
         return mu_tot[0][0] if mu_tot[0].size == 1 else mu_tot[0]
@@ -764,7 +765,7 @@ def mean(x, axis=None):
             x.comm.Allreduce(mu_proc, mu_tot, MPI.SUM)
 
             for i in range(1, x.comm.size):
-                mu_tot[0, 0], mu_tot[0, 1] = merge_moments(
+                mu_tot[0, 0], mu_tot[0, 1] = __merge_moments(
                     (mu_tot[0, 0], mu_tot[0, 1]), (mu_tot[i, 0], mu_tot[i, 1])
                 )
             return mu_tot[0][0]
@@ -829,7 +830,7 @@ def mean(x, axis=None):
     )
 
 
-def merge_moments(m1, m2, bessel=True):
+def __merge_moments(m1, m2, bessel=True):
     """
     Merge two statistical moments. If the length of m1/m2 (must be equal) is == 3 then the second moment (variance)
     is merged. This function can be expanded to merge other moments according to Reference 1 as well.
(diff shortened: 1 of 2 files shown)


heat/core/statistics.py
```diff
@@ -807,99 +811,68 @@ def reduce_means_elementwise(output_shape_i):
...
+        a tuple of the merged moments
 
     References
     ----------
     [1] J. Bennett, R. Grout, P. Pebay, D. Roe, D. Thompson, Numerically stable, single-pass, parallel statistics
         algorithms, IEEE International Conference on Cluster Computing and Workshops, 2009, Oct 2009, New Orleans, LA,
         USA.
     """
+    n1, n2 = m1[-1], m2[-1]
+    mu1, mu2 = m1[-2], m2[-2]
```
mtar: Check the lengths of m1 and m2?

heat/core/statistics.py
```diff
@@ -1328,32 +1293,54 @@ def reduce_vars_elementwise(output_shape_i):
...
+                    (var_tot[i, 0], var_tot[i, 1], var_tot[i, 2]),
+                    bessel=bessel,
                 )
-                var_tot[0, 0] = merged[0]
-                var_tot[0, 1] = merged[1]
-                var_tot[0, 2] = merged[2]
-
             return var_tot[0][0]
 
     else:  # axis is given
```
mtar: Small inconsistency: When axis is set, integer tensors are not supported any longer. Numpy does support ints and floats, while pytorch only supports floats in general. This behaviour is also in 'mean'.

heat/core/statistics.py
```diff
@@ -1328,32 +1293,54 @@ def reduce_vars_elementwise(output_shape_i):
...
         # case for var in one dimension
         output_shape = list(x.shape)
-        if isinstance(axis, int):
```
mtar: Duplicates in axis list: Numpy checks for it, pytorch (and HeAT) doesn't. This is also in 'mean'.

heat/core/statistics.py
coquelin77: because torch does not support ints so we dont either. if its needed, the easiest way to do this would be to cast the input to a float. I dont think that we need to do this though. i dont think that we need to be so religious to numpy

heat/core/statistics.py
coquelin77: does it raise an exception or does it just ignore the duplicates?

mtar:
Would you be so kind to remove the failing test? As of today, torch.logical_xor allows non-bool tensors in the new pytorch version.

coquelin77:
> Would you be so kind to remove the failing test? As of today, torch.logical_xor allows non-bool tensors in the new pytorch version.

yes this is an issue that needs to be addressed. are you working on this? Bjoern wants to put a release out today. I am in a meeting at the moment, but once this is done I will get to work on a fix for the latest version of pytorch

coquelin77:
> Would you be so kind to remove the failing test? As of today, torch.logical_xor allows non-bool tensors in the new pytorch version.

should the fix be to use the bitwise xor?

mtar:
I will make a pull request

mtar:
One last thing, I was playing with var and realized that heat uses Bassel's correction as the default, similar to  [pytorch](https://pytorch.org/docs/stable/torch.html?highlight=var#torch.var). [numpy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.var.html),however, uses the biased one as default instead. Shall we keep it that way?

coquelin77:
> One last thing, I was playing with var and realized that heat uses Bassel's correction as the default, similar to [pytorch](https://pytorch.org/docs/stable/torch.html?highlight=var#torch.var). [numpy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.var.html),however, uses the biased one as default instead. Shall we keep it that way?

i think that Bessel's correction should be the default. But this is mostly personal preference. I think that for general statistical methods (as well as more advanced methods) it will provide more statistically reliable results

Issue #444, opened 2019-12-19
Define and test neutral_element function for reduction operations


ClaudiaComito:
**Related**
#369, #443 

**Feature functionality**
operations.__reduce_op() now accepts the kwarg `neutral` (a torch factory function) to replace the local tensor with a neutral tensor should the former contain no data.

At the moment, `neutral` is set to `torch.ones` for all reduction operations except ht.sum() (where `neutral=torch.zeros`).

Appropriate neutral local tensors need to be defined and tested for:

- statistics.argmax()
- statistics.argmin()
- statistics.max()
- statistics.min()
- logical.all()
- logical.any()

ClaudiaComito:
Closed with #446

Issue #369, opened 2019-09-02
reduce_op breaks on empty local tensor for some reduce operations


TheSlimvReal:
**Description**
Some of the torch reduce operations behave differently when passing in an empty tensor. For example `torch.sum` returns 0 but `torch.max` or `torch.min` throw an exception when an empty tensor is given as argument.

**To Reproduce**
Run with 4 processes:
```
>>> import heat as ht
>>> a = ht.ones((3, 3), split=1)
>>> a.max()
RuntimeError: invalid argument 1: cannot perform reduction function max on tensor with no elements because the operation does not have an identity at /pytorch/aten/src/TH/generic/THTensorEvenMoreMath.cpp:189
```

1. Which module/class/function is affected?
All methods using the 'reduce_op' function
2. What are the circumstances under which the bug appears?
Some of the internally used torch functions do not work on empty local tensors. 
3. What is the exact error-message/errorous behavious?
`RuntimeError: invalid argument 1: cannot perform reduction function max on tensor with no elements because the operation does not have an identity at /pytorch/aten/src/TH/generic/THTensorEvenMoreMath.cpp:189`

**Expected behavior**
The local functions should return a neutral value that does not affect the result (like the 0 for sum). The problem is to define this neutral argument for some other functions and I can not think of an general way to define this. Fact is that the mpi allreduce breaks when not every process provides values.

TheSlimvReal:
@ClaudiaComito Do you have any idea how an elegant solution might look like?

ClaudiaComito:
@TheSlimvReal thanks and sorry I missed this message. I'll look into this.

ClaudiaComito:
Hi again @TheSlimvReal. I could reproduce the error although not 100% (my system simply gets stuck, it doesn't throw an exception, possibly I just haven't waited long enough).

After researching a bit, I'm not sure that we should do anything at all, apart from maybe raising a warning when chunking and an exception when calling reduce_op. At that point we already know that some nodes have no data and will run into trouble.

I haven't been able to find a way to exclude nodes from MPI collective operations. I've been playing around with the option of using comm.Exscan instead of Allreduce. But that would imply:
- expanding the size of the input tensor in the split dimension to have non-empty partial tensors on every rank
- filling the fake added tensors with bogus data (factories.empty? ones?)
- (keeping track of which rank is the last one with actual data, let's call it rank k)
- running comm.Exscan instead of comm.Allreduce. 
[...]

Issue #101, opened 2019-01-17
operations: "make me more numpy API complete" 


ClaudiaComito:
check for "numpy API completeness", remove general TODO, create specific issues.

fschlimb:
Any chance for supporting `savez`/`savez_compressed` and loading NpzFiles?

ClaudiaComito:
this seems to have been addressed. Closing.

PR #446, opened 2020-01-08
Bug/369 reduce op empty tensor

ClaudiaComito:
## Description

Addresses  #369 and #444 

I'm introducing a kwarg `neutral` for operations.__reduce_op(). `neutral` is set to a scalar within the methods that call __reduce_op(). It allows __reduce_op() to define a neutral/identity local tensor in the (rare) occurrency in which a distributed reduction operation cannot proceed because one of the processes is left with no data. See #369 

## Type of change

Select relevant options.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

- [ ] yes
- [x] no

- [ ] yes
- [x] no

- [ ] yes
- [x] no

- [ ] yes
- [x] no


heat/core/logical.py
```diff
@@ -206,4 +208,6 @@ def any(x, axis=None, out=None, keepdim=False):
     def local_any(t, *args, **kwargs):
         return torch.any(t != 0, *args, **kwargs)
 
-    return operations.__reduce_op(x, local_any, MPI.LOR, axis=axis, out=out, keepdim=keepdim)
+    return operations.__reduce_op(
+        x, local_any, MPI.LOR, axis=axis, out=out, neutral=1, keepdim=keepdim
```
TheSlimvReal: For logical or I think the neutral value should also be 0?

heat/core/operations.py
```diff
@@ -385,18 +385,25 @@ def __reduce_op(x, partial_op, reduction_op, **kwargs):
...
+        neutral = kwargs.get("neutral")
+        if neutral is None:
+            neutral = float("nan")
+        neutral_shape = x.lshape[:split] + (1,) + x.lshape[split + 1 :]
+        partial = torch.full(neutral_shape, fill_value=neutral, dtype=x._DNDarray__array.dtype)
+    else:
+        partial = x._DNDarray__array
     if axis is None:
-        partial = partial_op(x._DNDarray__array).reshape(-1)
+        partial = partial_op(partial).reshape(-1)
```
TheSlimvReal: Wouldn't it be better to provide the neutral value only in the later call to `Allreduce`? I don't think it's necessary to run the neutral values through the `partial_op`.

heat/core/operations.py
```diff
@@ -385,18 +385,25 @@ def __reduce_op(x, partial_op, reduction_op, **kwargs):
...
     axis = stride_tricks.sanitize_axis(x.shape, kwargs.get("axis"))
-    split = x.split
+    if isinstance(axis, int):
+        axis = (axis,)
     keepdim = kwargs.get("keepdim")
+    split = x.split
 
+    # if local tensor is empty, replace it with the identity element
+    if 0 in x.lshape and (axis is None or (x.split in axis)):
+        neutral = kwargs.get("neutral")
```
TheSlimvReal: I would suggest to add `neutral` to the named arguments instead of passing it through the `kwargs`. This way we could also provide a description how this argument might be used.

heat/core/operations.py
```diff
@@ -385,18 +385,25 @@ def __reduce_op(x, partial_op, reduction_op, **kwargs):
...
-        partial = partial_op(x._DNDarray__array).reshape(-1)
+        partial = partial_op(partial).reshape(-1)
         output_shape = (1,)
     else:
-        if isinstance(axis, int):
-            axis = (axis,)
-
-        if isinstance(axis, tuple):
-            partial = x._DNDarray__array
+        if isinstance(axis, tuple):  # TODO: do we need this check at all?? axis has been sanitized
```
coquelin77: Checking if the axis is an int I think is needed to make the loop syntax in the for loop work correctly. I have a proposed solution which I will commit for you to look over

heat/core/operations.py
ClaudiaComito: In principle you're right, however the block of code 402-419 also gives partial (neutral or not) the right shape.

heat/core/logical.py
ClaudiaComito: Correct, thanks!

heat/core/operations.py
ClaudiaComito: I'll think about it a bit longer.

heat/core/operations.py
ClaudiaComito: My reasoning behind that decision was that neutral elements are specific to their functions and I'm not sure the user should be tempted to specify a different value. 

What I still haven't done is implement the `initial` argument (see Issue #101) which goes more in the direction of what you're saying, I think. It's on my list 😬

heat/core/operations.py
ClaudiaComito: @TheSlimvReal I gave it some thought. As I wrote yesterday, in the current implementation the tasks of running partial through partial_op() and assigning partial/neutral the appropriate shape are entangled. I could take them apart, write a nested function for the calculation of the appropriate shapes so that task is independent. It's feels like quite a bit of reshuffling though, and not my highest priority at the moment, so I suggest having a separate issue and dealing with it later. Let me know what you think.

coquelin77: @TheSlimvReal @ClaudiaComito i dont think that we need to get rid of it. i think that this is fine. if the node had no data on it to begin with we are not changing the amount of work done as the node is already allocated.

heat/core/operations.py
coquelin77: can you make an issue for this? (and tag it with the 'first issue' tag)

ClaudiaComito: @coquelin77 done, see #448.

Issue #472, opened 2020-02-04
Unsplit matrix multiplication changes split of first matrix argument


lehr-fa:
**Description**

The split of the first argument (X in the following example) gets changed from None to 0 during matrix multiplation. Example:

**To Reproduce**

```python
X = ht.zeros((200, 9), split=None)
A = ht.zeros((9, 10), split=None)
y = ht.matmul(X,A)
print(X.split)
```

```
 >> 0
```

1. Which module/class/function is affected: ```ht.linalg.matmul```
2. What are the circumstances under which the bug appears: Allways
3. What is the exact error-message/errorous behavious: The split changes from None to 0.

**Expected behavior**
 The split axis of the input args should be unchanged. In this case, X.split should be None

**Version Info**
HeAT Master

lehr-fa:
@Cdebus In line https://github.com/helmholtz-analytics/heat/blob/f3eca8c518d26cdc7ae2ee88e2a785fcece959db/heat/core/linalg.py#L154

the tensor is resplit. Why is this necessary here? Both tensors are unsplit und could be simple multiplied. IMHO, no MPI is necessary at all for this particular case.

Another issue here could be, that the resplit really changes the input and not only a temporary variable.

coquelin77:
Firstly, thank you for the issue! the rational here was to divide the computational load between the processes. 

The split is done assuming that the data size is large enough to see a benefit from parallelization. It does not do this on a temporary variable as that would require copying the memory which can be computationally expensive. However, this does raise a good point and an optional argument could be added to `matmul` to make this behavior optional. 

If you are experiencing any other problems please let us know!

rainman110:
IMHO, resplitting should only be done for matrices that are "large enough", as a resplitting introduces additional overhead. Therefore, it is probably a good idea to check for a certain minimum size before resplitting.

PR #473, opened 2020-02-05
Bug/472 matmul updates

coquelin77:
## Description

Updates to the matmul and dot functions to increase their 

Issue/s resolved: #472 

## Changes proposed:
- flag added to matmul to avoid resplit of the first argument (`a`)
- dot not returns a split=None vector of both vectors are split=None

## Type of change

Remove irrelevant options:
- Documentation update

## Due Diligence

It should not change the behavior of other functions but if they use dot and expect a split axis this may need to be addressed.


heat/core/linalg.py
```diff
@@ -68,7 +77,7 @@ def dot(a, b, out=None):
         raise NotImplementedError("ht.dot not implemented for N-D dot M-D arrays")
 
 
-def matmul(a, b):
+def matmul(a, b, no_resplit=False):
```
Cdebus: default value for resplitting input tensors should be False i believe, meaning that no_resplit=True

heat/core/linalg.py
```diff
@@ -145,7 +163,7 @@ def matmul(a, b):
         b = c_type(b, device=b.device)
 
     if a.split is None and b.split is None:  # matmul from torch
-        if len(a.gshape) < 2 or len(b.gshape) < 2:
+        if len(a.gshape) < 2 or len(b.gshape) < 2 or no_resplit:
             # if either of A or B is a vector
             return factories.array(
                 torch.matmul(a._DNDarray__array, b._DNDarray__array), device=a.device
```
Cdebus: two lines below (ll 172) a = a.resplit_(0) is not the appropriate syntax: resplit_ is inplace, but this gives the impression that it is out of place. 
Better would be a.resplit_(0)

heat/core/linalg.py
coquelin77: I think that it should be True. if we are targeting the usage on large datasets i think that we should try to get as much speed as possible. The downside being that it makes the learning curve steeper for users. @Markus-Goetz what do you think?

coquelin77:
> I'm not quite sure if it is sufficient to only consider the vector case for this.
> When I implemented the resplit function for out-of-place case, I adapted the HeAT functions, however did not change the current implemenation design (aka most functions kept the inplace resplit) since I didn't know enough on the algorithm design (especially in matmul).
> However, now might be a good design to discuss the overall behaviour and API, and wheter matmul should change the splitting of the inputs inplace.

it should only resplit in the case that is mentioned. but this could be done out of place as well, although it would require a copy call.

heat/core/linalg.py
Cdebus: @Markus-Goetz  Don't worry about it, @coquelin77 and I discussed this. We came to the conclusion that the best way would be to make the storage-wise more efficient version the default (aka no_resplit=False), offer the option to make the less efficient, but more expected and convenient method (not resplitting the inputs) optional via the flag and update the Docs to make it clear to the user.... so that they are warned
@rainman110 , @lehr-fa FYI

rainman110:
@coquelin77 I have some notes to your fix
 - It feels still wierd in a functional interface, when input args are changed after the call. IMHO ```a``` should be copied to a temporary before resplit. I know, that this could be a performance issue though
 - the new argument ```no_resplit=False``` feels a bit strange from the api design as it uses a double negative. Why not turning it around, i.e. ```resplit=True``` by default?

coquelin77:
@rainman110 @Cdebus I changed the default behavior to be the case that the split axis is not changed. I think that this solution is the most clear for users, those with more experience can choose if they want to allow for 'a' to be split or not

rainman110:
@coquelin77  Sounds good to me! Thank you very much!

coquelin77:
codecov is apparently reading a coverage change in kmeans but i believe that this is due to it reading the wrong base commit. there are no changes made to kmeans in this PR according to github

Cdebus:
Nice work!

Issue #425, opened 2019-12-06
Resplit from 1 -> 0 gives wrong order


coquelin77:
**To Reproduce**
1. Which module/class/function is affected? -> resplit and the functions which use this (allclose, ...)
2. What are the circumstances under which the bug appears? -> going from split=1 to split=0
3. What is the exact error-message/errorous behavious?

```python
st = torch.randn(9, 6)
a = ht.array(st, split=1)
a_comp = ht.array(st, split=0)
print(a_comp == a.resplit_(0)) ```

```
[0] tensor([[ True, False, False,  True, False, False],
[0]         [False, False, False, False, False, False],
[0]         [False,  True, False, False,  True, False],
[0]         [False, False, False, False, False, False],
[0]         [False, False,  True, False, False,  True]])
[1] tensor([[ True, False, False,  True, False, False],
[1]         [False, False, False, False, False, False],
[1]         [False, False, False, False, False, False],
[1]         [False, False,  True, False, False,  True]])
```

**Expected behavior**
A clear and concise description of what you expected to happen.
In example all of the values should be true

**Version Info**
 Which version are you using?
current master as of 6.12.19

ClaudiaComito:
Hi, I ran into the same problem while working on GNB (#340, #474). The problem arises because Alltoallv transposes the data on top of shuffling them around. See illustration below for a tensor:
`
X = ht.array(torch.arange(7 * 4).reshape(7, 4), split=1)` that we want to resplit to `axis=0`.

I've been playing around with an intermediate step - transposing the data first, then feeding them to Alltoallv - but I haven't gotten it to work yet and won't have time for it in the next few days.

The brute-force fix (gather first, then redistribute) takes a minute, yields the right results but is bound to blow up at some point.  


![Resplit1-0 001](https://user-images.githubusercontent.com/39374113/74008061-3ad68400-4980-11ea-80c9-e4f5de9e79b1.jpeg)

Cdebus:
@coquelin77 should we really close this issue once the tiling solution is implemented? Maybe then rename it to an all-to-all bug and add a comment that "resplit" might need rework in the future, once alltoall is fixed (that would require benchmarking to see which option performce better)

coquelin77:
so the issue itself is fixed by the pr. however, that PR works around the cause of the issue. the cause of this issue could itself be a different issue, since the code that causes it would be removed.

coquelin77:
actually, we could rename this one and add the code required to raise it here

PR #476, opened 2020-02-07
Brute-force (temporary) fix for Issue #425. When redistributing, gather first.

ClaudiaComito:
## Description

Redistributing from one split axis to another currently yields the right data but at the wrong indices, see #425. This is a temporary, brute-force fix, which gathers the tensor first and then redistributes it. I'll leave it to you guys and gals to decide whether we want to merge it for now and find a better solution later, or leave the bug in until we have the Alltoallv (or else) solution.
  

Issue/s resolved: #425 

## Changes proposed:
- dndarray.resplit_(x, axis). IF x.split is not None and axis is not None and x.split != axis, gather x first, then split it along axis.  

## Type of change

Remove irrelevant options:

## Due Diligence

no


ClaudiaComito:
> Do we have a test case for this now that fails if the resplit yields wrong results when running distributed? If not, then this should also be added.

We have a test case in GaussianNB, that's how I noticed :) But you're right @bhagemeier, I'll expand test_resplit.

PR #483, opened 2020-02-11
bugfixes dndarray initialization and .cpu()

mtar:
## Description
Bugfix: Wrong device of a heat array on initialization in some cases when underlying torch tensor is on different device.
Bugfix: .cpu() method does not change the heat device

## Changes proposed:
- move torch tensor to the new device on initialization
- change device to ht.cpu in .cpu()

## Type of change
Remove irrelevant options:

no


coquelin77:
does this need to be done in the `ht.array` as well?

mtar:
ht.array returns a dndarray. Playing with ht.array(torch.tensor(...)), I get the right devices now.

Issue #485, opened 2020-02-13
cdist tests failing on multiple GPUs


coquelin77:
**Description**
cdist unit tests are failing on multiple GPUs

**To Reproduce**
1. Which module/class/function is affected?  - `cdist`
2. What are the circumstances under which the bug appears? - multiple GPUs on multiple nodes
3. What is the exact error-message/errorous behavious?

```
        if compute_mode == 'use_mm_for_euclid_dist_if_necessary':
>           return torch._C._VariableFunctions.cdist(x1, x2, p, None)
E           RuntimeError: X1 and X2 must have the same device type. X1: cuda X2: cpu

../heat-venv/lib/python3.6/site-packages/torch/functional.py:671: RuntimeError
```

**Expected behavior**
Test cases should run on GPUs the same as on CPUs

**Version Info**
Current master (13.02.2020)

Issue #468, opened 2020-01-28
Implement sanitation module


ClaudiaComito:
**Feature functionality**
Collect sanitation checks in a dedicated module. This would include the omnipresent 

```
if not isinstance(x, dndarray.DNDarray):
        raise TypeError("expected x to be a ht.DNDarray, but was {}".format(type(x)))
```
and variations thereof, but also more specific checks needed in high-level algorithms. See sklearn's validation module:

https://github.com/scikit-learn/scikit-learn/blob/b194674c4/sklearn/utils/validation.py

Cdebus:
This should probably also handle communication and device sanitation. That is something that pops up frequently, but is a bit tedious to check for in each function individually. So if you make a general sanitation function, could you include this?

ClaudiaComito:
Closed by #662

PR #486, opened 2020-02-13
cdist tests on GPU

coquelin77:
## Description

Added missing device calls in cdist tests

Issue/s resolved: #485 

## Changes proposed:
- added device parameters

## Type of change

Remove irrelevant options:

## Due Diligence


mtar:
The test errors on my machine. There are still some device calls missing.

Cdebus:
Cdist still needs input sanitation with respect to the device (so inputs need to have the same results). As of now, we rely on pytorch to handle the error, but this would be something that could be covered with Issue #468

Issue #453, opened 2020-01-17
Submodulerize linalg file after QR is added


coquelin77:
**Related**
With QR the linalg file has gotten to be rather large and cumbersome. to make this a submodule would make it easier to read and then future enhancements can be made rather easily.

This should be done after QR is merged (#429)

PR #495, opened 2020-03-06
Features/453 submodule linalg

coquelin77:
## Description

As the linalg file is rapidly expanding, this PR splits it into its a sub-module of the heat core. There will be more things to come in the linear algebra package as well, this change is mostly cosmetic. 

The only algorithmic changes take place in the abstraction of QR. With the changes proposed here the qr functions are reduced to be functions to be called on each of the tile columns created. This change will make it less technical for anyone to call the internal QR functions to reduce a tile column into QR components. This change is motivated by interweaving QR and LQ decompositions

Issue/s resolved: #453 

## Changes proposed:
- create submodule for the linear algebra package
- further abstract the QR functions

## Type of change

Remove irrelevant options:
- Documentation update (minor)

## Due Diligence

no


coquelin77:
To reach the required codecov/patch 78 more lines need to be covered. the distribution of missed lines in the linalg submodule is as follows:

- 42  in tests (extended tests for qr and GPU test option)
- 17 in jitted function in matmul
- 12 in corner cases in tiling
- 6 in corner cases in matmul
- 3 in corner cases in QR

total: 80. if these lines are included in the calculation, both codecov requirements pass

Issue #481, opened 2020-02-09
introduce Halos


krajsek:
This feature is a preparation to issue #79 "conv1d, conv2d, conv3d and a generic conv function"

krajsek:
Expands the array attribute withoptional left and right extensions. Introduces an access pattern to the original array, the halos as well as the whole new array = old array + halos

PR #500, opened 2020-03-11
Features/halorize

lucaspataro:
## Description

Introducing of halos before and after an DNDarray along the split axis in 1D and 2D.

Issue/s resolved: #481 

## Changes proposed:
-
-
-
-

## Type of change

## Due Diligence

yes
- factories.array


heat/core/dndarray.py
```diff
@@ -62,6 +64,9 @@ def __init__(self, array, gshape, dtype, split, device, comm):
         ):
             self.__array = self.__array.to(devices.sanitize_device(self.__device).torch_device)
 
+        if self.__halo > 0:
```
Markus-Goetz: Why would axis zero not have halos? Is '0' used as placeholder for no split? If so, 'None' seems more meaningful as default.

heat/core/dndarray.py
```diff
@@ -45,13 +45,15 @@ def __setitem__(self, key, value):
 
 
 class DNDarray:
-    def __init__(self, array, gshape, dtype, split, device, comm):
+    def __init__(self, array, gshape, dtype, split, device, comm, halo=0):
```
Markus-Goetz: Would claim that their should be no explicit default here for halo. We need to make sure that the rest of the code is passing correct halo informations down internally.

heat/core/dndarray.py
```diff
@@ -74,6 +79,74 @@ def device(self):
     def dtype(self):
         return self.__dtype
 
+    @property
+    def halo(self):
+        """
+        Returns
+        -------
+        int : number of rows or columns used as halo
```
Markus-Goetz: How would this work with arbitrary dimensions? In my opinion it would be good to pass an array with length equal to the number of dimensions everywhere that indicate the halo sizes in each dimensional direction

heat/core/dndarray.py
```diff
@@ -74,6 +79,74 @@ def device(self):
...
+
+        halo = self.sanitize_halo(halo)
+        has_halo = self.is_halorized()
+
+        # if tensor is already halorized ...
+        if has_halo:
+            # ... check if new halo differs from current one
+            if halo == self.__halo:
+                # new halo size equals current halo, so no need to do anything
+                warnings.warn(
```
Markus-Goetz: This probably should not issue a warning. Some code might recall this function even if halorized and should end up to be a no-op.

heat/core/dndarray.py
```diff
@@ -74,6 +79,74 @@ def device(self):
...
+                    "provided halo is equal to current halo, no further actions taken",
+                    RuntimeWarning,
+                )
+                return
+            else:
+                # new halo size is different from current one, so delete current halos
+                # TODO: is there a special way to delete torch tensor?
```
Markus-Goetz: `
del tensor
`

heat/core/dndarray.py
```diff
@@ -1464,6 +1537,59 @@ def __gt__(self, other):
...
+            cat.append(self.__halos[0])
+
+        cat.append(self.__array)
+
+        if self.__halos[1] is not None:
+            cat.append(self.__halos[1])
+
+        out = torch.cat(cat, self.split)
+
+        return DNDarray(out, out.shape, self.dtype, None, self.device, None, None)
```
Markus-Goetz: This is a dangerous thing to do. The new array content would be now the actual array plus the catted halos on the side. Some low-level code will directly access `__array` and will not get all data including halos. I believe it would be good to introduce a new member that is called something along the lines of `__raw_data`, `__storage`,` __array_with_halos` and make `__array` to be a property that will return the actual array content.

heat/core/dndarray.py
lucaspataro: self.__halo is the with of the halo not the axis. The axis is always the same as self.split

heat/core/dndarray.py
lucaspataro: As i understood it, it is only possible to split an tensor along one axis right? If so then there is only the need for one halo along that axis?

heat/core/dndarray.py
lucaspataro: Good point, i will remove the warning.

heat/core/dndarray.py
lucaspataro: Okay thanks, i wasn't sure this was sufficient.

rainman110:
@krajsek this is no commitment from outside. Luca was assigned by me to implement convolutions. The creation of halos is required for this task.

This work was discussed with @Markus-Goetz . We agreed, that we need this now if we finally want to support CNNs. As we have not seen any work related to it on the public repo, we now made the step forward.

Please do not attack @lucaspataro . He did a good job. Please be more constructive in your criticism.

krajsek:
> 
> 
> @krajsek this is no commitment from outside. Luca was assigned by me to implement convolutions. The creation of halos is required for this task.
> 
> This work was discussed with @Markus-Goetz . We agreed, that we need this now if we finally want to support CNNs. As we have not seen any work related to it on the public repo, we now made the step forward.
> 
> Please do not attack @lucaspataro . He did a good job. Please be more constructive in your criticism.

@rainman110 There is no need to be more constructive from my side. My statements are clear and constructive: If someone wants to take over a task that has already been assigned to someone else, he or she should contact that person in order to coordinate the work. I have already spend some time on halo/convolution and my work already covers what has been done by @lucaspataro. However, I withdrew  my pull request as my approach was not computational efficient enough and I wanted to improve it. Coordination of our efforts can prevent double work and my request to contact people before taking over their tasks are therefore constructive.

heat/core/dndarray.py
Markus-Goetz: You are absolutely right

Markus-Goetz:
@rainman110 thought we only discussed that having a linear layer would be good. Nevertheless, would be good if @lucaspataro and @krajsek coordinated on this

rainman110:
@krajsek Okay... I admit there could have been a better communication. So lets be more result-oriented: Can you please discuss with @lucaspataro , what is missing from your perspective and give him everything that you already implemented.

In future, we should avoid having some almost-finished code not available for the other devs. Looking at the public branches, I could not find anything related to convolutions or halos that was not  implemented by @lucaspataro .

I now searched again the PRs and found this particular one: https://github.com/helmholtz-analytics/heat/pull/77

Is this something @lucaspataro could use? Why did you close this PR in the end?

lucaspataro:
@krajsek and I are now in contact and will discuss how to proceed with the halorization and convolution topics.

heat/core/dndarray.py
krajsek: Checking for halo size is not enough to determine if a halo is up to date. The remote local array that contains the halo information might be updated independently.

coquelin77:
is this taken care of by #541 ?

lucaspataro:
Yes #541 solves this pull request fully.

PR #501, opened 2020-03-12
Features/176 flatten

mtar:
## Description

implementation of flatten(). It returns a new flattened tensor.

Issue/s resolved: #176 

## Changes proposed:
- add flatten function

## Type of change

Remove irrelevant options:

## Due Diligence

no


mtar:
> this does flatten the array but i think that it might also destroy the global ordering. can you look into this?

What do you mean?

coquelin77:
> > this does flatten the array but i think that it might also destroy the global ordering. can you look into this?
> 
> What do you mean?

```python
x = torch.arange(18).reshape(3, 6)
x1 = ht.array(x, split=1)
print(x1)
print(ht.flatten(x1))
```

```
[0] tensor([[ 0,  1,  2],
[0]         [ 6,  7,  8],
[0]         [12, 13, 14]])
[1] tensor([[ 3,  4,  5],
[1]         [ 9, 10, 11],
[1]         [15, 16, 17]])

[0] tensor([ 0,  2,  7,  3,  5, 10,  1,  6,  8])
[1] tensor([ 4,  9, 11, 12, 13, 14, 15, 16, 17])
```

also there should be an inline function for flatten so one can call `x.flatten()`

mtar:
It's a problem of the ht.resplit function then, isn't it?

```python
x = torch.arange(18).reshape(3, 6)
x1 = ht.array(x, split=1)
x0 = ht.resplit(x1, 0)
print(ht.MPI_WORLD.rank, x0)
```
```console
0 tensor([[ 0,  2,  7,  3,  5, 10],
        [ 1,  6,  8,  4,  9, 11]])
1 tensor([[12, 13, 14, 15, 16, 17]])
```

coquelin77:
#425 is already created. the resplit problem is known and it is being worked on but it is more complex then it first appears. 

that being said, your algorithm works for split=0. i do not have an eta on when the resplit function solution will be up. in the meantime a raise/warning can be implemented and then leave a comment on the mentioned error. that way when the issue is solved then the raise/warning will be removed

coquelin77:
> #425 is already created. the resplit problem is known and it is being worked on but it is more complex then it first appears.
> 
> that being said, your algorithm works for split=0. i do not have an eta on when the resplit function solution will be up. in the meantime a raise/warning can be implemented and then leave a comment on the mentioned error. that way when the issue is solved then the raise/warning will be removed

replit fix is in the works. im going to say that we pause this PR until that is done. I have a working resplit algo (not touching split=None yet though) that i am working on finishing up. hopefully done in the coming days

heat/core/manipulations.py
```diff
@@ -558,6 +560,49 @@ def expand_dims(a, axis):
...
+            dtype=a.dtype,
+            is_split=None,
+            device=a.device,
+            comm=a.comm,
+        )
+
+    # The resplit function scramble the tensor when switching axes, see issue 425
+    if a.split > 0:
+        a = resplit(a, 0)
+        warnings.warn("The flattened tensor may have a wrong order for split axes > 0", UserWarning)
```
Markus-Goetz, line 593: Warning should be removed, after resplit is working.

heat/core/manipulations.py
```diff
@@ -558,6 +560,49 @@ def expand_dims(a, axis):
...
+    a = factories.array(
+        torch.flatten(a._DNDarray__array),
+        dtype=a.dtype,
+        is_split=a.split,
+        device=a.device,
+        comm=a.comm,
+    )
+    a.balance_()
```
Markus-Goetz, line 601: replit() + balance_() is a very expensive way of doing flattening. It entails superfluous communication that is then undone again by the next communication. Can we get this working in one go?

PR #506, opened 2020-03-15
Fix version in setup.py

melven:
## Description

The old code adjusted sys.path and imported heat/core/version which
implicitly imports heat itself. This leads to two problems:
 * setup cannot run without dependencies,
   e.g. just "python setup.py --version" fails when py-torch is not available
 * setuptools skips copying the complete heat code into the target
   directory in some cases as heat is already in sys.path, see https://setuptools.readthedocs.io/en/latest/setuptools.html:
    >--always-copy, -a
    >  Copy all needed distributions to the staging area, even if they are already present in another directory on sys.path. By default, if a requirement can be met using a distribution that is already available in a directory on sys.path, it will not be copied to the staging area.
  

The new variant directly executes the version.py file and avoids the problems above.

Issue/s resolved: #

## Changes proposed:

Execute version.py in setup.py directly instead of importing heat.

## Type of change

Remove irrelevant options:

## Due Diligence

no


coquelin77:
this is failing in pre-commit. please see the README for how to set this up

coquelin77:
this is still failing. is there any update on this?

Markus-Goetz:
Bump

coquelin77:
the line that you removed markus was there to satisfy flake. `__version__` will be initialized in the exec but flake doesnt know that so it fails

melven:
Hi

Sorry, I didn't answer earlier. I'm not working most of the time, currently.
(I just took holidays to play with the kids at home)

So feel free to fix/improve this (thanks!).

I can also look at it after Easter.

Issue #492, opened 2020-02-21
bug distributed array from scalar


mtar:
**Description**
factories.array does not support 0-dim tensors if split is not None.

**To Reproduce**
1. Which module/class/function is affected?
factories.array / stride_tricks.sanitize_axis
2. What are the circumstances under which the bug appears?
create splitted heat array from scalar
3. What is the exact error-message/erroneous behaviour?
```bash
Traceback (most recent call last):
  File "pt.py", line 4, in <module>
    a = ht.array(1, split=0)
  File "heat/core/factories.py", line 323, in array
    split = sanitize_axis(obj.shape, split)
  File "heat/core/stride_tricks.py", line 109, in sanitize_axis
    raise ValueError("axis {} is out of bounds for shape {}".format(axis, shape))
ValueError: axis 0 is out of bounds for shape torch.Size([])
Traceback (most recent call last):
  File "pt.py", line 4, in <module>
    a = ht.array(1, split=0)
  File "heat/core/factories.py", line 323, in array
    split = sanitize_axis(obj.shape, split)
  File "heat/core/stride_tricks.py", line 109, in sanitize_axis
    raise ValueError("axis {} is out of bounds for shape {}".format(axis, shape))
ValueError: axis 0 is out of bounds for shape torch.Size([])
```

**Expected behaviour**
Create 0-dim heat tensor

**Illustrative**
```
a = ht.array(1, split=0)
```
```
b = ht.array(torch.tensor(0), split=0)
```
**Version Info**
v0.3.0

**Additional comments**
This bug affects issue #490

ClaudiaComito:
@mtar do you think this is causing #490? I don't see why the tests wouldn't fail with 2 processes already if this were the problem.

(why distribute a scalar would be my next question)

mtar:
The bug happens more specifically in indexing.nonzero() lines 74-75 when lcl_nonzero has only one element:
```python
if a.numdims == 1:
    lcl_nonzero = lcl_nonzero.squeeze()
return factories.array(lcl_nonzero, is_split=is_split, device=a.device, comm=a.comm)
```
In this particular case, squeeze will return a 0-dim tensor and stuff happens.

On seven processes, you run into it (line 346).

PR #507, opened 2020-03-16
bug distributed scalar

mtar:
## Description

This PR fixes the error creating a scalar with split axis != None. The axis will be set to None implicitly. 

Issue/s resolved: #492 

## Changes proposed:
- In sanitize axis change axis to None if len(shape) == 0

## Type of change

Remove irrelevant options:

## Due Diligence

yes / no


heat/core/stride_tricks.py
```diff
@@ -90,6 +90,10 @@ def sanitize_axis(shape, axis):
     >>> sanitize_axis((5, 4), 1.0)
     TypeError
     """
+    # scalars are handled like unsplit matrices
+    if len(shape) == 0:
```
coquelin77: can you change this to use `numel()`? this would allow for the case that there is a 1 element torch tensor in a DNDarray

mtar: So the split axis of 1x1 matrices shall also be None?

mtar: So the split axis change shall be expended to higher dimensional tensors like this `np.prod(shape) == 1`? This change will let some unit tests fail.

Markus-Goetz: Scalars should be zero-dimensional

CHANGELOG.md
```diff
@@ -3,8 +3,9 @@
 - [#429](https://github.com/helmholtz-analytics/heat/pull/429) Implementated QR
 - [#429](https://github.com/helmholtz-analytics/heat/pull/429) Added PyTorch Jitter to inner function of matmul for increased speed
 - [#483](https://github.com/helmholtz-analytics/heat/pull/483) Bugfix: Underlying torch tensor moves to the right device on array initialisation
-- [#483](https://github.com/helmholtz-analytics/heat/pull/483) Bugfix:DNDarray.cpu() changes heat device to cpu
+- [#483](https://github.com/helmholtz-analytics/heat/pull/483) Bugfix: DNDarray.cpu() changes heat device to cpu
 - Update documentation theme to "Read the Docs"
+- [#507](https://github.com/helmholtz-analytics/heat/pull/507) Bugfix: sanitize_axis changes axis of 0-dim scalars to None
```
Markus-Goetz, line 8: Changelog conflic

Issue #490, opened 2020-02-19
test_gaussiannb fails when distributed on 7 processes


ClaudiaComito:
**Description**
Testing gets stuck, possibly nodes being left without data (iris test dataset is quite small).

**To Reproduce**

on branch features/340-GaussianNB
mpirun -n 7 python -m unittest

**Expected behavior**
Tests should not get stuck even if nodes are left without data

mtar:
Which branch do you use?

ClaudiaComito:
> Which branch do you use?

Sorry, branch is features/340-GaussianNB, I'll add this info in the general description.

PR #508, opened 2020-03-18
Bug/490 gnb 7 node testing

ClaudiaComito:
## Description

This PR fixes assorted bugs that led test_gaussiannb to fail on 7 nodes. The problems were related to how heat functions/methods deal with locally empty tensors and, in one case, 1-element tensors.

Issue/s resolved: #490 

## Changes proposed:

- ht.nonzero() now returns dndarray.DNDarray() instead of factories.array(), as the latter gets stuck when one of the nodes has no data (see #492 - which is NOT solved)
- dndarray.__getitem__ in distributed mode now returns the correct result both with 0-dim input (a scalar), and with an empty tensor as input.
- ht.var() ~of a single element tensor no longer returns nan, like torch, but 0.0 like numpy~ and ht.std() now behave like numpy.var() and numpy.std(), as in: whether the Bessel correction is applied or not, is defined by kwarg `ddof`. **N.B. default is now `ddof=0`, corresponding to setting `unbiased=False` in torch.var()**  Thanks @mtar for pointing me to the right direction.

## Type of change

Remove irrelevant options:
- Bug fix WITH breaking change: ht.var(), ht.std()

## Due Diligence

no


ClaudiaComito:
> Please check your latest commit. It removes older, already merged pull requests.

It should remove the latest merge master. We shouldn't merge master onto bugfix branches, I did by mistake and took it back.

mtar:
okay, I said nothing :zipper_mouth_face:

ClaudiaComito:
> okay, I said nothing 🤐

:) no worries, it is a bit confusing. Thanks for looking into it by the way

mtar:
It's not the only thing that confuses me :confused:. The file comparison is also a mess :dizzy:  :dizzy_face:

heat/core/statistics.py
```diff
@@ -1222,7 +1222,10 @@ def reduce_vars_elementwise(output_shape_i):
 
         if x.lshape[x.split] != 0:
             mu = torch.mean(x._DNDarray__array, dim=axis)
-            var = torch.var(x._DNDarray__array, dim=axis, unbiased=bessel)
+            if x.lshape[x.split] == 1:
+                var = factories.zeros(output_shape_i, dtype=x.dtype, device=x.device)
+            else:
+                var = torch.var(x._DNDarray__array, dim=axis, unbiased=bessel)
```
mtar: Have you considered that np.var(ddof=1) <=> and np.var <=> torch.tensor(unbiased=False) respectively? In other words, the numpy uses a biased estimator and pytorch an unbiased one by default.
On a single value
- Pytorch's var returns 0 when it's biased.
- numpy's var returns nan when its biased.

coquelin77:
> > Please check your latest commit. It removes older, already merged pull requests.
> 
> It should remove the latest merge master. We shouldn't merge master onto bugfix branches, I did by mistake and took it back.

@mtar was right, by reverting the commit you would delete all of the changes which are upstream of you. you can see this in the line difference. that is why you have about 3000 lines deleted.

ClaudiaComito:
> > > Please check your latest commit. It removes older, already merged pull requests.
> > 
> > 
> > It should remove the latest merge master. We shouldn't merge master onto bugfix branches, I did by mistake and took it back.
> 
> @mtar was right, by reverting the commit you would delete all of the changes which are upstream of you. you can see this in the line difference. that is why you have about 3000 lines deleted.

Guys, you're absolutely right, by reverting the merge I remove the merge. I'm afraid that's exactly what I wanted.

I'm thinking I should make two different PRs of this, since fixing ht.var() and ht.std() has turned into a breaking change.



> > > Please check your latest commit. It removes older, already merged pull requests.
> > 
> > 
> > It should remove the latest merge master. We shouldn't merge master onto bugfix branches, I did by mistake and took it back.
> 
> @mtar was right, by reverting the commit you would delete all of the changes which are upstream of you. you can see this in the line difference. that is why you have about 3000 lines deleted.

heat/core/statistics.py
ClaudiaComito: You're absolutely correct. Thanks a lot. ht.var() and ht.std() are numpy-compliant now

ClaudiaComito:
Closing this PR and replacing it with slightly less messy #519

Issue #143, opened 2019-02-25
Implement reshape()


Markus-Goetz:
Contains a simple and a complicated sub-issues. If the reshaped tensor only receives new dimensions, aka expand_dims, the implementation is trivial and can be performed locally. In case of actual reshaping, depending on the split axis, redistribution of the data has to be performed.

coquelin77:
as decided, we will use the hard transmit option

PR #511, opened 2020-03-27
Feature reshape function

mtar:
## Description

This PR proposes a rudimentary reshape functionality on 1-dim flat tensors returned by some functions (i.e, ht.arange, ht.linspace, and ht.logspace). Additionally, reshaping of all non-distributed tensors is handled by the pytorch function.
~Functionality for other kind of distributed tensors is not implemented~
**EDIT**: It should now work for all tensors with split == 0 now
**EDIT**: All split axes are supported

Issue/s resolved: #143

## Changes proposed:
- local torch.reshape on non-distributed tensors
- reshape on vectors or flat tensors

## Type of change

Remove irrelevant options:
- Documentation update

## Due Diligence

no


mtar:
Support for all tensors with split == 0 added

coquelin77:
have you looked into how to get this to work for `split=1`?

heat/core/manipulations.py
```diff
@@ -699,6 +700,113 @@ def hstack(tup):
...
+        return factories.array(
+            torch.reshape(a._DNDarray__array, shape), dtype=a.dtype, device=a.device, comm=a.comm
+        )
+
+    # Check for currently not supported tensors
+    if a.split > 0:
+        raise NotImplementedError("Tensors with split axes > 0 are not supported yet")
+
+    # Create new flat result tensor
+    _, local_shape, _ = a.comm.chunk(shape, a.split)
```
coquelin77, line 785: can use `a.lshape` instead

heat/core/manipulations.py
```diff
@@ -699,6 +700,113 @@ def hstack(tup):
...
+    data = torch.empty(np.prod(local_shape), dtype=a.dtype.torch_type())
```
coquelin77, line 786: device call, and numpy doesnt run on GPU

heat/core/tests/test_dndarray.py
```diff
@@ -402,6 +402,23 @@ def test_redistribute(self):
             with self.assertRaises(ValueError):
                 st.redistribute_(target_map=torch.zeros((2, 4)))
 
+    def test_reshape(self):
```
coquelin77, line 405: these tests can be deleted. you can do the inline call in the manipulations tests. no need to replicate the tests

heat/core/tests/test_manipulations.py
```diff
@@ -874,6 +874,49 @@ def test_hstack(self):
...
+        self.assertTrue(ht.equal(reshaped, result))
+
+        # 1-dim distributed vector
+        a = ht.arange(8, dtype=ht.float32, split=0)
+        result = ht.array([[[0, 1], [2, 3]], [[4, 5], [6, 7]]], dtype=ht.float32, split=0)
+        reshaped = ht.reshape(a, (2, 2, 2))
+
+        self.assertEqual(reshaped.size, result.size)
+        self.assertEqual(reshaped.shape, result.shape)
+        self.assertTrue(ht.equal(reshaped, result))
```
coquelin77, line 894: test for dtype

coquelin77: also we use floats as the default so this doesnt test anything different. please change to float64

heat/core/manipulations.py
mtar: I need the new local shape here.

coquelin77: sorry i missed that. this is correct

heat/core/manipulations.py
mtar: Thanks for the hint. I hadn't the opportunity to test it on the GPU.
I don't really get the second remark. np.prod will return an integer?

mtar:
> have you looked into how to get this to work for `split=1`?

I couldn't come up with a nice solution I'm happy with. So I thought it's better to have some reshape functionality than none  :grinning:

heat/core/manipulations.py
mtar: It runs on GPU. DNDarray handles wrong devices, but, nevertheless, it is better to change it beforehand.

heat/core/tests/test_dndarray.py
mtar: Good to know. Thank you.

heat/core/manipulations.py
coquelin77: the numpy command will run on the CPU process even if the device is GPU, which is why we need to call the torch version. it isnt going to save a lot of time but out programming model is to put it all on one device.

so it would be:

```python
_, local_shape, _ = a.comm.chunk(shape, a.split)
lshp = torch.prod(torch.tensor(local_shape, device=a.device.torch_device), dim=0)
data = torch.empty(local_shape, dtype=a.dtype.torch_type(), device=a.device.torch_device)
```
im not sure if torch requires the dim argument for prod

mtar: If you say so, I will change it :confused: 
Does it really save time? It seems more like an overhead for me creating a tensor and pushing data to the gpu and back for the calculation.

coquelin77: with the use of torch it does not leave the gpu. when you use numpy and the device is gpu then it has to leave the gpu, go to the cpu, then go back to the gpu after it's done.

mtar:
All split axes are supported now.

heat/core/manipulations.py
```diff
@@ -735,6 +736,170 @@ def hstack(tup):
...
+    """
+
+    def reshape_counts_displs(csum1, csum2):
+        """
+        Calculate the counts and displacements needed for redistributing the data from two cumulative sums.
+        """
+        p = csum1[a.comm.rank - 1] if a.comm.rank > 0 else 0
+        i = 0
+
+        result = np.zeros((a.comm.size,), dtype=np.int)
```
coquelin77, line 784: torch not numpy

heat/core/manipulations.py
```diff
@@ -735,6 +736,170 @@ def hstack(tup):
...
+                maxpos = i
+
+        # write counts
+        while csum2[i] < csum1[a.comm.rank]:
+            result[i] = csum2[i] - p
+            p = csum2[i]
+            i += 1
+        result[i] = csum1[a.comm.rank] - p
+
+        return tuple(result), (0,) + tuple(np.cumsum(result[:-1]))
```
coquelin77, line 807: torch not numpy

heat/core/manipulations.py
```diff
@@ -735,6 +736,170 @@ def hstack(tup):
...
+
+    def reshape_argsort_counts_displs(shape1, lshape1, displs1, shape2, displs2, axis, comm):
+        """
+        Compute the send order, counts, and displacements.
+        """
+        mask = torch.zeros(lshape1).flatten()
```
coquelin77, line 813: dtype, device, or use `zeros_like`

heat/core/manipulations.py
```diff
@@ -735,6 +736,170 @@ def hstack(tup):
...
+
+        local_index = 0
+        global_index = displs1[comm.rank] * np.prod(shape1[axis + 1 :])
+        local_len = np.prod(lshape1[axis:])
+        global_len = np.prod(shape1[axis:])
+
+        ulen = np.prod(shape2[axis + 1 :])
+
+        while local_index < np.prod(lshape1):
```
coquelin77, line 822: torch not numpy

heat/core/manipulations.py
```diff
@@ -735,6 +736,170 @@ def hstack(tup):
...
+                # search new process id
+                for j, k in zip(displs2, range(comm.size)):
+                    if j <= jpos:
+                        mask[local_index + i] = k
+            global_index += global_len
+            local_index += local_len
+
+        uniques, counts = torch.unique(mask, return_counts=True)
+
+        mpicounts = torch.zeros(comm.size)
```
coquelin77, line 835: device + dtype

heat/core/manipulations.py
```diff
@@ -735,6 +736,170 @@ def hstack(tup):
...
+
+        for i, j in zip(uniques, counts):
+            mpicounts[int(i)] = int(j)
+
+        return (
+            mask.argsort(),
+            tuple(mpicounts.numpy()),
```
coquelin77, line 842: torch not numpy

heat/core/manipulations.py
```diff
@@ -735,6 +736,170 @@ def hstack(tup):
...
+            (0,) + tuple(torch.cumsum(mpicounts[:-1], dim=0).numpy()),
+        )
+
+    # Check the type of shape and number elements
+    shape = stride_tricks.sanitize_shape(shape)
+    if np.prod(shape) != a.size:
```
coquelin77, line 848: torch not numpy

heat/core/manipulations.py
```diff
@@ -735,6 +736,170 @@ def hstack(tup):
...
+            mpicounts[int(i)] = int(j)
+
+        return (
+            mask.argsort(),
+            tuple(mpicounts.numpy()),
+            (0,) + tuple(torch.cumsum(mpicounts[:-1], dim=0).numpy()),
+        )
+
+    # Check the type of shape and number elements
+    shape = stride_tricks.sanitize_shape(shape)
```
coquelin77, line 847: shape is ambiguous in this specific function. which shape do you mean here?

heat/core/manipulations.py
```diff
@@ -735,6 +736,170 @@ def hstack(tup):
...
+        a.comm.Alltoallv((send.flatten(), sendcounts, senddispls), (data, recvcounts, recvdispls))
+
+        # original order
+        backsort = torch.argsort(recvsort)
+        data = data[backsort]
+
+    else:
+        # turn displs into cumulative sums
+        old_csum = tuple(np.prod(a.shape[1:]) * i for i in old_displs[1:]) + (a.size,)
+        new_csum = tuple(np.prod(shape[1:]) * i for i in new_displs[1:]) + (a.size,)
```
coquelin77, line 886: torch not numpy

mtar:
I have problems with some matrices. I'm getting a wrong order.

mtar:
argsort is causing this. On more than 11 elements the order is lost.

heat/core/manipulations.py
mtar: There are now a bunch of torch function calls

mtar:
Axis parameter added. It is now possible to set a different split axis for the new shape.

Issue #514, opened 2020-03-31
Make ht.var(), ht.std() NumPy-compatible


ClaudiaComito:
**Related**
This is one of the several problems that causes #490 

**Feature functionality**
torch.var and torch.std deviate from their numpy equivalent, in that the default in torch is to apply Bessel's correction (argument `unbiased=True` by default). The opposite is true in numpy, where the biased-ness or unbiased-ness of the variance is defined via the Delta Degrees of Freedom argument (`ddof=0` by default equivalent to `unbiased=False`).

At the moment, ht.var() and ht.std() behave like their torch equivalent. They should take a kwarg `ddof` and return the biased result by default.

PR #515, opened 2020-03-31
Features/514 var std numpylike

ClaudiaComito:
## Description
Working on #490 led me to opening several cans of worms and resulted in a rather messy PR (#508). I'm going to withdraw #508 and submit the main changes separately.

With this PR I'm making ht.var() and ht.std() numpy-compliant. 

Issue/s resolved: #514 and partly #490 (because the numpy default is biased variance (no Bessel correction), ht.var() of a tensor of 1 element now returns 0 instead of nan).

## Changes proposed:
- kwarg `bessel` has been replaced by `ddof`, can be 0 or 1;
- Default is `ddof=0` like for numpy (equivalent to `unbiased=False` for torch and `bessel=False` for the current heat implementation).
- argument `bessel` (bool) can still be passed so scripts shouldn't break, but remember the default is now no Bessel correction

## Type of change

Remove irrelevant options:

## Due Diligence

functions that call var() or std() relying on unbiased variance will yield the wrong result (default is now biased variance, ddof=0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - [#483](https://github.com/helmholtz-analytics/heat/pull/483) Bugfix: DNDarray.cpu() changes heat device to cpu
 - Update documentation theme to "Read the Docs"
 - [#499](https://github.com/helmholtz-analytics/heat/pull/499) Bugfix: MPI datatype mapping: `torch.int16` now maps to `MPI.SHORT` instead of `MPI.SHORT_INT`
+- [#515] (https://github.com/helmholtz-analytics/heat/pull/515) ht.var(), ht.std() numpy-compliant (ddof=0)
 
 # v0.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - [#483](https://github.com/helmholtz-analytics/heat/pull/483) Bugfix: DNDarray.cpu() changes heat device to cpu
 - Update documentation theme to "Read the Docs"
 - [#499](https://github.com/helmholtz-analytics/heat/pull/499) Bugfix: MPI datatype mapping: `torch.int16` now maps to `MPI.SHORT` instead of `MPI.SHORT_INT`
-- [#515] (https://github.com/helmholtz-analytics/heat/pull/515) ht.var(), ht.std() numpy-compliant (ddof=0)
+- [#515] (https://github.com/helmholtz-analytics/heat/pull/515) ht.var() now returns the unadjusted sample variance by default, Bessel's correction can be applied by setting ddof=1.
 
 # v0.3.0
 
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -3399,6 +3399,15 @@ class DNDarray:
             represents the number of elements. Default: ddof=0. If ddof=1, the Bessel correction will be applied.
             Setting ddof > 1 raises a NotImplementedError.
 
+        Notes on ddof (from numpy)
+        --------------------------
+        The variance is the average of the squared deviations from the mean, i.e., var = mean(abs(x - x.mean())**2).
+        The mean is normally calculated as x.sum() / N, where N = len(x). If, however, ddof is specified, the divisor
+        N - ddof is used instead. In standard statistical practice, ddof=1 provides an unbiased estimator of the
+        variance of a hypothetical infinite population. ddof=0 provides a maximum likelihood estimate of the variance
+        for normally distributed variables.
+
+
         Examples
         --------
         >>> a = ht.random.randn(1,3)
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -3037,7 +3037,7 @@ class DNDarray:
         """
         return manipulations.squeeze(self, axis)
 
-    def std(self, axis=None, bessel=True):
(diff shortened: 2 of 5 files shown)


coquelin77:
i disagree with making a biased estimator the default of any statistical measure, regardless of how numpy does it.

ClaudiaComito:
> i disagree with making a biased estimator the default of any statistical measure, regardless of how numpy does it.

I have a lot of understanding and in my mind I've been calling this the "doof" argument all along.

Not sure it's worth jeopardizing our `import heat as np` stance for this though. @Markus-Goetz ?

coquelin77:
to further stir the pot, pytorch's default is an unbiased estimator

coquelin77:
> to further stir the pot, pytorch's default is an unbiased estimator

and to extend that, this would need to be changed in the var function as well to calculate the correct estimator.

ClaudiaComito:
> > to further stir the pot, pytorch's default is an unbiased estimator
> 
> and to extend that, this would need to be changed in the var function as well to calculate the correct estimator.

I think I've addressed all of this, but let me know once you had a look.

coquelin77:
> > i disagree with making a biased estimator the default of any statistical measure, regardless of how numpy does it.
> 
> I have a lot of understanding and in my mind I've been calling this the "doof" argument all along.
> 
> Not sure it's worth jeopardizing our `import heat as np` stance for this though. @Markus-Goetz ?

i do not think that we are jeopardizing this as the code will run clean with the bessel correction as default. I think that we should be using unbiased estimators statistics whenever possible. there are already differences which make it so that `import heat as np` does not distribute the data between processes. i think that the use of unbiased estimators as defaults can be put into this pile. i disagree with the default value setting in numpy in this matter and i do not think that we should follow numpy blindly.

ClaudiaComito:
> > > i disagree with making a biased estimator the default of any statistical measure, regardless of how numpy does it.
> > 
> > 
> > I have a lot of understanding and in my mind I've been calling this the "doof" argument all along.
> > Not sure it's worth jeopardizing our `import heat as np` stance for this though. @Markus-Goetz ?
> 
> i do not think that we are jeopardizing this as the code will run clean with the bessel correction as default. I think that we should be using unbiased estimators statistics whenever possible. there are already differences which make it so that `import heat as np` does not distribute the data between processes. i think that the use of unbiased estimators as defaults can be put into this pile. i disagree with the default value setting in numpy in this matter and i do not think that we should follow numpy blindly.

The code doesn't run clean with the Bessel correction in edge cases, where for example ht.var() of a 1-element tensor returns nan instead of 0. 

I'm not here to make a statement about best practices in statistics. I think deviating from numpy and setting `ddof=1` as default in this case is a bad idea:

- numpy users will expect the exact same results from heat for the same function call; by the way, sklearn uses the default np.var() profusely as well;
- we know for a fact that the current default (bessel=True) breaks the tests in the edge cases where splitting produces a 1-element tensor on 1 node; 
- we will have instances of this (1-element tensors) happening in real life, not just in tests; 
- finally but most importantly, being able to split tensors is the reason why HeAT exists! Unbiased estimators are not. We're talking about a 1/n factor for the variance, where n is usually very large. This is not at all the same pile. I don't see this as following numpy blindly, rather I feel that the opposite would be following torch blindly and setting us up for trouble unnecessarily. 

Let's agree to disagree as far as I'm concerned, and maybe the others should chip in as well at this point.

coquelin77:
this does a better job of making the case as to why it should be 0 and not 1. they didnt chose it arbitrarily. 

The mean is normally calculated as ``x.sum() / N``, where ``N = len(x)``.
If, however, `ddof` is specified, the divisor ``N - ddof`` is used
instead.  In standard statistical practice, ``ddof=1`` provides an
unbiased estimator of the variance of a hypothetical infinite population.
``ddof=0`` provides a maximum likelihood estimate of the variance for
normally distributed variables.

ClaudiaComito:
> this does a better job of making the case as to why it should be 0 and not 1. they didnt chose it arbitrarily.

Well, it never occurred to me that they did it arbitrarily!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - [#483](https://github.com/helmholtz-analytics/heat/pull/483) Bugfix: DNDarray.cpu() changes heat device to cpu
 - Update documentation theme to "Read the Docs"
 - [#499](https://github.com/helmholtz-analytics/heat/pull/499) Bugfix: MPI datatype mapping: `torch.int16` now maps to `MPI.SHORT` instead of `MPI.SHORT_INT`
+- [#515] (https://github.com/helmholtz-analytics/heat/pull/515) ht.var(), ht.std() numpy-compliant (ddof=0)
```
coquelin77, line 9: can you be more specific? `default is now max likelyhood ... instead of unbiased estimator`

coquelin77:
> > this does a better job of making the case as to why it should be 0 and not 1. they didnt chose it arbitrarily.
> 
> Well, I never thought that they did it arbitrarily! By the way thanks for giving me the chance to refresh statistics 101...

I was not sure why they chose it before. i still dont agree with it but if it has a logical reason behind it which i agree with then i cannot continue to oppose it simply because i have an opinion which leans the other way. (also this is not a major change, it was mostly a principle thing for me)

CHANGELOG.md
ClaudiaComito: OK done, both here and in the docs

Issue #438, opened 2019-12-18
Add CUDA-aware MPI support for other MPI stacks


Markus-Goetz:
**Feature functionality**
Heat currently supports CUDA-aware MPI for OpenMPI only. It would be good if we also supported other implementations. The minimum set of implementation should be MPICH and MVAPICH. Parastation, IntelMPI (?) and IBM MPI would be optional other candidates.

This requires changes to the header of the communication.py 

**Additional context**
MVAPICH has an environment variable called MV2_USE_CUDA set to 1 if CUDA-aware MPI is present.

PR #522, opened 2020-04-02
Added CUDA-aware MPI support detection for MVAPICH, MPICH and ParaStationMPI

Markus-Goetz:
## Description

Added support for CUDA-aware MPI for the other three supporting MPI platforms - MVAPICH, MPICH and ParaStationMPI.

NOTE: if your MPI installation is compiled with CUDA-aware MPI support it may still not use it by default. For most of them a specific environment variable needs to be set our code checks for.

Issue/s resolved: #438

## Type of change

Remove irrelevant options:

## Due Diligence

no

### DIFF
--- a/heat/core/communication.py
+++ b/heat/core/communication.py
@@ -13,11 +13,11 @@ if "openmpi" in os.environ.get("MPI_SUFFIX", "").lower():
     buffer = subprocess.check_output(["ompi_info", "--parsable", "--all"])
     CUDA_AWARE_MPI = b"mpi_built_with_cuda_support:value:true" in buffer
 # MVAPICH
-CUDA_AWARE_MPI = CUDA_AWARE_MPI and os.environ.get("MV2_USE_CUDA") == "1"
+CUDA_AWARE_MPI = CUDA_AWARE_MPI or os.environ.get("MV2_USE_CUDA") == "1"
 # MPICH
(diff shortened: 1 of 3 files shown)


Markus-Goetz:
I tested MVAPICH, seems to work fine on our cluster. Parastation has been tested on HDFML, it results in a permission denied error that Alex Strube is currently working on, but it definitely changes the behaviour of the MPI stack. I was unable to test MPICH as there is non available.

heat/core/communication.py
```diff
@@ -7,12 +7,17 @@
...
 from .stride_tricks import sanitize_axis
 
+CUDA_AWARE_MPI = False
 # check whether OpenMPI support CUDA-aware MPI
 if "openmpi" in os.environ.get("MPI_SUFFIX", "").lower():
     buffer = subprocess.check_output(["ompi_info", "--parsable", "--all"])
     CUDA_AWARE_MPI = b"mpi_built_with_cuda_support:value:true" in buffer
-else:
-    CUDA_AWARE_MPI = False
+# MVAPICH
```
coquelin77, line 15: is there any way to automatically get this? or even better, to automatically turn on cuda mpi?

Markus-Goetz: Well the modules could set the respective environment variable on load. Programatically from Heat there is no way (that I know of) in telling whether the binaries are actually compiled with CUDA support. There is the hacky possibility of checking ldd output and looking whether the MPI libs attempt to dynamically load the CUDA shared objects.

Issue #422, opened 2019-11-28
Concatenate changes comm


TheSlimvReal:
**Description**
When using the `concatenate` function, the comm of the result is eventually a different one than the comm of the input arrays. The comm will be changed when it's not `ht.MPI_WORLD` but another custom one. The `concatenate` function does not specify a comm internally and thus the `ht.MPI_WORLD` comm will be used for the result.

**To Reproduce**
```
a = ht.arange(23, split=0, comm=MPICommunication())
b = ht.zeros(2, comm=a.comm)
print('a', a.comm)
print('b', b.comm)

c = ht.concatenate((b, a))

print('c', c.comm)
```
1. Which module/class/function is affected?
`concatenate` in `manipualtions`
2. What are the circumstances under which the bug appears?
setting a custom communicator
3. What is the exact error-message/errorous behavious?
No error message, just unexpected behaviour.

**Expected behavior**
A comm of the input arrays should be used

**Additional comments**
I am not quite sure what the expected behaviour here would be.

coquelin77:
there are 2 options for this.
1. we leave it as it is. in this case the returned DNDarray has its own communicator and it is assumed that we are working on the whole world. 
2. we use the communicator from one of the inputs

I am partial to the first. i think that a new array should have its own communicator. however I have just noticed a different bug in cat -> the inputs are unbalanced in the process of the code but they are not re-balanced at the end of the functoin

Markus-Goetz:
comms of a and b should be consistent. So it will be the comm of either one of the operators

PR #525, opened 2020-04-03
Consistent comms for concatenate

Markus-Goetz:
## Description

Checks for consistency of comms in the concatenate function and passes them to the result DNDarray.

Issue/s resolved: #422

## Changes proposed:
- Comms must be identical, enforced via RuntimeException
- Operands comm off arr0 is passed to result DNDarray

## Type of change

Remove irrelevant options:
- Documentation update

## Due Diligence

no

### DIFF
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -41,6 +41,15 @@ def concatenate(arrays, axis=0):
     res: DNDarray
         The concatenated DNDarray
 
+    Raises
+    ------
+    RuntimeError
+        If the concatted DNDarray meta information, e.g. split or comm, does not match.
+    TypeError
+        If the passed parameters are not of correct type (see documentation above).
+    ValueError
+        If the number of passed arrays is less than two or their shapes do not match.
+
     Examples
     --------
     >>> x = ht.zeros((3, 5), split=None)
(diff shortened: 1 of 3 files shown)


heat/core/manipulations.py
```diff
@@ -93,62 +102,80 @@ def concatenate(arrays, axis=0):
...
         raise ValueError(
-            "Arrays cannot be concatenated, gshapes must be the same in every axis "
+            "Arrays cannot be concatenated, shapes must be the same in every axis "
             "except the selected axis: {}, {}".format(arr0.gshape, arr1.gshape)
         )
 
-    s0, s1 = arr0.split, arr1.split
+    # different communicators may not be concatenated
+    if arr0.comm != arr1.comm:
+        raise RuntimeError("Communicators of passed arrays mismatch.")
```
coquelin77, line 138: i think that this is a dangerous precedent to set. because when we create new arrays we frequently do not ensure that they have the same comms. we could generate one global comm and the use this for all created arrays in the script. that would ensure that we have the same object as default everywhere

Markus-Goetz: Well we do have one global communicator, which is COMM_WORLD and we actually need to do what is coded above everywhere. What happens if we have two different comms for some variables a and b that have size 3 and 5 respectively?

Issue #528, opened 2020-04-06
ht.eye fails for distributed non-squared tensors


Cdebus:
**Description**
Creating a distributed 2D ht.eye with different x and y dimensions throws errors

**To Reproduce**
1. Affected: heat>factories>eye
2.   To Reproduce:
```
      a = ht.eye( (ht.MPI_WORLD.size * 2, ht.MPI_WORLD.size * 3), dtype=ht.float32, split=1)
     >> mpirun -n 2 
```
3. Error message
```
  File "/p/home/jusers/debus1/hdfml/heat/heat/core/factories.py", line 524, in eye
    data[pos_x][pos_y] = 1
IndexError: index 4 is out of bounds for dimension 0 with size 4
```

PR #540, opened 2020-04-09
lasso examples

krajsek:
## Description

This PR removes PyTorch and Numpy examples from example/lasso/demo.py

Issue/s resolved: #528

## Changes proposed:
Removed PyTorch and Numpy Lasso examples

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/regression/lasso.py
+++ b/heat/regression/lasso.py
@@ -116,17 +116,26 @@ class Lasso(ht.RegressionMixin, ht.BaseEstimator):
         # Get number of model parameters
         _, n = X.shape
 
+        if len(y.shape) == 1:
+            y = ht.expand_dims(y, axis=1)
+
         # Initialize model parameters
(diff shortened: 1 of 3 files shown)


heat/regression/lasso.py
```diff
@@ -116,17 +116,29 @@ def fit(self, X, y):
         # Get number of model parameters
         _, n = X.shape
 
+        assert len(y.shape) <= 2, "y.shape needs to be 1 or 2 dimensional"
+        assert len(X.shape) == 2, "X.shape needs to be 2 dimensional"
```
coquelin77, line 120: should these be raises instead?

Markus-Goetz: Would also claim that should be consistent exceptions

krajsek: Is there a predefined error type for shapes in Python?

Markus-Goetz: Should be ValueError

Issue #517, opened 2020-04-01
Enhance HDF5 and netCDF4 API


Markus-Goetz:
**Feature functionality**
Enhance the HDF5 and netCDF4 API to allow for attribute writes/slicing/naming.

PR #559, opened 2020-04-24
Features/517-netcdf4-features

ben-bou:
## Description
Adding the following features to the `save_netcdf` method:
- Naming of dimensions 
- Creating unlimited dimensions
- Usage of already existing dimensions and variables
- Specifying the Location of data in the netcdf4-File using indices or slices

Issue/s resolved: #517 (partially)

## Changes proposed:
-
-
-
-

## Due Diligence

no


coquelin77:
some of the formatting is wrong here. this is fixed automatically by pre-commit (most of the time). all you need to do is install pre-commit (`pip install pre-commit` or ` conda install -c conda-forge pre-commit`) then run `pre-commit --instal` and `pre-commit -run --all-files`

ben-bou:
Thanks, I ran pre-commit and updated the brach.

coquelin77:
bump. is this being worked on? to run the tests locally you can use `mpirun -np # ...`

ben-bou:
What do you mean, is anything missing?

coquelin77:
the tests are failing. it looks like something in the IO is hanging. you can see it in the Travis CI - Pull Request details

ben-bou:
In Travis CI, IO is hanging, but if I do
`mpiexec -n 4 coverage run --source=heat --parallel-mode -m pytest heat/ && coverage combine && coverage report && coverage xml`, it passes for 1...4 processes.
Any suggestions? Maybe there are incompatibilities in the underlying netcdf Versions?

coquelin77:
check the versions. it is listed in the log for the travis builds but i have also listed it here:

```
Package        Version  Location
-------------- -------- --------
atomicwrites   1.3.0    
attrs          19.1.0   
certifi        2019.3.9 
cftime         1.1.2    
chardet        3.0.4    
codecov        2.0.15   
coverage       4.5.3    
future         0.18.2   
h5py           2.10.0   
heat           0.3.0    /heat   
idna           2.8      
more-itertools 7.0.0    
mpi4py         3.0.3    
netCDF4        1.5.2    
numpy          1.18.4   
pip            19.0.3   
pluggy         0.9.0    
py             1.8.0    
pytest         4.4.0    
requests       2.21.0   
setuptools     40.8.0   
six            1.12.0   
torch          1.5.0    
urllib3        1.24.1 
```

ben-bou:
Hi, I've got a quick question:

Lets say I've got a dndarray `data` and a shape `out_shape`, which is basically the same as `data.shape` but it might have ones added in between. Is there a way to find out the hypothetical split-axis of `data` if it was broadcast to `out_shape`?

ben-bou:
ok, seems like `out_split = data.split + out_shape[: data.split].count(1) - data.shape[: data.split].count(1)` does this

ben-bou:
These formatting errors have nothing to do with my changes. Where do they suddenly come from? Is there a way to suppress them? I don't get any errors when running `pre-commit` locally.

coquelin77:
please extend the unit tests to cover the new additions

ben-bou:
The coverage is so low because codecov does not run with parallel netcdf and therefore disregards a large part as not covered

heat/core/io.py
```diff
@@ -354,36 +395,181 @@ def save_netcdf(data, path, variable, mode="w", **kwargs):
...
+                        variable, data.dtype.char(), dimension_names, **kwargs
+                    )
+                mergedSlices = mergeSlices(var, file_slices, data)
+                try:
+                    var[mergedSlices] = (
+                        data._DNDarray__array.cpu()
+                        if is_split
+                        else data._DNDarray__array[slices].cpu()
+                    )
+                except RuntimeError:
```
coquelin77, line 547: (im unfamiliar with netCDF) why is this raised?

heat/core/io.py
```diff
@@ -354,36 +395,181 @@ def save_netcdf(data, path, variable, mode="w", **kwargs):
...
+            stop = start + stride * count
+            new_slices = []
+            for begin, end, step in zip(start, stop, stride):
+                if begin.size == 1:
+                    begin, end, step = begin.item(), end.item(), step.item()
+                    new_slices.append(slice(begin, end, step))
+                else:
+                    begin, end, step = begin.flatten(), end.flatten(), step.flatten()
+                    new_slices.append(
+                        np.r_[
```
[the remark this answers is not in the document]
ben-bou: I did not test any of this on GPUs. I'm using numpy here because I'm working on `start, count, stride` which are generated by `nc.utils` and are ndarrays. That's ok because `var_slices` is usually (a tuple of) slices/ints/lists and the same for every process. You're right though, this will fail when using a dndarray for indexing because a netcdf.Variable does not support to be indexed by a dndarray (results in `IndexError: only integers, slices (`:`), ellipsis (`...`), and 1-d integer or boolean arrays are valid indices`). However, it does support being indexed by a `torch.tensor` so it might be interesting to see what happens when using a gpu-tensor for indexing.

heat/core/io.py
```diff
@@ -354,36 +395,181 @@ def save_netcdf(data, path, variable, mode="w", **kwargs):
...
+                if isinstance(new_slices[out_split], slice):
+                    start, stop, step = (
+                        new_slices[out_split].start,
+                        new_slices[out_split].stop,
+                        new_slices[out_split].step,
+                    )
+                    sliced = range(start, stop, step)[slices[data.split]]
+                    a, b, c = sliced.start, sliced.stop, sliced.step
+                    a = None if a < 0 else a
+                    b = None if b < 0 else b
```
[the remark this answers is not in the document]
ben-bou: Actually, it might be the most elegant solution to just use `sliced` (which is of type range) as index instead of transforming it into a slice

heat/core/io.py
ben-bou: Sometimes, collective IO is needed (e.g. for writing into an unlimited dimension). A `RuntimeError` is thrown when attempting this without collective IO.
Someone with more experience in netcdf might know a better convention for when to use collective IO.

heat/core/io.py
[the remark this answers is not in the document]
ben-bou: Not that its relevant here, but why does indexing with a dndarray behave different than a torch.tensor?
```python
In [1]: import heat as ht                                                                                                                                                                                   
In [2]: import numpy as np                                                                                                                                                                                  
In [3]: import torch                                                                                                                                                                                        
In [4]: arr = ht.arange(3).expand_dims(1) + ht.arange(3)                                                                                                                                                    
In [5]: arr                                                                                                                                                                                                 
Out[5]: 
tensor([[0, 1, 2],
        [1, 2, 3],
        [2, 3, 4]], dtype=torch.int32)

In [6]: ind = [0,1]                                                                                                                                                                                         
In [7]: arr[ind]                                                                                                                                                                                            
Out[7]: 
tensor([[0, 1, 2],
        [1, 2, 3]], dtype=torch.int32)

In [8]: arr[np.array(ind)]                                                                                                                                                                                  
Out[8]: 
tensor([[0, 1, 2],
        [1, 2, 3]], dtype=torch.int32)

In [9]: arr[torch.tensor(ind)]                                                                                                                                                                              
Out[9]: 
tensor([[0, 1, 2],
        [1, 2, 3]], dtype=torch.int32)

In [10]: arr[tuple(ind)]                                                                                                                                                                                    
Out[10]: tensor(1, dtype=torch.int32)

In [11]: arr[ht.array(ind)]                                                                                                                                                                                 
Out[11]: tensor(1, dtype=torch.int32)
```

heat/core/io.py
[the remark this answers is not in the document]
ben-bou: > Actually, it might be the most elegant solution to just use sliced (which is of type range) as index instead of transforming it into a slice

No, even though slicing with ranges is generally possible, this fails the `indexing netcdf file: broadcasting var`-test.

I don't think step needs to be corrected because it is generated by nc.utils (so possible 'weird' values are already dealt with) and is used to calculate stop. Also the slicing doesn't change the step because ht.comm.chunk() always uses `None` as step.

heat/core/io.py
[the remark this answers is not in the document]
coquelin77: the second is because the advanced indexing using heat DNDarrays is not yet implemented, but that is on the list of things to do.

@Markus-Goetz what is the verdict on NumPy here? im not sure. because its a saving routine it is bound to the CPU anyway. but we would need to specify that higher up in the function. if that is done then i have no problem with it.

heat/core/io.py
```diff
@@ -354,36 +396,178 @@ def save_netcdf(data, path, variable, mode="w", **kwargs):
...
+                ne_before_split
+            ]  # index of (first nonempty element after split) in squeezed shape
+            return max(
+                i
+                for i, v in enumerate(expandedShape[: max(ex_ind_nonempty[:ind_ne_after_split])])
+                if v == 1
+            )
+
+        def __merge_slices(var, var_slices, data, data_slices=None):
+            """ Using var[var_slices][data_slices] = data
```
Markus-Goetz, line 453: linebreak

Markus-Goetz: Documentation a little hard to follow what it actually is doing

heat/core/io.py
```diff
@@ -354,36 +396,178 @@ def save_netcdf(data, path, variable, mode="w", **kwargs):
...
+                datashape=data.shape,
+                put=True,
+            )
+            out_shape = nc._netCDF4._out_array_shape(count)
+            out_split = __get_expanded_split(data.shape, out_shape, data.split)
+
+            start, count, stride = start.T, count.T, stride.T  # transpose for iteration
+            stop = start + stride * count
+            new_slices = []
+            for begin, end, step in zip(start, stop, stride):
```
Markus-Goetz, line 493: Can we do this here vectorized, seems possible with a little division

heat/core/io.py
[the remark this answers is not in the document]
Markus-Goetz: Have an issue about the advanced indexing. 

Numpy seems fine here, ensure that the data is actually on cpu if necessary

heat/core/io.py
```diff
@@ -354,36 +396,178 @@ def save_netcdf(data, path, variable, mode="w", **kwargs):
...
+            return max(
+                i
+                for i, v in enumerate(expandedShape[: max(ex_ind_nonempty[:ind_ne_after_split])])
+                if v == 1
+            )
+
+        def __merge_slices(var, var_slices, data, data_slices=None):
+            """ Using var[var_slices][data_slices] = data
+            combines a __getitem__ with a __setitem__ call, therefore it does not allow
+            parallelization of the write-operation and does not work is var_slices = slice(None)
```
Markus-Goetz, line 455: Not clear why parallelization would not work here. I am hesitant to merge this

heat/core/io.py
```diff
@@ -354,36 +396,178 @@ def save_netcdf(data, path, variable, mode="w", **kwargs):
...
+                if variable in handle.variables:
+                    var = handle.variables[variable]
+                else:
+                    for name, elements in zip(dimension_names, data.shape):
+                        if name not in handle.dimensions:
+                            handle.createDimension(name, elements if not is_unlimited else None)
+                    var = handle.createVariable(
+                        variable, data.dtype.char(), dimension_names, **kwargs
+                    )
+                mergedSlices = __merge_slices(var, file_slices, data)
```
Markus-Goetz, line 538: underscores

heat/core/io.py
Markus-Goetz: All metadata operations, e.g. creation, deletion, ..., are collective. This means all processors need to call this operation otherwise the application deadlocks. Should only be necessary in parallel mode to begin with

heat/core/io.py
ben-bou: My reasoning is that because var_slices specifies the location of the (global) dndarray (and therefore is the same for every process), doing `var[var_slices]` loads the global data onto every single process as a ndarray, only to overwrite the local chunk using `[data_slices] = data`. I believe this does not allow parallelization because in the only method we call on the netcdf.Variable, the `__getitem__`, every process requests access to same location (var_slices).

Also, when I tried to use `var[var_slices][data_slices] = data`, the tests failed (in some cases, it did not write at all), so I came up with the idea to merge the var_slices and data_slices into a single set (tuple) of slices (or other indices).

heat/core/io.py
ben-bou: I'm not sure how to vectorize this

heat/core/io.py
coquelin77: there was an update to the slicing recently. can you see if the new slicing solved your issue?

Markus-Goetz:
Also, I will have time to look at it next week an make some addition/changes. Sorry for the long delay

coquelin77:
bump, there was an update which removed the `ht_device` parameter. see other tests for how to use it

heat/core/io.py
ben-bou: Thanks, but I don't believe this is relevent here.

Maybe I didn't make it clear what this method actually does:
Each process knows its location inside the global dndarray via `comm.chunk()`. This method calculates (for each process) its location inside the (global) netCDF-File, such that every process can write to its corresponding location.  Therefore this method does not apply any slicing in HeAT.

Anyway, where can I find the changes made to slicing? I couldnt find them in the changelog

mtar:

heat/core/io.py
ClaudiaComito: Hi Ben,

`comm.chunk` also returns the offset of the local 0-index with respect to the global one, would  using `var.larray[var_slices-offset]`  allow you to parallelize the write operation? (if I understand the problem that is)

heat/core/io.py
```diff
@@ -352,34 +391,178 @@ def save_netcdf(data, path, variable, mode="w", **kwargs):
...
+                    for name, elements in zip(dimension_names, data.shape):
+                        if name not in handle.dimensions:
+                            handle.createDimension(name, elements if not is_unlimited else None)
+                    var = handle.createVariable(
+                        variable, data.dtype.char(), dimension_names, **kwargs
+                    )
+                merged_slices = __merge_slices(var, file_slices, data)
+                try:
+                    var[merged_slices] = (
+                        data._DNDarray__array.cpu()
```
ClaudiaComito, line 536: Hi Ben, please replace `_DNDarray__array__` with the new `larray` for consistency with the rest - also in the follow-up code

heat/core/io.py
```diff
@@ -352,34 +391,178 @@ def save_netcdf(data, path, variable, mode="w", **kwargs):
...
+                        if name not in handle.dimensions:
+                            handle.createDimension(name, elements if not is_unlimited else None)
+                    var = handle.createVariable(
+                        variable, data.dtype.char(), dimension_names, **kwargs
+                    )
+                var.set_collective(False)  # not possible with non-parallel netcdf
                 if is_split:
-                    var[slices] = data.larray.cpu()
+                    merged_slices = __merge_slices(var, file_slices, data)
+                    var[merged_slices] = data._DNDarray__array.cpu()
```
ClaudiaComito, line 563: `_DNDarray__array` --> `larray`

heat/core/io.py
ben-bou: Hi Claudia,
I don't understand why you would index the local array with the global offset? Also, please keep in mind that `var` is the `netcdf4.Variable`-object. I am now reworking the documentation for this method to clarify what it does

ben-bou: Hopefully, this is more clear:

            This method allows replacing:
            ``var[var_slices][data_slices] = data``
            (a `netcdf4.Variable.__getitem__` and a `numpy.ndarray.__setitem__` call)
            with:
            ``var[ __merge_slices(var, var_slices, data, data_slices) ] = data``
            (a single `netcdf4.Variable.__setitem__` call)

            This is necessary because performing the former would, in the `__getitem__`, load the
            global dataset onto every process in local `numpy-ndarrays`. Then, the `__setitem__`
            would write the local `chunk` into the `numpy-ndarray`.

            The latter allows the netcdf4 library to parallelize the write-operation by directly
            using the `netcdf4.Variable.__setitem__` method.

ClaudiaComito:
Thanks a lot @ben-bou for your contribution!

Issue #561, opened 2020-04-27
ht.squeeze() doesn't follow split semantics


ClaudiaComito:
**Description**
ht.squeeze(axis) returns a non-distributed tensor in all cases, it should follow split semantics, e.g.: 

`output.split = None if input.split == None or input.split in axis`
`output.split = input.split - n`, where `n` is the number of elements in `axis` that are smaller than input.split

Issue #273, opened 2019-05-23
Reinstate portion of test_squeeze() testing uneven tensor distribution across ranks


ClaudiaComito:
In test_manipulations.test_squeeze(), the "4D split tensor" test is commented out, as it leads to uneven distribution of columns across the ranks (PR #231).

Solution: in manipulations.squeeze() replace Allgather with  Allgatherv, specify "displs" argument (see link below). This will be possible with the upcoming Allgatherv update (Issue #233). 

MPI_Allgatherv standard doc: 
https://www.mpich.org/static/docs/latest/www3/MPI_Allgatherv.html

ClaudiaComito:
Unnecessary after merging #562

PR #562, opened 2020-04-27
Bug/squeeze split semantics

ClaudiaComito:
## Description

 Changes to manipulations.squeeze() to follow split semantics. In the previous version, the output tensor was being "gathered" and the function returned a non-distributed tensor. 

Split semantics for squeeze() are now as follows: a distributed tensor will keep its original split dimension after "squeezing", which, depending on the squeeze axis, may result in a lower numerical 'split' value, as in:
```
    >>> x.shape
    (10, 1, 12, 13)
    >>> x.split
    2
    >>> x.squeeze().shape
    (10, 12, 13)
    >>> x.squeeze().split
    1

```
The edge case` x.split = axis` returns a non-distributed tensor.

Issue/s resolved: #561,  #273 

## Changes proposed:
- I removed the Allgather part and made sure the global output shape was set correctly in returning the DNDarray. I also reinstated the test_squeeze() portion on unevenly distributed tensors.

 
## Type of change

Bug fix, breaking change. 

## Due Diligence

yes  if they rely on ht.squeeze() returning a non-distributed tensor.


coquelin77:
sorry, these tests are already included. merging now

Issue #570, opened 2020-05-19
Possible bug or undefined behavior in CG or matmul?


d1saster:
**Description**
If I run the following code with two MPI ranks, it gives the wrong solution to the linear system:
```(python3)
#!/usr/bin/env python
import heat as ht
# silence all HeAT internal warnings
import warnings
warnings.filterwarnings("ignore")

def some_linear_model(params, input_data):
    return (params[2] * input_data + params[1]) * input_data + params[0]

gen_params = ht.array([1.0, 2.0, -0.5],split=None)

x_data = ht.random.rand(1000,1,dtype=ht.float64,split=0)
x_data = x_data + x_data.comm.rank
y_data = some_linear_model(gen_params, x_data)

def jacobian_of_some_linear_model(input_data):
    return ht.hstack([ht.ones_like(input_data), input_data, input_data**2])

A = jacobian_of_some_linear_model(x_data)

# introducing the following two resplits fixes the problem.
rhs_cg = A.T @ y_data
#rhs_cg = ht.resplit(rhs_cg, axis=None)

lhs_cg = A.T @ A
#lhs_cg = ht.resplit(lhs_cg, axis=None)

rhs_cg = rhs_cg.flatten()
params_cg = ht.linalg.cg(lhs_cg, rhs_cg, ht.zeros_like(rhs_cg))

print("Solution CG:", params_cg)
print("Error CG:   ", params_cg-gen_params)

```

**To Reproduce**
`mpiexec -np 2 bugcg.py` gives the solution [1,1,-0.5].

**Expected behavior**
The solution to the linear system should be [1,2,-0.5].

**Version Info**
master

**Additional comments**
It is odd that an iterative algorithm like CG gives the correct result with the exception of one entry, which is off by exactly 1. Since CG is not doing much communication itself, I suspect that this somehow resulst from the underlying matmul behavior. But even if there would be something wrong, I find it puzzling that CG still manages to get the correct results for the other two components.

Cdebus:
Will check it out

coquelin77:
found the problem. it was an error in the matmul logic for split 1 and split 0. a PR with a fix is should be up today or tomorrow.

PR #573, opened 2020-05-25
bugfix for matmul bug (split 10)

coquelin77:
Matmul had a bug for when kB (the inner blocking dimension is 1, when this was the case, it would do a second loop which would re-add previous results to the result array.

## Description

matmul of 2 vectors is handled as an early out
in split 10 case for matmul, the remainder handling does not run if kB is equal to 0

Issue/s resolved: #570 

## Changes proposed:
- see description

## Type of change

## Due Diligence

no


coquelin77:
codecov patch is failing but i cannot see why. however, the full coverage is hopefully fine.

PR #577, opened 2020-06-02
feature ndim

mtar:
## Description

Add method ndim to dndarray for NumPy-API-compatibility. Will replaces numdims()

Issue/s resolved: None

## Changes proposed:
- additional DNDarray method ndim
- deprecation warning on numdims
- replace function calls

## Type of change

New feature

## Due Diligence

no


mtar:
I think so, too. Neither Numpy nor Pytorch use 'numdims'. ndim should be the preferred method name. 
Does Numpy have any aliases for this function?

Pytorch's equivalent function is called dim(). ndim() and ndimension() are two aliases I'm aware of. Should we add them? I don't think that we need those as we're striving for numpy API compatibility.

Issue #579, opened 2020-06-03
implement tolist


mtar:
**Feature functionality**
A clear and concise description of what the feature should do exactly and any alternative solutions.

Return a python list of a tensor.

PR #581, opened 2020-06-04
tolist() implementation

mtar:
## Description

Issue/s resolved: #579 

Implentation of tolist() method for DNDarray class. Returns a copy of the DNDarray as a python list.

## Changes proposed:
- add tolist method in dndarray

## Type of change

- new feature

## Due Diligence

no

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -3474,13 +3474,13 @@ class DNDarray:
         -------
         list: list
             A copy of the array as a python list.
-        
+
         Examples
         --------
         >>> a = ht.array([[0,1],[2,3]])
         >>> a.tolist()
         [[0, 1], [2, 3]]
-        
+
         >>> a = ht.array([[0,1],[2,3]], split=0)
         >>> a.tolist()
         (1/2) [[0, 1]]
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -3479,11 +3479,10 @@ class DNDarray:
 
         >>> a = ht.array([[0,1],[2,3]], split=0)
         >>> a.tolist()
-        (1/2) [[0, 1]]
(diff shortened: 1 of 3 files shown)


heat/core/dndarray.py
```diff
@@ -3479,6 +3479,33 @@ def tanh(self, out=None):
         """
         return trigonometrics.tanh(self, out)
 
+    def tolist(self):
```
Markus-Goetz, line 3482: Return type annotation, remove the return category

heat/core/dndarray.py
```diff
@@ -3479,6 +3479,33 @@ def tanh(self, out=None):
...
+        >>> a.tolist()
+        (1/2) [[0, 1]]
+        (2/2) [[2, 3]]
+
+        >>> a = ht.array([[0,1],[2,3]], split=1)
+        >>> a.tolist()
+        (1/2) [[0], [2]]
+        (2/2) [[1], [3]]
+        """
+        return self.__array.tolist()
```
Markus-Goetz, line 3507: The question is what we are expecting here. Do we want to have the partial lists of each DNDarray part or a gathered version. If we want to be consistent, e.g. with numpy(), we actually want to have the resplit(None) behaviour here.

mtar: I added a parameter keepsplit which defaults to False. Personally, however, I would only expect the full list if split==None, and a partial list otherwise.

Markus-Goetz: And this behaviour is implemented?

mtar: Keepdim defaults to False, so it will first gather the whole array (this is what you suggested)

Markus-Goetz: Yep my bad, one of the code lines was hidden in the preview. All good

PR #583, opened 2020-06-05
Features rot90

mtar:
## Description

Issue/s resolved: #555 

Introduces rot90. Rotate an array on a specified plane. May change the split axis if it is in the rotation plane similar to transpose.

## Changes proposed:
- add rot90()

## Type of change

New feature

## Due Diligence

no

### DIFF
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -9,6 +9,7 @@ from . import factories
 from . import stride_tricks
 from . import tiling
 from . import types
+from. import linalg
 
 
 __all__ = [
@@ -23,6 +24,7 @@ __all__ = [
     "hstack",
     "reshape",
     "resplit",
+    "rot90",
     "sort",
     "squeeze",
     "unique",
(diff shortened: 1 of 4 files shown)


heat/core/manipulations.py
```diff
@@ -9,6 +9,7 @@
 from . import stride_tricks
 from . import tiling
 from . import types
+from . import linalg
```
Markus-Goetz, line 12: please sort in alphabetically

heat/core/manipulations.py
```diff
@@ -936,6 +938,86 @@ def reshape_argsort_counts_displs(
     return factories.array(data, dtype=a.dtype, is_split=axis, device=a.device, comm=a.comm)
 
 
+def rot90(m, k=1, axes=(0, 1)):
```
Markus-Goetz, line 941: please use return type annotation

heat/core/manipulations.py
```diff
@@ -936,6 +938,86 @@ def reshape_argsort_counts_displs(
...
+    ----------
+    m : DNDarray
+        Array of two or more dimensions.
+    k : integer
+        Number of times the array is rotated by 90 degrees.
+    axes: (2,) int list or tuple
+        The array is rotated in the plane defined by the axes.
+        Axes must be different.
+
+    Returns
```
Markus-Goetz, line 955: Remove this section

heat/core/manipulations.py
```diff
@@ -936,6 +938,86 @@ def reshape_argsort_counts_displs(
     return factories.array(data, dtype=a.dtype, is_split=axis, device=a.device, comm=a.comm)
 
 
+def rot90(m, k=1, axes=(0, 1)):
+    """
+    Rotate an array by 90 degrees in the plane specified by axes.
+    Rotation direction is from the first towards the second axis.
+    Parameters
```
Markus-Goetz, line 945: newline

heat/core/manipulations.py
```diff
@@ -936,6 +938,86 @@ def reshape_argsort_counts_displs(
...
+
+    if axes[0] >= m.ndim or axes[0] < -m.ndim or axes[1] >= m.ndim or axes[1] < -m.ndim:
+        raise ValueError("Axes={} out of range for array of ndim={}.".format(axes, m.ndim))
+
+    if m.split is None:
+        return factories.array(
+            torch.rot90(m._DNDarray__array, k, axes), dtype=m.dtype, device=m.device, comm=m.comm
+        )
+
+    k %= 4
```
Markus-Goetz, line 1004: try-except block for ``k = int(k)``

heat/core/manipulations.py
```diff
@@ -936,6 +938,86 @@ def reshape_argsort_counts_displs(
...
+
+    if k == 0:
+        return m[:]
```
Markus-Goetz, line 1007: How about just returning m without the slice? Or do you want to do an explicit copy here. If so, clone might be the way to go.

heat/core/manipulations.py
```diff
@@ -936,6 +938,86 @@ def reshape_argsort_counts_displs(
...
+    if k == 2:
+        return flip(flip(m, axes[0]), axes[1])
```
Markus-Goetz, line 1009: Albeit not wrong, highly inefficient in terms of communication. I would think local rotations plus communication would be better.

heat/core/manipulations.py
mtar: I can't really decide between these options, but a copy is more consistent with the other cases.

heat/core/manipulations.py
mtar: Can you please elaborate on this.
flip() performs a local operation if split not the axis. Or Is there communication in the creation of an intermediate DNDarray?

heat/core/manipulations.py
Markus-Goetz: Copy is more consistent

heat/core/manipulations.py
Markus-Goetz: Nevermind, supposed to be skipped

heat/core/manipulations.py
Markus-Goetz: Nevermind, supposed to be skipped

heat/core/manipulations.py
Markus-Goetz: I was wrong, either one of the flips is local before/after the one along the split.

Markus-Goetz:
Bump, master needs merge

Issue #384, opened 2019-09-13
Implement sparse matrices 


Markus-Goetz:
**Feature functionality**
Design and implement a distributed sparse DNDarray class. Come up with one or more reasonable storage formats (e.g. COO, CRS, CSC, ...)

**Additional context**
PyTorch implements COO only at the moment with a very restricted operation set.

Interface design might be inspired by Armadillo and/or SciPy.sprarse

Markus-Goetz:
Here is a great paper summarizing formats and operations on sparse tensors. I would go with a COO format first where the set of entries is roughly evenly distributed 

http://groups.csail.mit.edu/commit/papers/2016/parker-thesis.pdf

ClaudiaComito:
CSR addressed in #1028 , thanks @Mystic-Slice

ClaudiaComito:
the `ht.sparse` module has been implemented by @Mystic-Slice and released with v1.3.0. For the time being it supports distributed CSR matrices and element-wise operations. More work needed.

Related issues: #1039 #1040 #1046 #1043 #1044 

I'll close this issue for now.
---
Reviewed within #1109

PR #596, opened 2020-06-15
Features/31 outer

ClaudiaComito:
## Description

Introducing the outer product of two DNDarrays as per [numpy.outer()](https://numpy.org/doc/1.18/reference/generated/numpy.outer.html), i.e.:

`ht.outer(a, b)`

`a` and `b` must be 1-dimensional, or will be flattened by default. 
In linear algebra, if `a` is a vector of _n_ elements, and `b` is a vector of _m_ elements, the outer product `outer(a,b)` is a matrix of size (_n_, _m_) and elements:

$\begin{pmatrix}  a_0\cdot b_0  & a_0\cdot b_1 & . & . &  a_0\cdot b_m \\  a_1\cdot b_0 & a_1\cdot b_1 & . & . & a_1\cdot b_m \\  . & . & . & . & .   \\   a_n\cdot b_0 & a_n\cdot b_1 & . & . & a_n\cdot b_m \end{pmatrix} $

### Implementation
- Locally, the outer product is calculated via PyTorch's Einstein summation, as in:
`torch.einsum("i,j->ij", a, b)`

- The memory-distributed implementation works under the assumption that the vectors / DNDarrays are dense (implementation for sparse case will be addressed at a later point (#384). Procedure:
    - `a` and `b` are distributed 
    - one of the DNDarrays always stays local, the other is passed around the ranks in ring communication until all `a_i * b_j` multiplications have taken place.
    - outer product (`torch.einsum`) is calculated locally, written into the appropriate slice of `out = ht.outer(a,b)`, and never needs to be communicated among ranks
    - `out` is by definition split along 0 if `b` is sent around and `a` stays put. Conversely, it is split along 1 if `a` is sent around and `b` stays put. As  a consequence...
    - ... split semantics: I have introduced a kwarg `split` to allow users to choose the split direction of the output matrix. `split` decides which DNDarray stays and which one goes round the ranks. If split is None in the distributed case, `ht.outer(a,b).split` will be the same as `a.split` and `b.split`, i.e. 0. 

Note that the distributed case requires that at least one among `a.split`, `b.split` and `out.split` be `not None`. Otherwise we fall back to local.

Issue/s resolved: #31 

## Changes proposed:
- with respect to np.outer, ht.outer has an extra kwarg `split` to define the split dimension of the output matrix, e.g.: `out = ht.outer(a, b, split=1)`

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/linalg/basics.py
+++ b/heat/core/linalg/basics.py
@@ -881,11 +881,12 @@ def outer(a, b, out=None, split=0):
         rank = a.comm.rank
         size = a.comm.size
         outer_split = split
+        t_out_slice = 2 * [slice(None, None, None)]
 
         # Decide which DNDarray gets sent around ring communication
-        # case 3: out.split = None --> bigger (element size) DNDarray stays put, smaller one gets sent around
+        # out.split = None --> bigger (element size) DNDarray stays put, smaller one gets sent around
         if split is None:
-            # TODO replace with nbytes property when available #590
+            # TODO replace with nbytes property when available, #590
             a_nbytes = a._DNDarray__array.storage().element_size() * a.size
             b_nbytes = b._DNDarray__array.storage().element_size() * b.size
             if b_nbytes <= a_nbytes:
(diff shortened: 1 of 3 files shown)


heat/core/linalg/basics.py
```diff
@@ -827,6 +827,225 @@ def norm(a):
     return exponential.sqrt(d).item()
 
 
+def outer(a, b, out=None, split=None):
+    """
+    Compute the outer product of two 1-D DNDarrays.
+
+    Given two vectors, ``a`` = [a_0, a_1, ..., a_n] and ``b`` = [b_0, b_1, ..., b_m], the outer product is:
```
Markus-Goetz, line 834: This is not essential, perhaps useful though, could we have the Mathjax version of the docstrings here? I really liked the way it displayed in the Github message.

heat/core/linalg/basics.py
```diff
@@ -827,6 +827,225 @@ def norm(a):
...
+
+    [[a_0*b_0  a_0*b_1 ... a_0*b_m]
+     [a_1*b_0     .
+     [ ...        .
+     [a_n*b_0     .        a_n*b_m]]
+
+    Parameters
+    ----------
+
+    a(n,): DNDarray
```
Markus-Goetz, line 844: Is this the way dimension information is displayed now with the new docstring format?

heat/core/linalg/basics.py
```diff
@@ -827,6 +827,225 @@ def norm(a):
...
+
+    out(n, m): DNDarray, optional
+            A location where the result is stored
+
+    split: int, optional #TODO check out docstring format
+            Split dimension of the resulting DNDarray. Can be 0, 1, or None.
+            This is only relevant if the calculations are memory-distributed,
+            in which case default is ``split=0`` (see Note).
+
+    Note: parallel implementation of outer product, arrays are dense. #TODO sparse #384
```
Markus-Goetz, line 858: The TODO should probably disappear from the docstring itself, as it would be copied over into the text that is being read by users. Perhaps we could have this in the code?

heat/core/linalg/basics.py
```diff
@@ -827,6 +827,225 @@ def norm(a):
...
+            dest_rank = rank + 1 if rank != size - 1 else 0
+            # prepare for receiving
+            origin_rank = rank - 1 if rank != 0 else size - 1
+            actual_origin = origin_rank - p
+            if origin_rank < p:
+                actual_origin += size
+            # blocking send and recv
+            if split == 0:
+                b.comm.Send(t_b, dest_rank)
+                t_b = torch.empty(lshape_map[actual_origin], dtype=t_outer_dtype, device=t_a.device)
```
Markus-Goetz, line 1018: Can we initialize this buffer once outside the loop and then reuse it? Would save up some initialization time

heat/core/linalg/basics.py
```diff
@@ -827,6 +827,225 @@ def norm(a):
...
+                b.comm.Recv(t_b, origin_rank)
+                _, _, remote_slice = b.comm.chunk(
+                    b.gshape, b.split, rank=actual_origin, w_size=size
+                )
+                t_outer_slice[1] = remote_slice[0]
+            elif split == 1:
+                a.comm.Send(t_a, dest_rank)
+                t_a = torch.empty(lshape_map[actual_origin], dtype=t_outer_dtype, device=t_a.device)
```
Markus-Goetz, line 1026: see intialization above

heat/core/linalg/basics.py
ClaudiaComito: Thanks @Markus-Goetz ,

Now I have it down as 

```
.. math::

        
           a_0 \times b_0  & a_0 \times b_1 & . & . &  a_0 \times b_m \\
           a_1 \times b_0 & a_1 \times b_1 & . & . & a_1 \times b_m \\
           . & . & . & . & .   \\
           a_n \times b_0 & a_n \times b_1 & . & . & a_n \times b_m

```

but according to the sphynx documentation what we really want is 

```
.. math::
        :nowrap:

        \begin{pmatrix}
           a_0 \times b_0  & a_0 \times b_1 & . & . &  a_0 \times b_m \\
           a_1 \times b_0 & a_1 \times b_1 & . & . & a_1 \times b_m \\
           . & . & . & . & .   \\
           a_n \times b_0 & a_n \times b_1 & . & . & a_n \times b_m
        \end{pmatrix}
```

except flake8 complains about `invalid escape sequence '\e'`. Haven't found a good solution yet.

ClaudiaComito: OK Done now, thanks @Markus-Goetz and @Cdebus

heat/core/linalg/basics.py
ClaudiaComito: fixed

heat/core/linalg/basics.py
ClaudiaComito: right, fixed

heat/core/linalg/basics.py
ClaudiaComito: That's a good point. I'm initializing a buffer for the "traveling" data now, `t_b_run` and `t_a_run`, and slicing them to the correct size afterwards.

ClaudiaComito:
> Misclick beforehand...

too bad...

PR #598, opened 2020-06-17
Changed to intuitive casting, removed dtype tests from basic test class

coquelin77:
## Description

The dtype comparison was removed as the basic tests works with numpy, which does safe casting instead of intuitive casting. other changes were adjustments for types

Issue/s resolved: 

## Changes proposed:
- change from safe casting to intuitive casting (this come with the possibility of loss during the conversion)
-
-
-

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/types.py
+++ b/heat/core/types.py
@@ -452,13 +452,13 @@ def can_cast(from_, to, casting="intuitive"):
         Scalar, data type or type specifier to cast from.
     to : type, str, ht.dtype
         Target type to cast to.
-    casting: str {'no', 'safe', 'same_kind', 'unsafe', "intuitive"}, optional
+    casting: str {"no", "safe", "same_kind", "unsafe", "intuitive"}, optional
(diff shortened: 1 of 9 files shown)


heat/core/types.py
```diff
@@ -452,12 +452,13 @@ def can_cast(from_, to, casting="safe"):
         Scalar, data type or type specifier to cast from.
     to : type, str, ht.dtype
         Target type to cast to.
-    casting: str {'no', 'safe', 'same_kind', 'unsafe'}, optional
+    casting: str {'no', 'safe', 'same_kind', 'unsafe', "intuitive"}, optional
```
ClaudiaComito, line 455: double vs. single inverted commas

coquelin77:
>     * Changelog needs an update
> 
> 
> Claudia's comment:
> 
>     * Update to the docstring with double quotation marks

docstrings have double quotation marks now, changelog is updated

Issue #563, opened 2020-05-03
Implement distributed topk()


Inzlinger:
https://pytorch.org/docs/stable/torch.html?highlight=topk#torch.topk
@Inzlinger

ClaudiaComito:
We should call this argpartition(), correct?
https://numpy.org/doc/1.18/reference/generated/numpy.argpartition.html

Indeed it would be great to have this.

Markus-Goetz:
This is actually slightly different from argpartition due to better suitability to GPUs. Even though it is not in numpy, you can find it in Torch. I would want to make a small exception here in favor of topk. What do you think @ClaudiaComito

ClaudiaComito:
> This is actually slightly different from argpartition due to better suitability to GPUs. Even though it is not in numpy, you can find it in Torch. I would want to make a small exception here in favor of topk. What do you think @ClaudiaComito

I'm all for it @Markus-Goetz . They do slightly different things, I only thought of argsort() because that's what I would use to get a topk-type of functionality in numpy. Maybe we can have both eventually.

Better suited for GPUs in which sense?

PR #602, opened 2020-06-23
Features/563 topk

Inzlinger:
## Description

Issue/s resolved: #563 

## Due Diligence

- [ x] All split configurations tested
- [ x] Multiple dtypes tested in relevant functions
- [ x] Documentation updated (if needed)

no


coquelin77:
The extra dev dependency pulls in additional tools to support the enforcement of coding conventions (Black) and to support a pre-commit hook to do the same. In order to fully use this framework, please also install the pre-commit hook with

    $ pre-commit install

In order to check compliance of your changes before even trying to commit anything, you can run

    $ pre-commit run --all-files

this should make it so your code is formatted properly on every commit

heat/core/manipulations.py
```diff
@@ -1687,3 +1690,180 @@ def vstack(tup):
...
+    return final_array, final_indices
+
+
+def mpi_topk(a, b, mpi_type):
+    # Parse Buffer
+    a_parsed = torch.from_numpy(np.frombuffer(a, dtype=np.float64))
+    b_parsed = torch.from_numpy(np.frombuffer(b, dtype=np.float64))
+
+    # Collect metadata from Buffer
+    k = int(a_parsed[0].item())
```
Markus-Goetz, line 1835: parsing necessary? can the values be left as float64 instead?

Inzlinger: No, torch.topk only accepts int/bool values. Throws TypError otherwise.

Issue #89, opened 2018-11-28
Add custom __str__ and __repr__ functions for tensors


Markus-Goetz:
Currently, we are using the print functions of pytorch to print tensors to standard out.

Implement a custom version that deals with distributed tensors

ClaudiaComito:
I'm giving up this Issue as I'd rather focus on the higher-level algorithms. @simon-schmitz would you like to take it on?

simon-schmitz:
I can take it on @ClaudiaComito. Is there already some code and/or more information on this issue?

Markus-Goetz:
Bump. What is the status on this?

simon-schmitz:
Unfortunately I won't get time to work on this anytime soon. So far I don't have decent code for this. If the issue is urgent I'd suggest to assign it to somebody else. Otherwise I could tackle this again at some point in the next month.

PR #614, opened 2020-06-30
Features/89 str repr

Markus-Goetz:
## Description

Added the capability to print out DNDarrays

Issue/s resolved: #89 

## Changes proposed:
- Added a __str__ function to the DNDarray class
- Got rid of the __repr__ as it is only a fallback when __str__ is not implemented
- Introduced printing module containing the 
-

## Type of change
- Breaking change - requires all documentation to be updated with respect to new printing format in the ``Examples`` sections.
- Documentation update

## Due Diligence

no


heat/core/dndarray.py
```diff
@@ -3348,10 +3350,11 @@ def std(self, axis=None, ddof=0, **kwargs):
...
 
-    def __str__(self, *args):
-        # TODO: document me
-        # TODO: generate none-PyTorch str
-        return self.__array.__str__(*args)
+    def __str__(self) -> str:
+        """
+        Computes a string representation of the passed DNDarray.
+        """
+        return printing.__str__(self)
```
mtar, line 3357: Should the output resemble torch or numpy? In numpy repr() and str() are giving a slightly different output.

Markus-Goetz: Yes, I am aware of this. I was wondering also on how to proceed with this and whether to chose NumPy's behaviour (i.e. different) or Torch's (same). I think there is no discussion for ``__repr__()`` as it is supposed to 

> Return a string containing a printable representation of an object. For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to ``eval()``

according to the Python documentation. For ``__str__()``, though, the Python documentation says

> ``str(object)`` returns ``object.__str__()``, which is the “informal” or nicely printable string representation of object. 

In my opinion this also entail the extra meta-information of split-axis and device as they are integral for understanding everything about the ``DNDarray's`` state.

heat/core/printing.py
```diff
@@ -0,0 +1,175 @@
...
+
+__all__ = ["get_printoptions", "set_printoptions"]
+
+
+# set the default printing width to a 120
+_DEFAULT_LINEWIDTH = 120
+torch.set_printoptions(profile="default", linewidth=_DEFAULT_LINEWIDTH)
+
+
+def get_printoptions() -> Dict:
```
mtar, line 15: Bracket with types is missing after Dict

heat/core/printing.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def set_printoptions(
+    precision: int = None,
+    threshold: int = None,
+    edgeitems: int = None,
+    linewidth: int = None,
+    profile: int = None,
+    sci_mode: int = None,
+):
```
mtar, line 29: missing return value

heat/core/printing.py
```diff
@@ -0,0 +1,175 @@
...
+    tensor_string = _tensor_str(dndarray, __INDENT + 1)
+    if dndarray.comm.rank != 0:
+        return ""
+
+    return "{}({}, dtype=ht.{}, device={}, split={})".format(
+        __PREFIX, tensor_string, dndarray.dtype.__name__, dndarray.device, dndarray.split
+    )
+
+
+def _torch_data(dndarray, summarize: bool) -> torch.Tensor:
```
mtar, line 83: missing parameter Type for dndarray

heat/core/printing.py
```diff
@@ -0,0 +1,175 @@
...
+                buffer.seek(0)
+                received[i] = torch.load(buffer)
+
+            # stack them along the split axis
+            data = torch.cat(received, dim=dndarray.split)
+
+    return data
+
+
+def _tensor_str(dndarray, indent: int) -> str:
```
mtar, line 154: dndarray type

Markus-Goetz: See above

heat/core/printing.py
Markus-Goetz: I can only provide these, as soon as we have the aliasing in place. Otherwise I have circular imports

heat/core/printing.py
Markus-Goetz: There is no return value on this function

mtar: The python documentation uses `-> None`

Markus-Goetz: Fixed

heat/core/printing.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if dndarray.comm.rank == 0:
+            # deserialize the buffers
+            for i, ele in enumerate(received):
+                buffer.seek(0)
+                buffer.write(ele)
+                buffer.seek(0)
+                received[i] = torch.load(buffer)
+
+            # stack them along the split axis
```
ClaudiaComito, line 148: concatenate? stack creates a new axis

Markus-Goetz:
I had the same issue during testing. It is actually not an issue with the printing functionality itself, but rather with ``resplit_`` and the usage of ``alltoallv``, see also:

> ``[1,0]<stderr>:  File "/Users/c.comito/HAF/heat/heat/core/dndarray.py", line 2692, in resplit_``

This is a known issue with alltoallv. I dont think I should be fixing this in this PR. Any suggestions on how to proceed?

ClaudiaComito:
> I had the same issue during testing. It is actually not an issue with the printing functionality itself, but rather with `resplit_` and the usage of `alltoallv`, see also:
> 
> > `[1,0]<stderr>:  File "/Users/c.comito/HAF/heat/heat/core/dndarray.py", line 2692, in resplit_`
> 
> This is a known issue with alltoallv. I dont think I should be fixing this in this PR. Any suggestions on how to proceed?

I figured, I'm a bit puzzled because I thought the current version of resplit is bypassing Alltoallv.

And actually, the error message above comes from resplit_(None). It's Allgatherv, not Alltoallv. Why doesn't this work? We're using it all over the library.

(I can see that my 3rd example is an Alltoallv problem though)

mtar:
Can it be a problem with __get_item__? The shape of a[0] above returns (0,2) and (0,)

coquelin77:
> Can it be a problem with **get_item**? The shape of a[0] above returns (2,2) and (0,).

which shape are you talking about?

```python
    a = ht.arange(3 * 4, split=0).reshape((3, 2, 2), axis=1)
    b = a[0]
    print(b.shape)
    print(b.lshape)
    print(b)
```
```
[0] (2, 2)
[0] (1, 2)
[0] DNDarray([[0, 2],
[0]           [1, 3]], dtype=ht.int32, device=cpu:0, split=1)
[1] (2, 2)
[1] (1, 2)
[1] 
```

mtar:
where the split axis is zero

coquelin77:
> where the split axis is zero

there are no elements in the tensor on that process. `lshape` returns the size of the local `torch.Tensor`. it does not know anything about the global size. this could be the source of the problem for Allgatherv.

coquelin77:
there is something that was omitted there:
```python
    a = ht.arange(3 * 4, split=0).reshape((3, 2, 2), axis=0)
    b = a[0]
    print(b.shape)
    print(b.lshape)
```
```
[0] (2, 2)
[0] (2, 2)
[1] /home/daniel/.git/heat/heat/core/dndarray.py:1583: ResourceWarning: This process (rank: 1) is without data after slicing, running the .balance_() function is recommended
[1]   ResourceWarning,
[1] (2, 2)
[1] (0,)
```
here is a working script:
```python
    a = ht.arange(3 * 4, split=0).reshape((3, 2, 2), axis=0)
    b = a[0]
    b.balance_()
    print(b.shape)
    print(b.lshape)
    print(b)
````
```
[1] /home/daniel/.git/heat/heat/core/dndarray.py:1583: ResourceWarning: This process (rank: 1) is without data after slicing, running the .balance_() function is recommended
[1]   ResourceWarning,
[0] (2, 2)
[0] (1, 2)
[0] DNDarray([[0, 1],
[0]           [2, 3]], dtype=ht.int32, device=cpu:0, split=0)
[1] (2, 2)
[1] (1, 2)
[1] 
```
the balance is key there. that is why it was failing

ClaudiaComito:
> 
> the balance is key there. that is why it was failing

I'm not amused, but willing to wait until the first users complain about this.

Issue #584, opened 2020-06-05
Feature: DataParallel Class


coquelin77:
use PyTorch nn forwards and backwards functions for the basics of local operations for data parallel nns

Issue #606, opened 2020-06-29
DPNN: dataloader pt2 shuffle


coquelin77:
implement a shuffling routine into the dataloader to shuffle the data between processes

coquelin77:
#660

Issue #605, opened 2020-06-29
DPNN: dataloader


coquelin77:
Create a HeAT class similar to the PyTorch DataLoader for feeding data into the network. Multiple ways to do this. One is to take one HeAT tensor and spit out torch Tensors to be fed into the network

coquelin77:
#660 closed

Issue #604, opened 2020-06-29
DPNN: comm hooks for backwards during the step


coquelin77:
add specialty backwards comm hooks for communicating data between processes. This is intended to receive the data to update the gradients from the last backwards pass. 

The idea is that when `step()` is called it will send the gradients to all of the processes and then the backwards hooks will take care of sending and receiving the data to be used to do a non-blocking update of the parameters

coquelin77:
closed see #660

PR #625, opened 2020-07-10
Features/584 data parallel pr1

coquelin77:
## Description

Implementation of Data Parallel Neural Networks. Documentation about how to use in still being written. This is a first iteration and there will be changes to this which are being working on. Some of these changes to come are a non-blocking data shuffle, partial data loading (i.e. batched data loading), and others! 

Issue/s resolved: #584 #585 #603 #604 #605 #606 

## Changes proposed:
- Implement `DataParallel` class
- Implement `DataLoader` class
- Implement `Dataset` class (template)
- call through to `torch.nn` modules and `torch.nn.functional` routines
- Added MNIST example as a working data parallel NN to examples/nn/mnist
- `MNISTDataset` added for loading data and preparing for the network architecture 
- moved datasets into heat/utils/data/datasets

## Type of change
- Documentation update

## Due Diligence

- [ ] ~~All split configurations tested~~

no


coquelin77:
Coverage on this branch is bad because of the MNIST example. Since this will _greatly_ extend the time required for unit tests, this is left out, this is the primary hangup there.

shout out to @lehr-fa for the great work on the updates of the optimizers!

coquelin77:
small hiccup: nonblocking parameter updates with a scheduler is not implemented yet. it is currently being worked on. MNIST example can only be run with blocking updates at the moment

coquelin77:
> small hiccup: nonblocking parameter updates with a scheduler is not implemented yet. it is currently being worked on. MNIST example can only be run with blocking updates at the moment

so this is fixed now. The MNIST example is up and running properly and should do as it is supposed to. The scheduler works locally at the end of each epoch and the optimizer synchronizes the parameters after each batch. Thank you for your patience

coquelin77:
the coverage in this branch is low because of the un-testable code in `heat/nn/data_parallel.py`. These lines are either in the hooks, or conditions which occur only seldomly in the inner workings of networks. It is not reasonable to test these with unit tests and to reach these points requires a full network architecture. within this files there are 19 lines which fit this category. with these lines the coverage would increase to 98.11 % for the diff.

heat/utils/data/mnist.py
```diff
@@ -0,0 +1,126 @@
...
+
+class MNISTDataset(datasets.MNIST):
+    """
+    Dataset wrapper for :class:`torchvision.datasets.MNIST`. This implements all of the required functions mentioned in
+    :class:`heat.utils.data.Dataset`. The ``__getitem__`` and ``__len__`` functions are inherited from
+    :class:`torchvision.datasets.MNIST`.
+
+    Parameters
+    ----------
+    root : str
```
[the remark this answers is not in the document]
coquelin77: this is how torch does it. essentially, the MNIST data is downloaded to a directory if it doesnt exist there already. then the data is loaded from the folder. since the data is small, it can be loaded fully into memory before being split. but this is a special case for this specific dataset. For larger datasets, a Dataset class needs to be defined.

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+        if list(module.parameters()) != optimizer.param_groups[0]["params"]:
+            raise ValueError("given module and optimizer don't share same parameters.")
+        else:
+            # take reference of optimizer's params
+            self._params_ref = optimizer.param_groups[0]["params"]
+
+        # get parameter indexing and slices
+        start_idx = 0
+        layer_name_prev = None
+        for idx, (name, param) in enumerate(module.named_parameters()):
```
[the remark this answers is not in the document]
coquelin77: i believe that all of the layers must have names. this is done automatically in the background by torch. but @lehr-fa would know more

heat/utils/data/mnist.py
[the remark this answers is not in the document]
rainman110: Then we should stick to it

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+        self._first_iter = True
+
+    def __iter__(self) -> Iterator:
+        # need a new iterator for each epoch
+        if self._first_iter:
+            self._first_iter = False
+        else:
+            # shuffle after the first epoch but before the iterator is generated
+            self.shuffle()
+        return self.DataLoader.__iter__()
```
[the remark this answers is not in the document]
coquelin77: so this is where things get complicated. from an outside perspective this is like calling `ht.DataLoader.DataLoader`. The `DataLoader` attribute of the heat DataLoader class is a torch DataLoader. It gets the local pytorch tensor and passes it to the network. this is done via the torch DataLoader's iterator. So the iterator function of the `ht.DataLoader` will shuffle the data between processes (after the first epoch) before returning a new iterator to iterate over the local data on the process

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def __len__(self) -> int:
+        return self.data.shape[0]
+
+    def shuffle(self):
+        """
+        Send half of the local data to the process ``self.comm.rank + 1`` if available, else wrap around. After
+        receiving the new data, shuffle the local tensor.
+        """
+        lc = self.htdata._DNDarray__array
```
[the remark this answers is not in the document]
coquelin77: i didnt want to overwhelm, shuffle is already abstracted for the next PR

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+    def __init__(self, array, transform: Callable = None):
+        self.htdata = array
+        self.comm = array.comm
+        # create a slice to create a uniform amount of data on each process
+        min_data_split = array.gshape[array.split] // array.comm.size
+        self.lcl_half = min_data_split // 2
+        arb_slice = [slice(None)] * array.ndim
+        arb_slice[array.split] = slice(min_data_split)
+        self._cut_slice = tuple(arb_slice)
+        self.data = array._DNDarray__array[self._cut_slice]
```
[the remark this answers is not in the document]
coquelin77: not necessarily. This is not a complex operation and it only happens on each process. 

What is happening here is the last element is cut off from `self.htdata` if the process has more elements than the others. After this is done, `self.data` is a local torch tensor which is the same size on every process. this ensures that the batches used on every process are the same size. A torch tensor is used here because it is what is needed by the torch DataLoader which is used to feed data into the network and to automatically determine batch sizes

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+        the local module
+    comm : heat.MPICommunicator
+        HeAT communicator to use
+    optimizer : torch.optim.Optimizer
+        Optimizer used for parameter updates.
+    blocking : bool (optional)
+        Flag for blocking synchronization. If not given, synchronization is blocking by default.
+    """
+
+    def __init__(
```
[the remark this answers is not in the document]
coquelin77: im not sure how registering the hooks would work from the optimizer for this. @lehr-fa would know more (he is otherwise incapacitated today)

heat/cluster/tests/test_kmeans.py
```diff
@@ -28,7 +28,7 @@ def test_get_and_set_params(self):
     def test_fit_iris_unsplit(self):
         for split in [None, 0]:
             # get some test data
-            iris = ht.load("heat/datasets/data/iris.csv", sep=";", split=split)
+            iris = ht.load("heat/utils/data/datasets/iris.csv", sep=";", split=split)
```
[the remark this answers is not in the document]
coquelin77: this was done to put all of the data-related things in one place. i thought that utils made sense because the data shipped with heat is primarily used for proof-of-concept applications as well as examples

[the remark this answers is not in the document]
rainman110: Imho, this should be discussed in a larger round as it also affects the existing folder structure.

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+        self._param_slices = dict()
+        # pytorch internal parameter indexing
+        self._param_indices = dict()
+        # reference of optimizer's params
+        self._params_ref = None
+        # flag indicating if optimizer should take a step during next iteration (only relevant for non-blocking)
+        self._update_next = False
+
+        # check if optimizer matches module
+        if list(module.parameters()) != optimizer.param_groups[0]["params"]:
```
krajsek, line 83: Sometimes I like to optimize only parts of the parameters, which seems to not possible here ?

coquelin77: im not sure. do you have an example of how you would do that in a standard torch network? im unfamiliar with how the code would look for that

rainman110: @krajsek Could you please provide a small example of how you update only a part of the parameters?

Are you using it for transfer learning or what is the intention?

heat/nn/data_parallel.py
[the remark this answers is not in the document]
krajsek: I don't really want to do this now. It was rather meant to give the possibility to pass a subset. You can disable parameters with, requires_grad, but it is also allowed to pass only a part of the parameters, simple example: pass  filter(lambda p: p.requires_grad, model.parameters()) instead of model.parameters() after some parameters are marked inactive

heat/nn/data_parallel.py
krajsek: I don't really want to do this now. It was rather meant to give the possibility to pass a subset. You can disable parameters with, requires_grad, but it is also allowed to pass only a part of the parameters, simple example: pass filter(lambda p: p.requires_grad, model.parameters()) instead of model.parameters() after some parameters are marked inactive

heat/nn/data_parallel.py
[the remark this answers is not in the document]
krajsek: In case of  the Sequential class, the layers are numbered. In case of the Module class, layers have the names you assign to them in the init() method. e.g. 
self.fc = nn.Linear(D, H)
names will be:
fc.weight
fc.bias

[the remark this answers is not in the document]
lehr-fa: I needed an unique attribute to distinguish between layers and get a clear assignment. Naming works as @krajsek explained. In both cases, there is no possibility that two different layers have the same name. But I see a little mistake here, occurring when one creates nested networks: The string returned by `named_parameters()` may only be split at the rightmost occurrence of dot delimiter, currently it is split at the leftmost occurrence.

heat/nn/data_parallel.py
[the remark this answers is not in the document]
lehr-fa: We have to distinguish between two steps here: backward and optimization. Data parallel backward only accumulates gradients, which happens in the hook functions. Therefore, hooks should be part of the `DataParallel` class, as they currently are. 
I agree that your design is closer to the pytorch design. Since invoking `Optimizer.step()` and therewith performing optimization doesn't necessarily have to part of `DataParallel` class, it'll be easy to source out into a separate `ParallelOptimizer` class.

heat/nn/data_parallel.py
lehr-fa: For blocking communication, this is trivial to be implemented. For non-blocking, it'll be more complicated. But I’ll see what I can do.

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+        """
+        Force optimizer to update model parameters. For blocking, optimizer immediately updates parameters. For
+        non-blocking, optimizer will update parameters during next forward.
+        """
+        if self.blocking:
+            self.optimizer.step()
+        else:
+            self._update_next = True
+
+    def _async_update(self, param_slice: slice = None, layer_names: List[str] = None):
```
krajsek, line 165: The term "asynchronious"  needs an explanation, otherwise it may lead to confusion. In the literature, asynchronous SGD schemes (e.g.  J. Tsitsiklis, D. Bertsekas, and M. Athans. 1986.  
 Distributed asynchronous deterministic and stochastic gradient optimization algorithms.IEEE Trans.Automat. Control31, 9 (1986), 803–812.)  are methods in which the gradients from calculations of different time steps are averaged. The term has a different meaning here.

heat/nn/data_parallel.py
krajsek: Could you not just gather all parameters in the optimizer at the beginning and then just work with them?

heat/nn/data_parallel.py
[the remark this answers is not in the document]
rainman110: In my proposal the dp_optimizer has access to the parallel model and can modify things there if required.

But as I said, this is a matter of taste and if things are too complicated to change or do not make sense, just let the design as it currently is.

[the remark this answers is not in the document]
lehr-fa: Actually it's not too complicated. The problem is that the relationship 'dp_optimizer knows dp_model' isn't sufficient or even necessary. We need the relationship 'dp_model knows dp_optimizer' though. The question is, how to pass dp_optimizer to dp_model in an optimal way (from design perspective)? More than ever, if there might be several dp_optimizers to be assigned to a single dp_model. I suggest to implement a simple `add_optimizers()` function, which takes either one or multiple dp_optimizers to be appended to a list attribute of dp_model.

heat/nn/data_parallel.py
coquelin77: ive changed it to `_iparam_update` to clear this up. thanks for pointing this out

heat/nn/data_parallel.py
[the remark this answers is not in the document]
coquelin77: this is being worked on at the moment.

heat/nn/data_parallel.py
lehr-fa: Maybe I didn't understand your suggestion correctly, but I don't see how this could help. The problem with multiple optimizers and non-blocking parameter updates is that parameter updates are performed layer-wise currently. Since one optimizer could possibly deal with only a part of a layer's parameters and another optimizer could deal with another part, parameter updates would have to be performed parameter-wise to cover the general case. And this would increase not only code complexity but also computation time, because `optimizer.step()` wouldn't have to be invoked per layer as it is now, but per parameter. If we consider linear and conv layers, which contain two parameters each (bias and weight), this would result in doubling computation time caused by network's optimization step.

krajsek: What do you mean with multiple optimizers? Currently you can only pass one optimizer.

lehr-fa: Exactly, but we work on allowing to pass a bunch of optimizers. If one decides to use more than one though, communication during parameter updates will be blocking. This is the trade-off/compromise we have to accept.

krajsek: o.k. that is new for me. Can you give me an example where you need multiple optimizers?

lehr-fa: E.g. https://discuss.pytorch.org/t/different-optimisers-in-different-layers/72484
May be a corner case, but still.

Passing a single optimizer which deals only with a subset of parameters will be possible as well (this was your original request) and maybe will also work for non-blocking, but as I said, this is tricky. For blocking, it's trivial because optimizer.step() is invoked once to update all parameters being assigned. I don't have to care if it's actually a subset of parameters of if it's the total set of parameters. 
But for non-blocking, it's more complex. Here, parameters are updated layer-wise, so I need a mapping: layer -> parameter. This mapping isn't provided by parameters attribute of optimizer, therefore I have to do it on my own. And this becomes complex, if set of parameters being held by optimizer isn't equal to model's set of parameters.
Sorry for probably misleading in my former answer, I hope the actual problem of your original request is clear now.

krajsek: o.k., now it is clear for me, thanks. Maybe we can consider the  feature of having only a subset of parameters for later implementations

examples/nn/mnist.py
```diff
@@ -0,0 +1,170 @@
...
+sys.path.append("../../")
+import heat as ht
+import heat.nn.functional as F
+import heat.optim as optim
+from heat.optim.lr_scheduler import StepLR
+from heat.utils import vision_transforms
+from heat.utils.data.mnist import MNISTDataset
+
+"""
+
```
Cdebus, line 15: superfluous line

examples/nn/mnist.py
```diff
@@ -0,0 +1,170 @@
...
+This file is an example script for how to use the HeAT DataParallel class to train a network on the MNIST dataset.
+To run this file execute:
+mpirun -np N python mnist.py
+where N is the number of processes, in the examples/nn/ directory
+"""
+
+
+class Net(ht.nn.Module):
+    def __init__(self):
```
Cdebus, line 24: Add class doc ^^ (You could modify the docstring above a bit to fit here)

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+from heat.core.communication import MPI
+
+__all__ = ["DataParallel"]
+
+
+class DataParallel(tnn.Module):
+    """
+    Implements data parallelism across multiple processes. This means that the same model will be run locally
+    on each process. Creation of the model parallels to PyTorch, the only changes are using HeAT layers (ht.nn.layer)
+    in the initialization of the network. I.E. (example code contains)
```
Cdebus, line 20: ht.nn.layer as cross-reference

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+                self.net1 = ht.nn.Linear(10, 10)
+                self.relu = ht.nn.ReLU()
+                self.net2 = ht.nn.Linear(10, 5)
+
+            def forward(self, x):
+                return self.net2(self.relu(self.net1(x)))
+
+        ht_model = ht.nn.DataParallel(TestingModel(), comm, optimizer)
+
+    and a requirement of giving a HeAT communicator (``comm``) as well as a Torch optimizer (``optimizer``). For the
```
Cdebus, line 35: Cross-reference for heat Communicator

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+    It is highly recommended that a HeAT DataLoader is used, see :func:`..utils.data.datatools.DataLoader`. The
+    default communications scheme for this is blocking. The blocking scheme will average the model parameters during
+    the backwards step, synchronizing them before the next model iteration.
+
+    To use the non-blocking communications for parameter updates, negate the optional flag ``blocking``.
+
+    Attributes
+    ----------
+    module : torch.nn.Module
+        the local module
```
Cdebus, line 47: The

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+    comm : heat.MPICommunicator
```
Cdebus, line 48: remove "heat.", add direct cross-ref (also, does it necessarily have to be the MPICommunicator, or a generalized ht communicator ? )

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+        HeAT communicator to use
+    optimizer : torch.optim.Optimizer
+        Optimizer used for parameter updates.
+    blocking : bool (optional)
```
Cdebus, line 52: bool, optional

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+                param.register_hook(self._nonblocking_hook(layer_name, name))
+        self._param_slices[layer_name_prev] = slice(start_idx, len(self._param_indices))
+
+    def __setattr__(self, name, value):
+        # auto-detect end of epoch's training phase and finalize wait handles (only relevant for non-blocking)
+        if name == "training" and not value and not self.blocking:
+            self._iparam_update()
+        super(DataParallel, self).__setattr__(name, value)
+
+    def forward(self, *inputs: tuple, **kwargs: dict) -> torch.Tensor:
```
Cdebus, line 115: Docstring missing

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+        else:
+            self._update_next = True
+
+    def _iparam_update(self, param_slice: slice = None, layer_names: List[str] = None):
+        """
+        Update parameters asynchronously via wait handles.
+
+        Parameters
+        ----------
+        param_slice : slice (optional)
```
Cdebus, line 171: slice, optional

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+            Slice object for creating a view onto optimizer's params list. If None, the whole params list is taken.
+        layer_names : list(str) (optional)
```
Cdebus, line 173: use Typehints, write ",optional"

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+            List of layer names, whose parameters are to be updated. Has to match param_slice. If None, all layers are
```
Cdebus, line 174: param_slice and None in code style ``None``

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+            self._update_next = True
+
+    def _iparam_update(self, param_slice: slice = None, layer_names: List[str] = None):
+        """
+        Update parameters asynchronously via wait handles.
+
+        Parameters
+        ----------
+        param_slice : slice (optional)
+            Slice object for creating a view onto optimizer's params list. If None, the whole params list is taken.
```
Cdebus, line 172: Code style for None

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+        Add a blocking hook to the PyTorch DAG for all of the backwards calls.
+
+        Parameters
+        ----------
+        grad_loc : torch.Tensor
+            the local gradient
+
+        References
+        ----------
+        - (cf. https://pytorch.org/docs/stable/tensors.html#torch.Tensor.register_hook).
```
Cdebus, line 219: References with [1]

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+            self.optimizer.step()
+
+    def _blocking_hook(self, grad_loc: torch.Tensor) -> torch.Tensor:
+        """
+        Add a blocking hook to the PyTorch DAG for all of the backwards calls.
+
+        Parameters
+        ----------
+        grad_loc : torch.Tensor
+            the local gradient
```
Cdebus, line 215: The

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+        return grad_loc
+
+    def _nonblocking_hook(self, layer_name: str, param_name: str) -> Callable:
+        """
+        Add a nonblocking hook to send and receive the averaged parameters after the backwards step
+
+        Parameters
+        ----------
+        layer_name : str
+            name of the layer
```
Cdebus, line 234: Name

heat/nn/data_parallel.py
```diff
@@ -0,0 +1,275 @@
...
+        param_name : str
+            name of the parameter
```
Cdebus, line 236: Name

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+from typing import Callable, Iterator, Union
+
+from heat.core import dndarray
+
+__all__ = ["DataLoader", "Dataset"]
+
+
+class DataLoader:
+    """
+    Data Loader. The combines either a ``DNDarray`` or a torch ``Dataset`` with a sampler. It will provide an iterable
```
Cdebus, line 16: Use Cross-reference for DNDarray here

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+    over the local dataset and it will have a ``shuffle()`` function which calls the ``shuffle()`` function of the
+    given Dataset. If a HeAT DNDarray is given a general Dataset will be created.
```
Cdebus, line 18: Codestyle for DNDarray and Dataset

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
+import h5py
+import numpy as np
+import base64
+import os
+import torch
+from torch.utils import data as torch_data
+from typing import Callable, Iterator, Union
+
+from heat.core import dndarray
```
Cdebus, line 9: use dndarray.DNDarray for proper referencing

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+    Arguments:
+        data : Dataset or DNDarray
+            dataset from which to load the data.
+        batch_size : int, optional
+            how many samples per batch to load (default: ``1``).
+        num_workers : int, optional
+            how many subprocesses to use for data loading. ``0`` means that the data will be loaded in the main process.
+            (default: ``0``)
+        collate_fn : callable, optional
+            merges a list of samples to form a mini-batch of Tensor(s).  Used when using batched loading from a
```
Cdebus, line 32: Torch Tensors or DNDarrays? If DNDarray, please use that or array to avoid confusion

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+            map-style dataset.
+        pin_memory : bool, optional
+            If ``True``, the data loader will copy Tensors into CUDA pinned memory before returning them.  If your
```
Cdebus, line 35: Tensor or DNDarray?

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+            the batch size. If ``False`` and the size of dataset is not divisible by the batch size, then
+            the last batch will be smaller. (default: ``False``)
+        timeout : int or float, optional
+            if positive, the timeout value for collecting a batch from workers. Should always be non-negative.
+            (default: ``0``)
+        worker_init_fn : callable, optional
+            If not ``None``, this will be called on each worker subprocess with the worker id
+            (an int in ``[0, num_workers - 1]``) as input, after seeding and before data loading. (default: ``None``)
+        lcl_dataset : torch.Dataset
+            a PyTorch dataset from which the data will be returned by the created iterator
```
Cdebus, line 49: Torch instead of PyTorch

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+        transform : Callable
+            transform to be given to Dataset creation if a Dataset is created
```
Cdebus, line 51: Crossref to Dataset

heat/utils/data/datatools.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+class Dataset(torch_data.Dataset):
+    """
+    An abstract class representing a given dataset. This inherits from torch.utils.data.Dataset.
+
+    This class is a general example for what should be done to create a Dataset. When creating a dataset all of the
+    standard attributes should be set, the ``__getitem__``, ``__len__``, and ``shuffle`` functions must be defined.
+
+        - ``__getitem__`` : how an item is given to the network
```
Cdebus, line 132: use \n at end of line for proper rendering

heat/utils/data/mnist.py
```diff
@@ -0,0 +1,126 @@
...
+        directory containing the MNIST dataset
+    train : bool, optional
+        if the data is the training dataset or not, default is True
```
Cdebus, line 22: CodeStyle for True

heat/utils/data/mnist.py
```diff
@@ -0,0 +1,126 @@
...
+    transform : Callable, optional
+        transform to be applied to the data dataset in the ``__getitem__`` function, default is None
```
Cdebus, line 24: CodeStyle for None

heat/utils/data/mnist.py
```diff
@@ -0,0 +1,126 @@
...
+    target_transform : Callable, optional
+        transform to be applied to the target dataset in the ``__getitem__`` function, default is None
+    download : bool, optional
+        if the data does not exist in the directory, download it if True (default)
+    split : int, optional
+        on which access to split the data when it is loaded into a DNDarray
```
Cdebus, line 30: CodeStyle for DNDarray

coquelin77:
PSA: thank you for all of the comments. The next iteration of this is nearly done already with a partial data loader for datasets to be loaded in the background. It is unclear if it will be added to this PR or if it will be a separate PR. but the changes mentioned here will be implemented before either is implemented into the master

mtar:
Can one of the admins verify this patch?

Cdebus:
@mtar See comment above, I think we are waiting for the next iteration/update. And it is unclear whether it will be in this branch or a new one

examples/nn/mnist.py
coquelin77: this is just an example network, it is excluded in the docs. The text at the top is some info about the file in general for a user and is not related to the network. It will likely be added to a readme for the nn examples eventually

heat/nn/data_parallel.py
coquelin77: there isnt something to reference here because the `ht.nn.layer` calls through to torch

heat/nn/data_parallel.py
coquelin77: the MPICommunicator is the general heat communicator

coquelin77:
closed as stale. this is also all covered by #660

Issue #568, opened 2020-05-19
get_halo behaviour if no data on process


ClaudiaComito:
**Description**
Currently, `dndarray.get_halo` raises a ValueError if the size of the tensor along the split axis is smaller than the number of MPI ranks (i.e., empty tensors on some ranks).

Couldn't `get_halo` treat the last non-empty rank as the last rank, and leave the others alone (empty), but still return `array_with_halos` on the populated ranks? 

**To Reproduce**
See errors in test_statistics.test_percentile() here
https://travis-ci.com/github/helmholtz-analytics/heat/builds/166837904

where data.gshape was (6, 10, 10) on 7 nodes.

(I've currently changed the data size to (10, 10, 10) in order to be able to build the PR)

**Expected behavior**
Ideally the program shouldn't crash if  `x.gshape[x.split] < x.comm.size`

**Version Info**
Branch features/22-percentile

lucaspataro:
Hi @ClaudiaComito,

sorry for my late reply. 
I agree with you, such a scenario should definitely be possible. I will have a look next week how this can be realised.

PR #631, opened 2020-07-13
Bug/568 get halo behaviour if no data on process

lucaspataro:
## Description
Issue/s resolved: #568

Empty ranks at the end of the communicator are now ignored.

## Type of change

## Due Diligence

no


heat/core/dndarray.py
```diff
@@ -324,11 +324,25 @@ def get_halo(self, halo_size):
...
-            if halo_size > min_chunksize:
+            # gather lshapes
+            lshape_map = self.create_lshape_map()
+            rank = self.comm.rank
+            size = self.comm.size
+            next_rank = rank + 1
+            prev_rank = rank - 1
+            last_rank = size - 1
+
+            # if local shape is zero and its the last process or the next one is also zero we can exit here
```
ClaudiaComito, line 335: You want to be checking if a DNDarray is balanced here. Otherwise you could hypothetically have 2 empty processes and then some data on one of the next ones. 
On the other hand, if a DNDarray is balanced and lshape[split] on rank is 0, you don't need to check the next rank.

coquelin77:
bump, requested changes and conflict resolution

mtar:

heat/core/dndarray.py
lucaspataro: What is the desired behavior if there are empty ranks in the middle? Preventing the halo all together, or skipping the empty ranks and using the next/prev non-empty rank?

ClaudiaComito: Hi @lucaspataro,  that's an interesting question. I think  so far we're assuming that operations are performed on balanced DNDarrays, in that case the empty ranks will always be at the end. I would say that for now an empty rank "in the middle" is an edge case we don't really need to worry about. What do you think, @krajsek ?

lucaspataro: Hi @ClaudiaComito, i added a RuntimeError if tensor is not balanced and remove the check for the next rank like you suggested.

ClaudiaComito (approved):
Great, thanks a lot!

PR #645, opened 2020-08-03
Fixed install command in readme

rainman110:
## Description

The installation command contained spaces, which at least my pip didn't like.

## Type of change

- Documentation update

no


coquelin77:
do we need to remove these spaces? does that do something different? i thought that they were ignored by the interpreter

rainman110:
Using spaces, I had trouble with the installation. So yes, it seems to be required.

mtar:
I can confirm that. The spaces are problematic on bash.

coquelin77:
thanks for the confirmation @mtar ill update this branch and merge it today

coquelin77 (approved):

Issue #650, opened 2020-08-06
Wrong use of NotImplemented in Communication


Inzlinger:
**Description**
In class Communication is_distributed, chunk and init return NotImplemented, which is reserved for binary Operations.

**Expected behavior**
raise NotImplementedError

PR #651, opened 2020-08-06
Replaced NotImplemented with NotImplementedError

Inzlinger:
## Description
Replaced return NotImplemented with raise NotImplementedError

Issue/s resolved: #650 

## Type of change

## Due Diligence

no


mtar:
Can one of the admins verify this patch?

coquelin77:
crashing due to torch 1.6.0 updates, patch being implemented currently. there is nothing here which should change with the tests. Once the patch is up, then this will be merged

PR #653, opened 2020-08-07
Updates for pytorch 1.6.0

mtar:
## Description

Update the unittests for argmax and argmin for the changed return behaviour (numpy like) resulting from changes in pytorch 1.6.0 (torch.max, torch.min return first value).
Additionally, mpi_argmax and mpi_argmin sort the inputs by starting index beforehand. Some MPI implementations do not bind the first input to the lower process.

## Changes proposed:
- argmax test update
- argmin test update
- requires torch 1.6.0
- sort arguments in mpi_argmax, mpi_argmin (Fix bug in some MPI implementations)
- updated gpu calls in a few tests (see code for details)
- io tests use the current working directory for the tests instead of a temp directory

## Type of change

## Due Diligence

Unknown


coquelin77:
these tests will not run on GPU. is there a reason that we cant use the torch argmax/argmin for the tests?

coquelin77:
build failing -> due to issues with jenkins
codecov -> most changes are within aspects which are not touched (5 lines) by travis, 4 in MPI operations, 1 is a print change for old torch versions

Issue #657, opened 2020-08-25
matmul fails with ints on GPU


coquelin77:
**To Reproduce**
```python
import heat as ht

a = ht.random.randint(0, 30, size=(30, 30), device="gpu", split=0)
b = ht.random.randint(0, 30, size=(30, 30), device="gpu", split=0)
ht.dot(a, b)
```

result:
```
RuntimeError: "addmm_cuda" not implemented for 'Int'

Traceback (most recent call last):
  File "dot-test.py", line 5, in <module>
    ht.dot(a, b)
  File "/p/home/jusers/coquelin1/hdfml/heat/heat/heat/core/linalg/basics.py", line 56, in dot
    ret = matmul(a, b)
  File "/p/home/jusers/coquelin1/hdfml/heat/heat/heat/core/linalg/basics.py", line 501, in matmul
    c=c._DNDarray__array,
RuntimeError: The following operation failed in the TorchScript interpreter.
Traceback of TorchScript (most recent call last):
  File "/p/home/jusers/coquelin1/hdfml/heat/heat/heat/core/linalg/basics.py", line 803, in __mm_c_block_setter
                    c_start0 = a_start0
                    c_start1 = b_start1
                    c[c_start0 : c_start0 + mB, c_start1 : c_start1 + nB] += a_block @ b_block
                                                                             ~~~~~~~~~~~~~~~~~ <--- HERE
RuntimeError: "addmm_cuda" not implemented for 'Int'
```

**Expected behavior**
Should run without failure

**Version Info**
current master

PR #658, opened 2020-08-25
mm checks and correct int dtypes if A is on the gpu

coquelin77:
## Description

fixes bug by casting tensors to floats before mm operation

Issue/s resolved: #657 

## Changes proposed:
- for `matmul`: cast tensors to floats if the devices are GPUs and the dtypes are ints

## Type of change

## Due Diligence

no

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - [#652](https://github.com/helmholtz-analytics/heat/pull/652) Feature: benchmark scripts and jobscript generation
 - [#653](https://github.com/helmholtz-analytics/heat/pull/653) Printing above threshold gathers the data without a buffer now
 - [#653](https://github.com/helmholtz-analytics/heat/pull/653) Bugfixes: Update unittests argmax & argmin + force index order in mpi_argmax & mpi_argmin. Add device parameter for tensor creation in dndarray.get_halo().
+- [#658](https://github.com/helmholtz-analytics/heat/pull/658) Bugfix: `matmul` on GPU will cast away from `int`s to `float`s for the operation and cast back upon its completion. This may result in numerical inaccuracies for very large `int64` DNDarrays
 - [#659](https://github.com/helmholtz-analytics/heat/pull/659) New feature: `random.permutation` + `random.randperm`
 - [#662](https://github.com/helmholtz-analytics/heat/pull/662) Bugfixes: minimum() and maximum() split semantics, scalar input, different input dtype
 - [#664](https://github.com/helmholtz-analytics/heat/pull/664) New feature / enhancement: `random.random_sample`, `random.random`, `random.sample`, `random.ranf`, `random.random_integer`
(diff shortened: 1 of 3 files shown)


mtar:

mtar:
test this please

bhagemeier:
It works, but the results are now dependent on whether you compute on the GPU or CPU. Do we have other places where we do similar stuff? The actual difference at first sight is the difference in the data type returned (int64 vs. float64). Further down the road, we're losing 11 bits of precision for very large numbers. This may be a corner case, but I wanted to point it out before approving this.

```
>>> a = ht.array([[0, 1], [2, 3]])
>>> b = ht.array([[4, 5], [6, 7]])
>>> ht.dot(a, b)
DNDarray([[ 6,  7],
          [26, 31]], dtype=ht.int64, device=cpu:0, split=None)
>>> a.gpu()
DNDarray([[0, 1],
          [2, 3]], dtype=ht.int64, device=gpu:0, split=None)
>>> b.gpu()
DNDarray([[4, 5],
          [6, 7]], dtype=ht.int64, device=gpu:0, split=None)
>>> ht.dot(a, b)
DNDarray([[ 6.,  7.],
          [26., 31.]], dtype=ht.float64, device=gpu:0, split=None)
```

coquelin77:
as far as i know, this is the only place where we do a specific type change for GPUs. however, the function required for this operation, namely `addmm_cuda` is not implemented so we are forced to work around it. even if we were to do this:
```python
r = a_block @ b_block
c[c_start0 : c_start0 + mB, c_start1 : c_start1 + nB] += r
```
it still fails with the same error, unfortunately I do not see a way around this. we can loop back to it when this function is implemented in torch.

mtar:
The dtype of the returned array changes. As a user, I would expect it to be the same as the input.

coquelin77:
again, the operation which we need is **not implemented** in torch. this **cannot** be done a different way in this specific scenario. It can throw a warning but that seems excessive.

mtar:
You can change the type afterwards

coquelin77:
the precision is already lost even if the type is changed back. we can cast back at the end but it does require yet another check in all 8 return locations

Markus-Goetz:
Even if we do loose precision, we should be consistent on the types. For GPU, cast the output tensor back to int64. The case that you actually do int64 mm on GPU is relatively seldom to being with anyway.

mtar:
rerun tests

coquelin77:
this now casts back to the initial promoted dtype for ints on GPUs

CHANGELOG.md
```diff
@@ -58,10 +59,6 @@
 - [#652](https://github.com/helmholtz-analytics/heat/pull/652) Feature: benchmark scripts and jobscript generation
 - [#653](https://github.com/helmholtz-analytics/heat/pull/653) Printing above threshold gathers the data without a buffer now
 - [#653](https://github.com/helmholtz-analytics/heat/pull/653) Bugfixes: Update unittests argmax & argmin + force index order in mpi_argmax & mpi_argmin. Add device parameter for tensor creation in dndarray.get_halo().
-- [#659](https://github.com/helmholtz-analytics/heat/pull/659) New feature: distributed `random.permutation` + `random.randperm`
-- [#662](https://github.com/helmholtz-analytics/heat/pull/662) Bugfixes: `minimum()` and `maximum()` split semantics, scalar input, different input dtype
-- [#664](https://github.com/helmholtz-analytics/heat/pull/664) New feature / enhancement: distributed `random.random_sample`, `random.random`, `random.sample`, `random.ranf`, `random.random_integer`
-- [#666](https://github.com/helmholtz-analytics/heat/pull/666) New feature: distributed prepend/append for `diff()`.
```
bhagemeier, line 64: These lines should not have been removed from the CHANGELOG. I'll fix that.

coquelin77: oh sorry! i thought they were duplicates. my bad

bhagemeier (dismissed):
Looks good now from my point of view.

bhagemeier (approved):
Test have been added. I have a feeling that codecov is having trouble just now. Therefore, I vote for ignoring the minute (.01%) decrease in coverage.

PR #659, opened 2020-08-26
Features/554 permutation

mtar:
## Description

This Pull Request introduces the functions random.permutation and random.randperm imitating the behaviour of the respective NumPy or PyTorch function.

Issue/s resolved: #554 

## Changes proposed:
- add `random.permutation` [NumPy documentation](https://numpy.org/devdocs/reference/random/generated/numpy.random.permutation.html)
- add `random.randperm` [PyTorch documentation](https://pytorch.org/docs/stable/generated/torch.randperm.html)

## Type of change

New feature

## Due Diligence

no

### DIFF
--- a/heat/core/random.py
+++ b/heat/core/random.py
@@ -276,7 +276,7 @@ def permutation(x):
         return randperm(x)
 
     if not isinstance(x, dndarray.DNDarray):
-        raise TypeError("x must be integer or DNDarray")
+        raise TypeError("x must be int or DNDarray")
 
     # random permutation
     recv = torch.randperm(x.shape[0], device=x.device.torch_device)
@@ -313,7 +313,7 @@ def permutation(x):
     else:
         data = torch.empty_like(x._DNDarray__array)
 
-    return factories.array(data, dtype=x.dtype, is_split=x.split, comm=x.comm)
+    return factories.array(data, dtype=x.dtype, is_split=x.split, device=x.device, comm=x.comm)
 
 
 def rand(*args, dtype=types.float32, split=None, device=None, comm=None):
--- a/heat/core/random.py
+++ b/heat/core/random.py
@@ -543,7 +543,7 @@ def randperm(
 
     return factories.array(perm, dtype=dtype, device=device, split=split, comm=comm)
 
-    
+
 def random_sample(
     shape: Optional[Tuple[int]] = None,
     dtype=types.float32,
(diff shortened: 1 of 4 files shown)


mtar:
run tests

mtar:
rerun tests

mtar:
rerun tests

heat/core/random.py
```diff
@@ -237,6 +239,83 @@ def __kundu_transform(values):
...
+    >>> ht.random.permutation(arr)
+    DNDarray([[3, 4, 5],
+              [6, 7, 8],
+              [0, 1, 2]], dtype=ht.int32, device=cpu:0, split=None)
+    """
+    if isinstance(x, int):
+        return randperm(x)
+
+    if not isinstance(x, dndarray.DNDarray):
+        raise TypeError("x must be integer or DNDarray")
```
coquelin77, line 279: "x must be integer or DNDarray" -> "x must be an integer or DNDarray" or "x must be int or DNDarray"

heat/core/random.py
```diff
@@ -237,6 +239,83 @@ def __kundu_transform(values):
...
+            requests.append(x.comm.Isend(x.lloc[i].clone(), proc_send, tag=tag))
+
+        for req in requests:
+            req.Wait()
+
+        data = torch.stack(buf)
+    else:
+        data = torch.empty_like(x._DNDarray__array)
+
+    return factories.array(data, dtype=x.dtype, is_split=x.split, comm=x.comm)
```
coquelin77, line 316: device from `x` as well

heat/core/tests/test_random.py
```diff
@@ -310,6 +348,25 @@ def test_randn(self):
         self.assertFalse(np.allclose(a, c))
         self.assertFalse(np.allclose(b, c))
 
+    def test_randperm(self):
+        state = torch.random.get_rng_state()
+
+        # results
```
coquelin77, line 354: can you test with float64 as well?

also should the return values should be `int64` correct? Can you test to make sure that they are what is expected?

coquelin77 (changes requested):
These functions are very clean and easy to read. However, these solutions are still memory bound. This may be a problem in the future when a vector of length equal to `gshape[0]` cannot fit into memory. I am okay with adding this for now because I think this case is still a corner case, @Markus-Goetz what do you think?

heat/core/random.py
mtar: fixed

Markus-Goetz:
Let's have a implementation first and the improve on it when necessary.

mtar:
rerun tests

coquelin77 (approved):

Issue #661, opened 2020-09-01
minimum() problems: split semantics, dtype, dealing with scalars


ClaudiaComito:
**Description**
ht.minimum() has several quirks:
- it does not follow split semantics (the outcome is always "gathered" i.e. has split=None)
- returns a (torch) error if DNDarrays aren't of the same dtype, doesn't do type promotion
- cannot deal with scalar input

**To Reproduce**
1. Which module/class/function is affected?
statistics.minimum()
2. What are the circumstances under which the bug appears?
`ht.minimum(x1, x2)` where x1, x2 are distributed, or `x1.dtype not x2.dtype`, or x1, x2 are int or float

**Expected behavior**
- `ht.minimum(x1,x2)` should be distributed like x1, x2; 
- `if x1.dtype not x2.dtype:` type promotion

 

**Version Info**
 0.4.0 current main branch

Issue #379, opened 2019-09-13
Make use of __binary_op for two tensor binary high-level functions


Markus-Goetz:
**Related**
Might require issue #167 to be implemented first and potentially #377.

**Feature functionality**
Make use of the __binary_op for higher-level broadcastable binary operations, such as minimum(), where(), outer(), ... The logic for resplitting DNDarrays should already be present in __binary_op (or should eventually be there). It would be redundant and redundant to have them yet again in each of these high-level calls.

Individual calls might make it necessary to copy some parts of the behaviour. If so, think about how to split up __binary_op so that the code duplication is minimal.

ClaudiaComito:
This issue has been partially addressed (#662).

Breaking up `__binary_op()` in two (separating `__binary_checks()` ?) would indeed be helpful for many operations, so I'm leaving this issue open.

---
Reviewed within #1109

ClaudiaComito:
Adding `logical.isclose` #458 , `logical.allclose` to this list

PR #662, opened 2020-09-01
Bug/minimum split

ClaudiaComito:
## Description

This PR addresses several points within `statistics.minimum()` and `maximum()`

- first off, split semantics. Twofold problem:
  1. split semantics are not observed in the original implementation, as in: the output of `ht.minimum(x1,x2)` is "gathered" and has split=None, no matter whether the input arrays  are distributed. This PR changes this and the output has now the same split as the input.
  2. this is more of a general discussion we need to have about binary operations when the input arrays are distributed along different axes. So far we have been going out of our way to make sure that all split combinations are allowed, and the resplitting, if necessary, is done internally. After spending some time with real-life data analysis, I'm convincing myself that a common split axis should be enforced for binary operations. We can discuss this below. This PR introduces the requirement that the input arrays be split along the same axis, except if one of them is a scalar. Once there is agreement on this, we can take on #379

- speaking of scalars, the original implementation couldn't deal with one of the input arrays being a scalar. Fixed.

- finally, the original implementation required the input arrays to be of the same dtype (because torch.min() and torch.max() required it).  This requirement has now been lifted.

Issue/s resolved: #661 

## Changes proposed:
- `ht.maximum(x1, x2)`, `ht.minimum(x1, x2)` output has now the same split as the input
- input arrays must be split along the same axis, `split=None` allowed if ~~one of them is a scalar~~  broadcasting is involved 
- allow scalar input
- allow different input dtype
- update tests

## Type of change

`ht.maximum()`, `ht.minimum()` now return a distributed DNDarray if the input is distributed

## Due Diligence

No


mtar:

mtar:
ok to test

heat/core/statistics.py
```diff
@@ -558,25 +558,26 @@ def local_max(*args, **kwargs):
...
+        and the same split axis. If not provided or None, a freshly-allocated array is returned.
 
     Returns:
     --------
 
-    maximum: ht.DNDarray
+    maximum: DNDarray
             Element-wise maximum of the two input tensors.
 
     Examples:
```
lenablind, line 583: A few suggestions concerning the examples: 

- First of all, is it necessary to use `torch.manual_seed(1)` in line 587? Not the expert in pseudorandom generators, but if it isn't, I would recommend replacing this with our own function `ht.random.seed(1)` . 
- In line 616 I don't think `__setitem__` should be called directly (dunder method), better let Python do the work and use the function exploiting slicing as it is explained in the docs: https://heat.readthedocs.io/en/latest/autoapi/heat/cluster/index.html?highlight=__setitem__#id214
- The last one is more of a feature recommendation, but after the update of our print method and for future consistency, I would replace all `tensor` with `DNDarray` and the new presentation of DNDarrays, though I guess this will be done later for all docs either way

heat/core/statistics.py
```diff
@@ -626,77 +627,73 @@ def maximum(x1, x2, out=None):
...
+    # DNDarrays must be distributed along same axis
+    if split1 is not None and split2 is not None and split1 != split2:
+        raise ValueError(
+            "x1 and x2 must be distributed along the same axis, but x1.split is {}, x2.split is {}".format(
+                split1, split2
+            )
+        )
+    # Exception: split axes can differ if one of the DNDarrays is a scalar
+    if split1 is not None or split2 is not None:
+        if split1 is None and size1 > 1 or split2 is None and size2 > 1:
```
lenablind, line 655: What happens if one DNDarray is empty? Admittedly, this probably wouldn't make a huge difference for splits, but for pedantic reasons I would replace the > with !=

heat/core/tests/test_statistics.py
```diff
@@ -619,57 +619,51 @@ def test_maximum(self):
         self.assertTrue((maximum._DNDarray__array == torch.max(comparison1, comparison2)).all())
 
         # check maximum over float elements of split 3d tensors
-        # TODO: add check for uneven distribution of dimensions (see Issue #273)
-        size = ht.MPI_WORLD.size
         torch.manual_seed(1)
```
lenablind, line 622: Same question as in statistics - Can this be replaced with `ht.random.seed(1)` ?

lenablind (commented):
In general, I think it's great that these bugs have been fixed, so thank you for that. I haven't dived too deep into the split philosophy yet, so could you maybe list some of the reasons why you think a common split axis for binary operations might be the way to go?

heat/core/statistics.py
ClaudiaComito: Hi @lenablind , thanks for this, indeed I forgot or neglected to update the examples and the old implementation was from before ht.random.seed's times. I'll update everything before merging.

heat/core/statistics.py
ClaudiaComito: hmm, if the DNDarray is empty, the function should raise an error much earlier. Thanks for the reminder.

Otherwise, `size` is the total number of elements of the DNDarray so it can't be less than 1.

heat/core/tests/test_statistics.py
ClaudiaComito: yes, see above

heat/core/statistics.py
lenablind: Hi @ClaudiaComito, you're welcome, I'm glad I could help!

heat/core/statistics.py
lenablind: Alright, then I take my back my criticism. I know, though I think `size` can also be 0 as a DNDarray might also contain 0 elements/be empty. But if that case is already handled above I guess that doesn't matter.

ClaudiaComito:
> In general, I think it's great that these bugs have been fixed, so thank you for that. I haven't dived too deep into the split philosophy yet, so could you maybe list some of the reasons why you think a common split axis for binary operations might be the way to go?

OK good point. First of all, let's agree on the definition. Binary operations are operations that imply an element-by-element comparison between two arrays. The first requirement for binary operations is that the two arrays be of the same shape, or broadcastable to the same shape. I'll expand on this below, for now the way we have - at least tentatively - been implementing binary operations in heat is: it is allowed for the two input DNDarrays to be distributed along different axes. Because the element-by-element comparison happens locally via torch, logically the heat implementation has to make sure that one of the inputs gets redistributed within the function so that the split axes match.

Note that:
- performing a resplit internally slows down operations in ways that aren't immediately recognizable by the user; 
- performing all the checks to verify what needs to be resplit in which direction also weighs on the runtime; 
- last but not least, implementation is a lot more complicated (time-consuming but also complicating the code) if we have to take care of all the possible split combinations. 

It certainly makes sense to centralize all the split checks in operations.__binary_op() as much as possible, as @Markus-Goetz suggests in #379,  but really I'm not sure we should be spending so much time with the "all-splits-allowed" implementation, as I'm not sure that there's a good reason ever to want to perform a binary op on arrays with different splits. 

As I said above, we have 2 options, input arrays with same shape, or with different but broadcastable shapes.

**Case 1) `x1.shape == x2.shape`**

We want to calculate the element-by-element minimum, or difference, whatever, between two arrays with same shape. Maybe I'm just lacking in fantasy, but I can't imagine a reason why those 2 arrays should be split along different axes, if not by mistake. 
- If `x1.split != x2.split` by mistake, it's better to raise an error and get the mistake fixed
- If `x1.split != x2.split` by design for whatever reason, I think it's fair to expect the user to resplit one of the input arrays before performing the binary op. This way, they will be aware of the runtime load they're introducing. They might find ways to not have the inputs distributed differently after all.



**Case 2) `x1.shape != x2.shape`** 

Shapes are different but must be broadcastable. In practice we're talking about things like: subtracting a 1-D array from a 2-D array (just an example). In this case we must allow for different distribution, or in fact, for one of the DNDarrays to not be distributed at all. Example:

```
>>> x1 = ht.random.randn(500, 100, split=0)
>>> x1.shape
(500, 100)
>>> x1.split
0
```

then:

binary_op with `x1.split == x2.split`: **OK**

```
>>> x2 = ht.random.randn(500, 1, split=0)
>>> ht.minimum(x1, x2) 
```

binary_op with `x1.split not None` and `x2.split is None` because `x2.shape[x1.split] = 1`: **OK**
```
>>> x2 = ht.random.randn(1, 100, split=None)
>>> ht.minimum(x1, x2) 
```
binary_op with `x1.split != x2.split`: **raise error**
```
>>> x2 = ht.random.randn(1, 100, split=1)
>>> ht.minimum(x1, x2) 
ValueError(...)

>>> x2 = ht.random.randn(500, 1, split=None)
>>> ht.minimum(x1, x2) 
ValueError(...)

```

I hope this makes sense.

To summarize, I propose that distributed binary operations:
- don't resplit internally but raise an error if the input arrays are split along different axes
- allow for one of the inputs to not be distributed only when it makes sense, e.g. when `x2.shape[x1.split] = 1` (or the other way around, depending on which one is split).

P.S. (in case this wasn't long enough already) note that some binary ops are already implemented like this, see outer() or where(). I'm just trying to get to a uniform approach in view of dealing with #379.

lenablind:
> > In general, I think it's great that these bugs have been fixed, so thank you for that. I haven't dived too deep into the split philosophy yet, so could you maybe list some of the reasons why you think a common split axis for binary operations might be the way to go?
> 
> OK good point. First of all, let's agree on the definition. Binary operations are operations that imply an element-by-element comparison between two arrays. The first requirement for binary operations is that the two arrays be of the same shape, or broadcastable to the same shape. I'll expand on this below, for now the way we have - at least tentatively - been implementing binary operations in heat is: it is allowed for the two input DNDarrays to be distributed along different axes. Because the element-by-element comparison happens locally via torch, logically the heat implementation has to make sure that one of the inputs gets redistributed within the function so that the split axes match.
> 
> Note that:
> 
>     * performing a resplit internally slows down operations in ways that aren't immediately recognizable by the user;
> 
>     * performing all the checks to verify what needs to be resplit in which direction also weighs on the runtime;
> 
>     * last but not least, implementation is a lot more complicated (time-consuming but also complicating the code) if we have to take care of all the possible split combinations.
> 
> 
> It certainly makes sense to centralize all the split checks in operations.__binary_op() as much as possible, as @Markus-Goetz suggests in #379, but really I'm not sure we should be spending so much time with the "all-splits-allowed" implementation, as I'm not sure that there's a good reason ever to want to perform a binary op on arrays with different splits.
> 
> As I said above, we have 2 options, input arrays with same shape, or with different but broadcastable shapes.
> 
> **Case 1) `x1.shape == x2.shape`**
> 
> We want to calculate the element-by-element minimum, or difference, whatever, between two arrays with same shape. Maybe I'm just lacking in fantasy, but I can't imagine a reason why those 2 arrays should be split along different axes, if not by mistake.
> 
>     * If `x1.split != x2.split` by mistake, it's better to raise an error and get the mistake fixed
> 
>     * If `x1.split != x2.split` by design for whatever reason, I think it's fair to expect the user to resplit one of the input arrays before performing the binary op. This way, they will be aware of the runtime load they're introducing. They might find ways to not have the inputs distributed differently after all.
> 
> 
> **Case 2) `x1.shape != x2.shape`**
> 
> Shapes are different but must be broadcastable. In practice we're talking about things like: subtracting a 1-D array from a 2-D array (just an example). In this case we must allow for different distribution, or in fact, for one of the DNDarrays to not be distributed at all. Example:
> 
> ```
> >>> x1 = ht.random.randn(500, 100, split=0)
> >>> x1.shape
> (500, 100)
> >>> x1.split
> 0
> ```
> 
> then:
> 
> binary_op with `x1.split == x2.split`: **OK**
> 
> ```
> >>> x2 = ht.random.randn(500, 1, split=0)
> >>> ht.minimum(x1, x2) 
> ```
> 
> binary_op with `x1.split not None` and `x2.split is None` because `x2.shape[x1.split] = 1`: **OK**
> 
> ```
> >>> x2 = ht.random.randn(1, 100, split=None)
> >>> ht.minimum(x1, x2) 
> ```
> 
> binary_op with `x1.split != x2.split`: **raise error**
> 
> ```
> >>> x2 = ht.random.randn(1, 100, split=1)
> >>> ht.minimum(x1, x2) 
> ValueError(...)
> 
> >>> x2 = ht.random.randn(500, 1, split=None)
> >>> ht.minimum(x1, x2) 
> ValueError(...)
> ```
> 
> I hope this makes sense.
> 
> To summarize, I propose that distributed binary operations:
> 
>     * don't resplit internally but raise an error if the input arrays are split along different axes
> 
>     * allow for one of the inputs to not be distributed only when it makes sense, e.g. when `x2.shape[x1.split] = 1` (or the other way around, depending on which one is split).
> 
> 
> P.S. (in case this wasn't long enough already) note that some binary ops are already implemented like this, see outer() or where(). I'm just trying to get to a uniform approach in view of dealing with #379.

First of all, thank you for the detailed explanation! Makes sense to me. Therefore, I personally would approve a consistent common split axis approach for binary operations. Though I'm not sure how huge the impact of the restrictions might be. 

> > In general, I think it's great that these bugs have been fixed, so thank you for that. I haven't dived too deep into the split philosophy yet, so could you maybe list some of the reasons why you think a common split axis for binary operations might be the way to go?
> 
> OK good point. First of all, let's agree on the definition. Binary operations are operations that imply an element-by-element comparison between two arrays. The first requirement for binary operations is that the two arrays be of the same shape, or broadcastable to the same shape. I'll expand on this below, for now the way we have - at least tentatively - been implementing binary operations in heat is: it is allowed for the two input DNDarrays to be distributed along different axes. Because the element-by-element comparison happens locally via torch, logically the heat implementation has to make sure that one of the inputs gets redistributed within the function so that the split axes match.
> 
> Note that:
> 
>     * performing a resplit internally slows down operations in ways that aren't immediately recognizable by the user;
> 
>     * performing all the checks to verify what needs to be resplit in which direction also weighs on the runtime;
> 
>     * last but not least, implementation is a lot more complicated (time-consuming but also complicating the code) if we have to take care of all the possible split combinations.
> 
> 
> It certainly makes sense to centralize all the split checks in operations.__binary_op() as much as possible, as @Markus-Goetz suggests in #379, but really I'm not sure we should be spending so much time with the "all-splits-allowed" implementation, as I'm not sure that there's a good reason ever to want to perform a binary op on arrays with different splits.
> 
> As I said above, we have 2 options, input arrays with same shape, or with different but broadcastable shapes.
> 
> **Case 1) `x1.shape == x2.shape`**
> 
> We want to calculate the element-by-element minimum, or difference, whatever, between two arrays with same shape. Maybe I'm just lacking in fantasy, but I can't imagine a reason why those 2 arrays should be split along different axes, if not by mistake.
> 
>     * If `x1.split != x2.split` by mistake, it's better to raise an error and get the mistake fixed
> 
>     * If `x1.split != x2.split` by design for whatever reason, I think it's fair to expect the user to resplit one of the input arrays before performing the binary op. This way, they will be aware of the runtime load they're introducing. They might find ways to not have the inputs distributed differently after all.
> 
> 
> **Case 2) `x1.shape != x2.shape`**
> 
> Shapes are different but must be broadcastable. In practice we're talking about things like: subtracting a 1-D array from a 2-D array (just an example). In this case we must allow for different distribution, or in fact, for one of the DNDarrays to not be distributed at all. Example:
> 
> ```
> >>> x1 = ht.random.randn(500, 100, split=0)
> >>> x1.shape
> (500, 100)
> >>> x1.split
> 0
> ```
> 
> then:
> 
> binary_op with `x1.split == x2.split`: **OK**
> 
> ```
> >>> x2 = ht.random.randn(500, 1, split=0)
> >>> ht.minimum(x1, x2) 
> ```
> 
> binary_op with `x1.split not None` and `x2.split is None` because `x2.shape[x1.split] = 1`: **OK**
> 
> ```
> >>> x2 = ht.random.randn(1, 100, split=None)
> >>> ht.minimum(x1, x2) 
> ```
> 
> binary_op with `x1.split != x2.split`: **raise error**
> 
> ```
> >>> x2 = ht.random.randn(1, 100, split=1)
> >>> ht.minimum(x1, x2) 
> ValueError(...)
> 
> >>> x2 = ht.random.randn(500, 1, split=None)
> >>> ht.minimum(x1, x2) 
> ValueError(...)
> ```
> 
> I hope this makes sense.
> 
> To summarize, I propose that distributed binary operations:
> 
>     * don't resplit internally but raise an error if the input arrays are split along different axes
> 
>     * allow for one of the inputs to not be distributed only when it makes sense, e.g. when `x2.shape[x1.split] = 1` (or the other way around, depending on which one is split).
> 
> 
> P.S. (in case this wasn't long enough already) note that some binary ops are already implemented like this, see outer() or where(). I'm just trying to get to a uniform approach in view of dealing with #379.

First of all, thank you for your detailed explanation! Makes sense to me and will probably simplify usage for both developer and client (and prevent unnoticed mistakes as you already outlined). Therefore, though not having much insight in the impact of the restriction, I personally would approve a consistent common split axis approach for binary operations.

heat/core/statistics.py
```diff
@@ -626,77 +627,94 @@ def maximum(x1, x2, out=None):
...
+        in_x[not_on_gpu] = in_x[not_on_gpu].gpu()
+
+    # sanitize shapes
+    output_gshape = stride_tricks.broadcast_shape(shapes[0], shapes[1])
+
+    # sanitize splits
+    if not splits.count(None) == 2:
+        # DNDarrays must be distributed along same axis
+        if splits.count(None) == 0:
+            if splits.count(splits[0]) == len(splits):
```
coquelin77, line 665: i really like this logic here, but do we need to count in the last if loop? isnt this the same as `splits[0] == splits[1]` since its only of length two? this is just to make it slightly easier to read, i dont think that there will be any noticable speedup here so feel free to ignore this if you like it the way it is

heat/core/statistics.py
```diff
@@ -626,77 +627,94 @@ def maximum(x1, x2, out=None):
...
+            raise TypeError("expected `out` to be None or a DNDarray, but was {}".format(type(out)))
+        if out.gshape != output_gshape:
             raise ValueError(
-                "Expecting output buffer of shape {}, got {}".format(output_lshape, out.shape)
+                "Expecting output buffer of shape {}, got {}".format(output_gshape, out.shape)
+            )
+        if out.split is not split:
+            raise ValueError(
+                "Split axis of output buffer is inconsistent with split semantics (see documentation)."
             )
```
coquelin77, line 702: would it be possible to make this a `sanitize_out()` function? these seems like the kind of logic that we call could use for the out parameter in most function. maybe it could be combined with the setting logic as well? (lines 711-716)

heat/core/statistics.py
```diff
@@ -626,77 +627,94 @@ def maximum(x1, x2, out=None):
...
-        if x1.split is None:
-            x1.resplit_(x2.split)
-        if x2.split is None:
-            x2.resplit_(x1.split)
-        if x1.split != x2.split:
-            if np.prod(x1.gshape) < np.prod(x2.gshape):
-                x1.resplit_(x2.split)
-            if np.prod(x2.gshape) < np.prod(x1.gshape):
-                x2.resplit_(x1.split)
+    # TODO: move all of this to __binary_op
```
coquelin77, line 630: do you want to move this logic there before or after this is merged? have you done speed comparisons between the old and new implementations?

heat/core/statistics.py
```diff
@@ -626,77 +627,94 @@ def maximum(x1, x2, out=None):
...
             )
-        out._DNDarray__array = lresult._DNDarray__array
-        out._DNDarray__dtype = lresult._DNDarray__dtype
-        out._DNDarray__split = split
-        out._DNDarray__device = x1.device
-        out._DNDarray__comm = x1.comm
 
-    return lresult
+    # calculate process-local element-wise minimum
+    t_dtype = torch.promote_types(t_x[0].dtype, t_x[1].dtype)
```
coquelin77, line 705: i think that we should use the heat promote types. we currently use the same rules as torch for type promotion but if we ever decide to differ from that this would cause issues. it should be the same heat function just with heat dtypes

heat/core/statistics.py
```diff
@@ -1043,77 +1063,94 @@ def minimum(x1, x2, out=None):
...
```
coquelin77, line 1141: heat promote type?

heat/core/tests/test_statistics.py
```diff
@@ -619,71 +619,77 @@ def test_maximum(self):
...
-        self.assertEqual(maximum_volume_splitdiff._DNDarray__array.dtype, torch.float32)
-        self.assertEqual(maximum_volume_splitdiff.split, 0)
-
-        random_volume_1_splitdiff = ht.random.randn(size * 3, size * 3, 4, split=1)
-        random_volume_2_splitdiff = ht.random.randn(size * 3, size * 3, 4, split=0)
-        maximum_volume_splitdiff = ht.maximum(random_volume_1_splitdiff, random_volume_2_splitdiff)
-        self.assertEqual(maximum_volume_splitdiff.split, 0)
-
-        random_volume_1_split_none = ht.random.randn(size * 3, size * 3, 4, split=None)
-        random_volume_2_splitdiff = ht.random.randn(size * 3, size * 3, 4, split=1)
```
coquelin77, line 654: why are all of these removed? are they not valid? or are they covered somewhere else?

coquelin77 (changes requested):

Markus-Goetz:
Regarding the split consistency, a couple of thoughts. Generally, I am with you that if you want to code to actually be able to scale, you either need identical splits or a split and a non-split tensor. Other split combinations will eventually become a computational bottleneck. Let's see what the implications are on that if we were to enforce consistency:

* More run-time checks taking time (-)
* Less liberal programming, more exceptions (-)
* Less implementation overhead for making all-to-all combinations work (+)
* Probably better scaling code (+)

Moreover, there are some very reasonable operations, where different split combinations should actually be allowed/are advantageous, like matrix multiply for example. Unless I forgot some major point, it is a typical comfort versus execution time discussion. Personally, given our target audience, I would favour the latter and rather your argumentation line.

heat/core/statistics.py
ClaudiaComito: Hi @coquelin77 , thanks for the comments. 
I've just run a very rudimentary check on my laptop. Here's the code:
```
x = ht.array(np.random.randn(20000, 2000), dtype=ht.float32, split=0)
y = ht.array(np.random.randn(20000, 2000), dtype=ht.float32, split=0)

repeats = 10
ht.minimum(x, y) #warm-up

start = time.time()
for i in range(repeats):
    z = ht.minimum(x, y)
end = time.time()

print("Minimum took ", (end - start) / repeats, " sec on average with ", x.comm.size, " processes and split ", x.split)
```
And here are the results:
**on the main branch**
```
(py37) Claudias-MBP:heat c.comito$ mpirun --tag-output -n 2 --use-hwthread-cpus python local_test.py
[1,0]<stdout>:Minimum took  0.16146039962768555  sec on average with  2  processes and split  0
[1,1]<stdout>:Minimum took  0.1614467144012451  sec on average with  2  processes and split  0

(py37) Claudias-MBP:heat c.comito$ mpirun --tag-output -n 4 --use-hwthread-cpus python local_test.py
[1,1]<stdout>:Minimum took  0.2378169059753418  sec on average with  4  processes and split  0
[1,2]<stdout>:Minimum took  0.23781290054321289  sec on average with  4  processes and split  0
[1,0]<stdout>:Minimum took  0.23747990131378174  sec on average with  4  processes and split  0
[1,3]<stdout>:Minimum took  0.23850111961364745  sec on average with  4  processes and split  0
```
**On this branch (bug/minimum-split):**
```
(py37) Claudias-MBP:heat c.comito$ git checkout bug/minimum-split
Switched to branch 'bug/minimum-split'
Your branch is up to date with 'origin/bug/minimum-split'.
(py37) Claudias-MBP:heat c.comito$ mpirun --tag-output -n 2 --use-hwthread-cpus python local_test.py
[1,0]<stdout>:Minimum took  0.058056998252868655  sec on average with  2  processes and split  0
[1,1]<stdout>:Minimum took  0.05825142860412598  sec on average with  2  processes and split  0

(py37) Claudias-MBP:heat c.comito$ mpirun --tag-output -n 4 --use-hwthread-cpus python local_test.py
[1,0]<stdout>:Minimum took  0.05380411148071289  sec on average with  4  processes and split  0
[1,2]<stdout>:Minimum took  0.0541417121887207  sec on average with  4  processes and split  0
[1,3]<stdout>:Minimum took  0.05416550636291504  sec on average with  4  processes and split  0
[1,1]<stdout>:Minimum took  0.05364921092987061  sec on average with  4  processes and split  0
```
(yes yes I know I can format the floats :p )

Note that I haven't got 4 CPUs, those are 4 hardware threads.

I'm quite surprised it makes such a difference tbh, but anyway, there you have it.

heat/core/statistics.py
ClaudiaComito: those are two torch tensors for a torch operation, it didn't occur to me to go through the heat dtype promotion and then back. This about type promotion is a discussion we need to have again I'm afraid.

heat/core/tests/test_statistics.py
ClaudiaComito: Oh, I updated test_minimum and then deleted the old test_maximum and replaced it with the new version, that's why it looks like the whole thing is gone.

heat/core/statistics.py
ClaudiaComito: You're absolutely right, no need to be so generic here, I'll change it.

heat/core/statistics.py
ClaudiaComito: Yes, I think I should stop procrastinating on that sanitation module (#468)

heat/core/statistics.py
```diff
@@ -626,77 +627,94 @@ def maximum(x1, x2, out=None):
...
+        shapes.append(x.gshape)
+        comms.append(x.comm)
+        devices.append(x.device.device_type)
+        t_x.append(x._DNDarray__array)
+
+    # sanitize devices
+    if devices.count("gpu") == 1:
+        not_on_gpu = abs(devices.index("gpu") - 1)
+        # move to gpu
+        in_x[not_on_gpu] = in_x[not_on_gpu].gpu()
```
ClaudiaComito, line 656: @mtar do you think this check is needed? I'm thinking it might be redundant with the new setup

mtar: I don't think you will need it if an error is raised on different devices. However, you will need to specify the device in line 640 explicitly.

ClaudiaComito: I see. Hm, but at line 640 I don't know yet on what device the other array is. 

I'll leave this check in. It's just that it can't be tested  and codecov complains. I guess we'll have to live with that. Thanks @mtar!

ClaudiaComito:
> 
> Moreover, there are some very reasonable operations, where different split combinations should actually be allowed/are advantageous, like matrix multiply for example. Unless I forgot some major point, it is a typical comfort versus execution time discussion. Personally, given our target audience, I would favour the latter and rather your argumentation line.

Thanks @Markus-Goetz . 

OK guys:
- `__binary_op` can now write to `out`
- `minimum()` and `maximum()` now call `__binary_op()`
- basic `sanitation` module is in place.

heat/core/statistics.py
ClaudiaComito: sanitation module with basic functions (including sanitize_out) is in now

coquelin77 (approved):

Issue #665, opened 2020-09-07
axis parameter of reshape does not change split axis


lenablind:
**Description**
Setting the split axis using the `axis` parameter of `reshape` actually has no effect, respectfully the `split` axis remains `None`.

**To Reproduce**
1. Which module/class/function is affected?
   function reshape in module heat.core.manipulations /
   function reshape in module heat.core.dndarray

2. What are the circumstances under which the bug appears?
    Normal execution of the function

3. What is the exact error message / erroneous behavior?
   Split axis is `None`, though it should be set to a different value (i.e. in this case 0).
   No error message

**Expected behavior**
Setting a value for the `axis` parameter of `reshape` results in that being the `split` axis for the corresponding DNDarray.

**Illustrative**
Example:
`a = ht.arange(12)`
`>>> DNDarray([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11], dtype=ht.int32, device=cpu:0, split=None)`
`a = a.reshape(shape = (2, 2, 3), axis = 0)`
`>>> DNDarray([[[ 0,  1,  2],
           [ 3,  4,  5]],
          [[ 6,  7,  8],
           [ 9, 10, 11]]], dtype=ht.int32, device=cpu:0, split=None)`

The same happens (as expected) replacing `a = a.reshape(shape = (2, 2, 3), axis = 0)` with `a = ht.reshape(a, (2, 2, 3), 0)`

**Version Info**
 Which version are you using?
0.4.0

**Additional comments**
A presumption for this bug: eventually there is not enough data which causes the split avoidance (see example above). Still, I think there should at least be a warning if that is the case.

mtar:
It is intended that the parameter should only work on split DNDarray inputs.

lenablind:
> It is intended that the parameter should only work on split DNDarray inputs.
> reshape and then resplit should be faster on unsplit DNDarrays in general.

Okay, could we maybe integrate a hint to that in the docs?

PR #667, opened 2020-09-07
Enhancement reshape rename axis to new_split and add description

mtar:
## Description

Renaming of the `axis` parameter and additional hint that it only works on split DNDarrays. 

Issue/s resolved: #665 

## Changes proposed:
- rename `axis` to `new_split`
- add further description

## Type of change

- Documentation update

## Due Diligence

no


lenablind (approved):
Thank you for adapting the function so quickly!

Issue #671, opened 2020-09-14
communication MPIRequest.wait()


mtar:
**Description**
A clear and concise description of the bug and the associated functionality.

ht.communication substitutes the capitalized MPI functions only. The lowercase functions are the standard mpi4py functions.
Therefore, MPIRequest should not modify wait(). Also, Wait() should be called only if one of the MPI functions defined in communications is called.

To fix:
- delete wait() in MPIRequest
- rewrite all instances where MPIRequest.wait() is called wrongly

PR #672, opened 2020-09-14
remove MPIRequest.wait()

mtar:
## Description

Removes the `wait` method of `MPIRequest` such that it does not interfere with the lowercase non-blocking MPI functions of mpi4py.

Issue/s resolved: #671 

## Changes proposed:
- delete MPIRequest.wait()
- fix calls using wait() to Wait()

## Type of change

- Bug fix
- Breaking change

## Due Diligence

no


mtar:
rerun tests

coquelin77:
why do we need to remove `wait`? what was it interfering with? isnt it something that we should have?

coquelin77 (approved):

Issue #179, opened 2019-03-12
Implement repeat()


Markus-Goetz:

TheSlimvReal:
Be aware that `numpy.repeat` is equivalent to `torch.repeate_interleave` while `torch.repeat` is more similar to `numpy.tile`. 
In order to implement the repeat functionality #176 should be implemented first.

PR #674, opened 2020-09-18
Features/179 repeat

lenablind:
## Description

Implementation of function `repeat()` based on `np.repeat()`
For process-local operations, I use `torch.repeat_interleave`

Docs numpy: https://numpy.org/doc/stable/reference/generated/numpy.repeat.html
Docs pytorch: https://pytorch.org/docs/stable/generated/torch.repeat_interleave.html

### Strategy

In the following, only the distributed case (<=> `a.split != None`) will be explained more specifically as the algorithm otherwise results mainly in a direct call of the torch function.

#### axis is None
Implies `a.split = 0` as the (total) result would be in the wrong order otherwise.
To assure the correct distribution of `repeats` and  syntactical consistency with numpy (`repeats` must be 1-dimensional), `repeats` has to be reshaped to the global shape of `a`, redistributed along axis 0 and be flattened afterwards.
The last step is necessary due to compatibility with torch (1-dimensionality required as in numpy).

#### axis is not None
Depending on whether `a.split == axis`,  `repeats` has to be either split along axis 0 (case True) or gathered on all processes (case False). The eventually resulting redistribution of data is needed to assure correct local and therefore global results.

Issue/s resolved: #179 

## Changes proposed:
- Implemented new function: `manipulations.repeat()`
- Implemented corresponding tests

## Type of change

## Due Diligence

no


mtar:

heat/core/manipulations.py
```diff
@@ -911,6 +911,268 @@ def hstack(tup):
...
+        return a
+
+    # Broadcast (via int or 1-element DNDarray)
+    if isinstance(repeats, int) or repeats.gnumel == 1:
+        if axis is None and a.split is not None and a.split != 0:
+            print(
+                "\n!!! WARNING !!!\nIf axis is None, a has to be split along axis 0 if distributed.\nSplit axis of a "
+                "will be changed from {} to 0.".format(a.split)
+            )
+            a.resplit_(0)
```
coquelin77, line 1052: throw warning not print. see the `warning` module.

heat/core/manipulations.py
```diff
@@ -911,6 +911,268 @@ def hstack(tup):
...
+        if isinstance(repeats, int):
+            repeated_array_torch = torch.repeat_interleave(a._DNDarray__array, repeats, axis)
+        else:
+            if repeats.split is not None:
+                print(
+                    "\n!!! WARNING !!!\nFor broadcast via array_like repeats, repeats must not be "
+                    "distributed.\nSplit axis of repeats "
+                    "will be changed from {} to None.".format(repeats.split)
+                )
```
coquelin77, line 1061: warnings

heat/core/manipulations.py
```diff
@@ -911,6 +911,268 @@ def hstack(tup):
...
+    else:
+        # check if the data chunks of `repeats` and/or `a` have to be (re)distributed before call of torch function.
+
+        # UNDISTRIBUTED CASE (a not distributed)
+        if a.split is None:
+            if repeats.split is not None:
+                print(
+                    "\n!!! WARNING !!!\nIf a is undistributed, repeats also has to be undistributed.\nSplit axis of "
+                    "repeats will be changed from {} to None.".format(repeats.split)
+                )
```
coquelin77, line 1076: warning

heat/core/manipulations.py
coquelin77, line 1052: also, does `a` need to be resplit? could we get away with a copy and warn the user that the data is copied?

heat/core/manipulations.py
```diff
@@ -911,6 +911,268 @@ def hstack(tup):
...
+                repeats.resplit_(None)
```
coquelin77, line 1077: same comment as above, do we need to do an in-place resplit or can we get away with an out of place resplit?

heat/core/manipulations.py
```diff
@@ -911,6 +911,268 @@ def hstack(tup):
...
+
+            # Check correct input
+            if axis is None:
+                # check matching shapes (repetition defined for every element)
+                if a.gnumel != repeats.gnumel:
+                    raise ValueError(
+                        "Invalid input. Sizes of flattened a ({}) and repeats ({}) are not same. "
```
coquelin77, line 1084: can you capitalize `a`? or put it in quotes? something to indicate that its an input parameter and not just the letter a

heat/core/manipulations.py
```diff
@@ -911,6 +911,268 @@ def hstack(tup):
...
+                        " scalar.".format(a.gnumel, repeats.gnumel)
+                    )
+
+                if a.split != 0:
+                    print(
+                        "\n!!! WARNING !!!\nIf axis is None, a has to be split along axis 0 if distributed.\nSplit "
+                        "axis of a "
+                        "will be changed from {} to 0.".format(a.split)
+                    )
+                    a.resplit_(0)
```
coquelin77, line 1115: warnings module and resplit question

heat/core/manipulations.py
```diff
@@ -911,6 +911,268 @@ def hstack(tup):
...
+
+                repeats = repeats.reshape(a.gshape)
+                if repeats.split != 0:
+
+                    print(
+                        "\n!!! WARNING !!!\nIf axis is None, repeats has to be split along axis 0 if "
+                        "distributed.\nSplit axis of repeats "
+                        "will be changed from {} to 0.".format(repeats.split)
+                    )
+                    repeats.resplit_(0)
```
coquelin77, line 1125: warnings and resplit

coquelin77 (changes requested):
looks pretty good. good job!

lenablind:
> looks pretty good. good job!

Thank you!

heat/core/manipulations.py
lenablind: That's actually a really nice module I wasn't aware of - thanks for the advice!

lenablind: I think a copy will be fine as well, I was just expecting the code to run faster if I do the resplit in place. Could you explain to me when to use the function directly/without the copy? Also, to which category would the warning belong? ResourceWarning? Do we have a convention on that or is it up to each developer?

heat/core/manipulations.py
lenablind: I guess we can get away with an out of place resplit, see above

heat/core/manipulations.py
lenablind: For sure! I got confused about this as well :smile:

mtar:
ok to test

heat/core/manipulations.py
```diff
@@ -911,6 +911,263 @@ def hstack(tup):
...
+        raise ValueError(
+            "Invalid input for `axis`. Value has to be either None or between 0 and {}, not {}.".format(
+                len(a.shape) - 1, axis
+            )
+        )
+
+    # sanitation `repeats`
+    if not isinstance(repeats, (int, list, tuple, np.ndarray, dndarray.DNDarray)):
+        raise TypeError(
+            "`repeats` must be a an integer, list, tuple, np.ndarray or ht.DNDarray of integers, currently: {}".format(
```
coquelin77, line 984: "must be an integer" (extra 'a' in the middle)

heat/core/manipulations.py
```diff
@@ -911,6 +911,263 @@ def hstack(tup):
...
+                type(repeats)
+            )
+        )
+    # broadcast implied
+    if isinstance(repeats, int):
+        pass
+    else:
```
coquelin77, line 991: `if not isinstance(repeats, int):` has the same effect as this

heat/core/manipulations.py
```diff
@@ -911,6 +911,263 @@ def hstack(tup):
...
+                    dtype=types.int64,
+                    is_split=repeats.split,
+                    device=repeats.device,
+                    comm=repeats.comm,
+                )
+            else:
+                raise TypeError(
+                    "Invalid dtype for ht.DNDarray `repeats`. Has to be integer,"
+                    " but was {}".format(repeats.dtype)
+                )
```
coquelin77, line 1008: is there a reason why you cast repeats to `int64`? wouldnt this work with any int dtype?

coquelin77 (changes requested):
very very minor changes and only a couple questions. good work!

heat/core/manipulations.py
lenablind: @coquelin77 Yes, there is and no it wouldn't. Torch requires `repeats` to be a Long tensor, which is represented by `ht.int64` if I've got the type mapping correctly in mind

lenablind:
> very very minor changes and only a couple questions. good work!

@coquelin77 Thank you!

heat/core/manipulations.py
coquelin77: the in-place resplit will change the input matrix which can change things in future for the user. if they give a matrix to this function then try to do something later with it it can change the split of the matrix outside of the function. this is sort of bad practice, since in the other cases `a` would not be modified. you could replace this with `a = manipulations.resplit(0)` and it would probably be faster for small to large arrays (but not for very large arrays).

i think either Resource or User warnings would both be fine. its up to you

coquelin77 (approved):

Issue #510, opened 2020-03-27
implement split functions


mtar:
**Feature functionality**
A clear and concise description of what the feature should do exactly and any alternative solutions.

Provide split functions (split, hsplit, vsplit, dsplit, ...)

coquelin77:
what do you mean split function? functions to resplit the dndarray?

mtar:
No, really splitting it into two seperate tensors like in the numpy functions I mentioned.

Cdebus:
I think in our context, 'splitting' might be a poor choice of words. To avoid confusion, we should find other naming, even if it deviates from the numpy API. Just my two cents

lenablind:
> I think in our context, 'splitting' might be a poor choice of words. To avoid confusion, we should find other naming, even if it deviates from the numpy API. Just my two cents

I would agree. Any suggestions? @Cdebus @coquelin77 @mtar

PR #677, opened 2020-09-25
Features/510 split

lenablind:
## Description

Implementation of functions `split()`, `vsplit()`, `hsplit()`, `dsplit()` based on np analogues.
For process-local operations, I use `torch.split`

Docs numpy: 
- split: https://numpy.org/doc/stable/reference/generated/numpy.split.html
- vsplit: https://numpy.org/doc/stable/reference/generated/numpy.vsplit.html
- hsplit: https://numpy.org/doc/stable/reference/generated/numpy.hsplit.html
- dsplit: https://numpy.org/doc/stable/reference/generated/numpy.dsplit.html

Docs pytorch: https://pytorch.org/docs/stable/generated/torch.split.html

### differences numpy and torch 
This affects the input parameter `indices_or_sections` (np) / `split_size_or_sections` (torch).
As the variations in the given names might already reveal, the parameters contain different types of information depending on the used library.

_Numpy_:
- _integer_:
  In this case, `indices_or_sections` indicates in how many parts the input `ary` shall be split. 
  Therefore, `indices_or_sections = 2` will result into a list containing 2 DNDarrays.
- _array_like_
  On the other hand, `indices_or_sections` can be handled as a list containing the boundaries of a series of slices
  For instance, `indices_or_sections = [1, 3, 4]` will indicate the following data chunks:
      - [: 1]
      - [1: 3]
      - [3: 4]
      - [4: ]

_Torch_:
In contrary to numpy, the torch analogue always contains the size(s) of the resulting tensor(s).
If _integer_, all (function-)split tensors will have the same size, otherwise the possibly varying split sizes might be defined in an _array_like_.

According to these differences, the parameter semantics of _numpy_ are mapped to those of _torch_ as explained in the following.

### Strategy
As `vsplit, dsplit, hsplit` are no more than calls of `split` with a specific axis, I'll reduce the explanation to the algorithm of the latter.

Depending on whether `ary` is distributed in the same dimension as it is split within the function (therefore `ary.split == axis`), the strategy varies, as the data chunks on each node have to be reorganized correctly in the resulting list of (function-) split DNDarrays. (Challenge: Data of (function-) split DNDarrays is split via MPI)

#### axis != ary.split
Mapping np -> torch:
If `indices_or_sections` is _integer_, the resulting size for all DNDarrays will be calculated using the size of the affected axis.
If `indices_or_sections` is _array_like_, the resulting split sizes are calculated using`ht.diff` (1st discrete difference).

#### axis == ary.split
- `indices_or_sections` is _integer_:
   - **CASE 1** : number of processes == _indices_or_selections_
     In this case, the split matches the distribution and the data on the node can be used directly and inserted into the resulting list.
     The remaining DNDarrays are filled with empty DNDarrays of the needed shape.
   - **CASE 2**: number of processes != _indices_or_selections_
      The goal is, to use `torch.split` with the process-adapted/correctly calculated `indices_or_sections` parameter (`new_indices`)
      1. Calculate the index of the first element in the resulting list which needs data of the current process
      2. Calculate the amount of data which is needed to fill this chunk up
      3. Use the remaining data of the process for the following chunks
- `indices_or_sections` is _array_like_ / _DNDarray_ after sanitation
   The goal is, to use `torch.split` with the process-adapted/correctly calculated `indices_or_sections` parameter.
   Therefore, reduce the input information of `indices_or_sections` to the process relevant/ the given indices (of the global DNDarray) which correspond to the (local) data on the node. The needed information is provided via the slice out of `ht.comm.chunk`
   Afterwards, map the _numpy_ to the _torch_ semantics, as described above in `axis != ary.split`

      
In all cases, every DNDarray within the list is balanced before being returned.

#### Overview split functions and related axis
| Function       | Axis     | 
| :------------- | :----------: | 
|  _vsplit_ | 0   | 
| _hsplit_   | 1 | 
|  _dsplit_ | 2   | 

Hint: In contrary to the corresponding general split function call with axis 1, the input within `hsplit` is allowed to be 1-dimensional.
This results into some dimension changing operations in the HeAT version, more specifically a reshape adding one dimension and a flattening afterwards, to meet full consistency with numpy.

Issue/s resolved: #510 

## Changes proposed:
- Implemented new functions: `manipulations.split()`, `manipulations.hsplit()`, `manipulations.vsplit()`, `manipulations.dsplit()`
- Implemented corresponding tests

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -939,8 +939,70 @@ def flipud(a):
     return flip(a, 0)
 
 
-def hsplit(ary, indices_or_sections):  # TODO
-    pass
+def hsplit(ary, indices_or_sections):
+    """
+    Split array into multiple sub-arrays along the 2nd axis (horizontally/column-wise).
+
+    Please refer to the split documentation. hsplit is nearly equivalent to split with axis=1,
+    the array is always split along the second axis though, in contrary to split, regardless of the array dimension.
+
+    Parameters
+    ----------
+    ary : DNDarray
+        DNDArray to be divided into sub-DNDarrays.
+    indices_or_sections : int or 1-dimensional array_like (i.e. undistributed DNDarray, list or tuple)
+        If indices_or_sections is an integer, N, the DNDarray will be divided into N equal DNDarrays along the 2nd axis.
+        If such a split is not possible, an error is raised.
+        If indices_or_sections is a 1-D DNDarray of sorted integers, the entries indicate where along the 2nd axis
+        the array is split.
+        If an index exceeds the dimension of the array along the 2nd axis, an empty sub-array is returned correspondingly.
+
+    Returns
+    -------
+    sub_arrays : list of DNDarrays
+        A list of sub-DNDarrays as views into ary.
+
+    Raises
+    ------
+    ValueError
+        If indices_or_sections is given as integer, but a split does not result in equal division.
+
+    Examples
+    --------
+    >>> x = ht.arange(24).reshape((2, 4, 3))
+    >>> ht.hsplit(x, 2)
+        [
+            DNDarray([[[ 0,  1,  2],
+                       [ 3,  4,  5]],
+
+                      [[12, 13, 14],
+                       [15, 16, 17]]]),
+            DNDarray([[[ 6,  7,  8],
+                       [ 9, 10, 11]],
+
+                      [[18, 19, 20],
+                       [21, 22, 23]]])
+        ]
+
+    >>> ht.hsplit(x, [1, 3])
+        [
+            DNDarray([[[ 0,  1,  2]],
+
+                      [[12, 13, 14]]]),
+            DNDarray([[[ 3,  4,  5],
+                       [ 6,  7,  8]],
+
+                      [[15, 16, 17],
+                       [18, 19, 20]]]),
+            DNDarray([[[ 9, 10, 11]],
+
+                      [[21, 22, 23]]])]
+       """
+    sanitation.sanitize_input(ary)
+
+    if len(ary.lshape) < 2:
+        ary = ary.reshape(ary, (1, ary.lshape[0]))
+    return split(ary, indices_or_sections, 1)
 
 
 def hstack(tup):
@@ -2593,8 +2655,73 @@ def unique(a, sorted=False, return_inverse=False, axis=None):
     return return_value
 
 
-def vsplit(ary, indices_or_sections):  # TODO
-    pass
+def vsplit(ary, indices_or_sections):
+    """
+    Split array into multiple sub-arrays along the 1st axis (vertically/row-wise).
+
+    Please refer to the split documentation. hsplit is equivalent to split with axis=0,
+    the array is always split along the first axis regardless of the array dimension.
+
+    Parameters
+    ----------
+    ary : DNDarray
+        DNDArray to be divided into sub-DNDarrays.
+    indices_or_sections : int or 1-dimensional array_like (i.e. undistributed DNDarray, list or tuple)
+        If indices_or_sections is an integer, N, the DNDarray will be divided into N equal DNDarrays along the 1st axis.
+        If such a split is not possible, an error is raised.
+        If indices_or_sections is a 1-D DNDarray of sorted integers, the entries indicate where along the 1st axis
+        the array is split.
+        If an index exceeds the dimension of the array along the 1st axis, an empty sub-array is returned correspondingly.
+
+    Returns
+    -------
+    sub_arrays : list of DNDarrays
+        A list of sub-DNDarrays as views into ary.
+
+    Raises
+    ------
+    ValueError
+        If indices_or_sections is given as integer, but a split does not result in equal division.
+
+    Examples
+    --------
+    >>> x = ht.arange(24).reshape((4, 3, 2))
+    >>> ht.vsplit(x, 2)
+        [
+            DNDarray([[[ 0,  1],
+                       [ 2,  3],
+                       [ 4,  5]],
+
+                      [[ 6,  7],
+                       [ 8,  9],
+                       [10, 11]]]),
+            DNDarray([[[12, 13],
+                       [14, 15],
+                       [16, 17]],
+
+                      [[18, 19],
+                       [20, 21],
+                       [22, 23]]])
+        ]
+
+        >>> ht.vsplit(x, [1, 3])
+        [
+            DNDarray([[[0, 1],
+                       [2, 3],
+                       [4, 5]]]),
+            DNDarray([[[ 6,  7],
+                       [ 8,  9],
+                       [10, 11]],
+
+                      [[12, 13],
+                       [14, 15],
+                       [16, 17]]]),
+            DNDarray([[[18, 19],
+                       [20, 21],
+                       [22, 23]]])]
+
+           """
+    return split(ary, indices_or_sections, 0)
 
 
 def resplit(arr, axis=None):
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -1477,8 +1477,9 @@ def split(ary, indices_or_sections, axis=0):
                     ary.gshape, indices_or_sections, axis
                 )
             )
-        # to adapt torch syntax
-        indices_or_sections -= 1
+        # np to torch mapping - calculate size of resulting data chunks
+        indices_or_sections_t = ary.gshape[axis] // indices_or_sections
+
     elif isinstance(indices_or_sections, (list, tuple, dndarray.DNDarray)):
         if isinstance(indices_or_sections, (list, tuple)):
             indices_or_sections = factories.array(indices_or_sections)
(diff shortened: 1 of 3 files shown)


mtar:

mtar:
run tests

heat/core/manipulations.py
```diff
@@ -1749,6 +1897,252 @@ def sort(a, axis=None, descending=False, out=None):
...
+
+    sub_arrays_ht = [
+        factories.array(
+            sub_DNDarray, dtype=ary.dtype, is_split=ary.split, device=ary.device, comm=ary.comm
+        )
+        for sub_DNDarray in sub_arrays_t
+    ]
+
+    for sub_DNDarray in sub_arrays_ht:
+        sub_DNDarray.balance_()
```
coquelin77, line 2141: i think that this should be communicated in the docs. Since all of the sub-arrays are balanced, they are not linked to the original data points...this spawns a larger question.

coquelin77 (commented):
this all looks good to me. However, there is a larger comment to be discussed.

Do we want to have this return actual views of the DNDarray? If so, it would mean that we cannot balance the DNDarrays, and it might change some of the logic involved in the function itself. @ClaudiaComito @Markus-Goetz , thoughts?

coquelin77:
rerun tests

heat/core/manipulations.py
```diff
@@ -844,6 +921,77 @@ def flipud(a):
...
+                      [[21, 22, 23]]])]
+       """
+    sanitation.sanitize_input(ary)
+
+    if len(ary.lshape) < 2:
+        ary = reshape(ary, (1, ary.lshape[0]))
+        result = split(ary, indices_or_sections, 1)
+        result = [flatten(sub_array) for sub_array in result]
+    else:
+        result = split(ary, indices_or_sections, 1)
```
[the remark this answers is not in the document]
lenablind: @coquelin77  No, it isn't. `hsplit` is some kind of a special function among the three `split` variations. In contrary to the corresponding general `split` function call with axis 1, the `ary` is allowed to be 1-dimensional (in `numpy`) even if (functionally) split among the second axis. As it isn't in `torch`, `ary` has to be reshaped (adding one dimension). The subsequent flattening is needed to meet full consistency with `numpy`. 
As this exception doesn't occur within the other `split` functions, no shape manipulating is needed there.

heat/core/manipulations.py
lenablind: @coquelin77 I agree, but I guess this strongly depends on the general discussion above.

heat/core/manipulations.py
```diff
@@ -630,6 +637,79 @@ def diagonal(a, offset=0, dim1=0, dim2=1):
...
+            DNDarray([[[ 1,  2,  3],
+                        [ 5,  6,  7],
+                        [ 9, 10, 11]],
+                        [[13, 14, 15],
+                         [17, 18, 19],
+                         [21, 22, 23]]]),
+            DNDarray([])
+        ]
+
+    """
```
ClaudiaComito, line 709: here we should raise an exception if `ary.ndim < 3`

ClaudiaComito (changes requested):
Hi Lena,

before I look into the details, something important jumps to my eye: `np.split` & co.  return lists of subarrays **as views** into the original array. The current heat implementation also says so, but really it returns copies. To exemplify this:
```
>>> import numpy as np
>>> import heat as ht
>>> a = np.arange(16).reshape(4,4)
>>> a
array([[ 0,  1,  2,  3],
       [ 4,  5,  6,  7],
       [ 8,  9, 10, 11],
       [12, 13, 14, 15]])
>>> b, c = np.split(a, 2)
>>> b
array([[0, 1, 2, 3],
       [4, 5, 6, 7]])
>>> c
array([[ 8,  9, 10, 11],
       [12, 13, 14, 15]])
>>> b[0,0] = 10
>>> b
array([[10,  1,  2,  3],
       [ 4,  5,  6,  7]])
>>> a
array([[10,  1,  2,  3],
       [ 4,  5,  6,  7],
       [ 8,  9, 10, 11],
       [12, 13, 14, 15]])
>>> ht_a = ht.arange(16).reshape((4,4))
>>> ht_a
DNDarray([[ 0,  1,  2,  3],
          [ 4,  5,  6,  7],
          [ 8,  9, 10, 11],
          [12, 13, 14, 15]], dtype=ht.int32, device=cpu:0, split=None)
>>> ht_b, ht_c = ht.split(ht_a, 2)
>>> ht_b
DNDarray([[0, 1, 2, 3],
          [4, 5, 6, 7]], dtype=ht.int32, device=cpu:0, split=None)
>>> ht_c
DNDarray([[ 8,  9, 10, 11],
          [12, 13, 14, 15]], dtype=ht.int32, device=cpu:0, split=None)
>>> ht_b[0,0] = 10
>>> ht_b
DNDarray([[10,  1,  2,  3],
          [ 4,  5,  6,  7]], dtype=ht.int32, device=cpu:0, split=None)
>>> ht_a
DNDarray([[ 0,  1,  2,  3],
          [ 4,  5,  6,  7],
          [ 8,  9, 10, 11],
          [12, 13, 14, 15]], dtype=ht.int32, device=cpu:0, split=None)
```

heat/core/manipulations.py
```diff
@@ -2007,6 +2164,258 @@ def sort(a, axis=None, descending=False, out=None):
...
+    else:
+        raise TypeError(
+            "Expected `indices_or_sections` to be array_like (DNDarray, list or tuple), but was {}".format(
+                type(indices_or_sections)
+            )
+        )
+
+    # start of actual algorithm
+
+    if ary.split == axis and ary.split is not None and ary.comm.size > 1:
```
ClaudiaComito, line 2282: you can test the last two conditions with `ary.is_distributed()`

ClaudiaComito:
> this all looks good to me. However, there is a larger comment to be discussed.
> 
> Do we want to have this return actual views of the DNDarray? If so, it would mean that we cannot balance the DNDarrays, and it might change some of the logic involved in the function itself. @ClaudiaComito @Markus-Goetz , thoughts?

I'm with you @coquelin77, no balancing. We need `ht.view`

coquelin77:
> > this all looks good to me. However, there is a larger comment to be discussed.
> > Do we want to have this return actual views of the DNDarray? If so, it would mean that we cannot balance the DNDarrays, and it might change some of the logic involved in the function itself. @ClaudiaComito @Markus-Goetz , thoughts?
> 
> I'm with you @coquelin77, no balancing. We need `ht.view`

this isnt a major change from the code, it could be done with relative ease if the balance is dropped. however, if the hard splits result in a partial view on two processes, this would be hard to communicate to users

Markus-Goetz:
@coquelin77 @ClaudiaComito For now it should return a copy, generally however I would obviously like to have something like a `ht.view` of the data, especially silent return of it. I am not 100% sure whether we can actually achieve this easily, due to the additional properties that we carry around and would have to watch out for. Yet, if somebody would successfully implement this, that would be a solid success for squeezing out further performance

heat/core/manipulations.py
lenablind: If that is the case, an exception is raised within `split`, so I think raising an additional one in `dsplit` would be superfluous

heat/core/manipulations.py
```diff
@@ -630,6 +637,79 @@ def diagonal(a, offset=0, dim1=0, dim2=1):
...
+        If indices_or_sections is an integer, N, the DNDarray will be divided into N equal DNDarrays along the 3rd axis.
+        If such a split is not possible, an error is raised.
+        If indices_or_sections is a 1-D DNDarray of sorted integers, the entries indicate where along the 3rd axis
+        the array is split.
+        If an index exceeds the dimension of the array along the 3rd axis, an empty sub-array is returned correspondingly.
+
+    Returns
+    -------
+    sub_arrays : list of DNDarrays
+        A list of sub-DNDarrays as views into ary.
```
coquelin77, line 658: copies not views, also ary an be a code block (`ary`)

also can you add the copy statement to the description of the function as well. i think that people will often only read that part

coquelin77 (changes requested):
looks like the comment about views vs copies didnt make it into the docs of `dsplit`, `hsplit`. or `vsplit`

heat/core/manipulations.py
lenablind: Oh dear, sorry I missed that. I'll fix it and wrap `ary` into a code block.  Are all markdown features available in the dosctrings/correctly rendered in sphinx? And you're right, the general description is probably the most relevant part to most, I'll add the hint

lenablind:
rerun tests

coquelin77:
rerun tests

coquelin77 (approved):
looks like the comment about views vs copies didnt make it into the docs of `dsplit`, `hsplit`. or `vsplit`

coquelin77:
> looks like the comment about views vs copies didnt make it into the docs of `dsplit`, `hsplit`. or `vsplit`

sorry this was a remnant of before, ignore this

Issue #547, opened 2020-04-14
Implement histogram


ClaudiaComito:

mtar:
The current ht.histogram() function calls the ht.histc() function which uses PyTorch. It should work more like NumPy instead.

ClaudiaComito:
Closing this as it is addressed in #954

PR #679, opened 2020-10-02
Features/547 histc

mtar:
## Description

Introduction of a histogram function. It adds the histograms returned by torch.histc() from all processes.

Note: This function is equivalent to torch.histc(). A NumPy like histogram function still needs to be implemented as it offers much more options.

Issue/s resolved: #547 (partially)

## Changes proposed:
- histc function

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/statistics.py
+++ b/heat/core/statistics.py
@@ -21,6 +21,7 @@ __all__ = [
     "average",
     "bincount",
     "cov",
+    "histc",
     "kurtosis",
     "max",
     "maximum",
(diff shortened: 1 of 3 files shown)


heat/core/statistics.py
```diff
@@ -515,6 +516,67 @@ def cov(m, y=None, rowvar=True, bias=False, ddof=None):
     return c
 
 
+def histc(input, bins: int = 100, min: int = 0, max: int = 0, out=None):
```
coquelin77, line 519: can you add an alias for `histogram` to allow for numpy API compatibility? (also i think its easier for users)

coquelin77 (changes requested):
once that is done this looks fine to me.

more features can be added later but this is great so far.

coquelin77 (approved):
good work. thanks for the new feature :D

Issue #538, opened 2020-04-09
Introduce larray property


Markus-Goetz:
**Feature functionality**
Introduce the larray property that allows access to the underlying torch tensor of a DNDarray.

**Additional context**
Replace the internal member access _DNDarray__array using the new property.
Add a big fat warning, that manipulating the local contents of a DNDarray needs to be done with care and might corrupt/invalidate the metadata in a DNDarray instance.

lenablind:
@Markus-Goetz The only valid dtype for setting _DNDarray__array should therefore be a torch tensor, right?

PR #680, opened 2020-10-06
Features/538 larray

lenablind:
## Description

Introduction of property `larray`, which allows access to the underlying local torch tensor (`_DNDarray__array` / `__array` attribute of `DNDarray`).
For the setter, I included a warning as requested in the issue.

Consequently, I replaced all usages of `_DNDarray__array` within the project space with `larray`. 

As I specified torch.tensor as the only valid dtype for this attribute, I had to adapt function `average` in statistics, simply wrapping a float assignment into a tensor containing that specific float (just one line (328) modified).

Issue/s resolved: #538

## Changes proposed:
- Introduction of property `larray` in dndarray
- replacement of all usages of `_DNDarray__array` with `larray` within the project space
- Corresponding tests (especially for the setter)
- Minimal adaption of function `average` in statistics (line 328)

## Type of change

## Due Diligence

no (, not after the presented modification of `average` in statistics. Without the wrapping, a TypeError is thrown as specified in the setter of `larray`, which expected a torch.tensor but got a float (in the corresponding tests) instead).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Pending additions
 ## New features
+- [#680](https://github.com/helmholtz-analytics/heat/pull/680) Introduction of property larray
 ### Manipulations
 ### Statistical Functions
 ### Linear Algebra
--- a/heat/cluster/kmedians.py
+++ b/heat/cluster/kmedians.py
@@ -58,7 +58,7 @@ class KMedians(_KCluster):
             # Remove 0-element lines to avoid spoiling of median
             assigned_points = X * selection
             rows = (assigned_points.abs()).sum(axis=1) != 0
-            local = assigned_points._DNDarray__array[rows._DNDarray__array]
+            local = assigned_points.larray[rows._DNDarray__array]
             clean = ht.array(local, is_split=X.split)
             clean.balance_()
             # failsafe in case no point is assigned to this cluster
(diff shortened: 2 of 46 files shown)


Markus-Goetz:
Outstanding job Lena. As far as I can see, our core is now issuing a lot of warning. As per issue that was asked for. However, I feel, that I should modify this as to: if the user is using the larray setter a warning should be issued, if the core is using larray, then warnings should be suppressed, as we are supposedly know what we are doing. Do you think you can change it to be that way? Or do you have differing thoughts?

lenablind:
> Outstanding job Lena. As far as I can see, our core is now issuing a lot of warning. As per issue that was asked for. However, I feel, that I should modify this as to: if the user is using the larray setter a warning should be issued, if the core is using larray, then warnings should be suppressed, as we are supposedly know what we are doing. Do you think you can change it to be that way? Or do you have differing thoughts?

Thank you, Markus. Yes, I agree, this sounds reasonable to me. I'll try to change it to be that way.

coquelin77 (approved):

coquelin77:
@lenablind can you handle the merge here? there is some minor conflicts with the `nbytes` feature

lenablind:
> @lenablind can you handle the merge here? there is some minor conflicts with the `nbytes` feature

@coquelin77 Yes for sure! Thank you for taking care of this and the previous PRs

coquelin77 (approved):

Issue #682, opened 2020-10-09
`Reshape` for distributed boolean DNDarray


lenablind:
**Description**
A distributed DNDarray with dtype = ht.bool can't be reshaped.
Instead, a RuntimeError is thrown in line 1379 of `manipulations.py` assuming the size of the input to be 1 (no matter how large it actually is) and therefore refusing the reshape.

**To Reproduce**
1. Which module/class/function is affected?
`manipulations.reshape`
2. What are the circumstances under which the bug appears?
Trying to reshape a DNDarray with the following attributes:
- `dtype=ht.bool`
- `split is not None`
3. What is the exact error message / erroneous behavior?
RuntimeError in line 1379 of `manipulations.py`:
`shape '[2, 2]' is invalid for input of size 1`

**Expected behavior**
The DNDarray to be reshaped as defined.

**Illustrative**
If applicable, add screenshots or minimal examples to help explain your problem.
`>>>  x = ht.arange(4, split=0, dtype=ht.bool)`
`>>> x`
`DNDarray([False,  True,  True,  True], dtype=ht.bool, device=cpu:0, split=0)`
` >>> x.reshape((2,2))`
`RuntimeError: shape '[2, 2]' is invalid for input of size 1`

**Version Info**
 Which version are you using?
0.5.0

Issue #654, opened 2020-08-08
Bug in Reshape


Inzlinger:
**Description**
MPI.Exception : MPI_ERR_TRUNCATE on reshape.

**To Reproduce**
Run with n>1 threads:
```
import heat as ht

source_shape = (1600080, 17)
target_shape = (20001, 80, 17)

a = ht.zeros(source_shape, split=0)

print(a.lshape, a.gshape)
a = ht.reshape(a, target_shape)
print(a.lshape, a.gshape)

```

**Expected behavior**
Reshape to target shape.

**Version Info**
Heat on Branch Master, last commit 9b1f28bc9db00001085e26cb764b7de0714ffea1

**Additional comments**
Tried some different numbers for the shapes but did not find any meaningful pattern.

mtar:
Does #684 fix the problem?

PR #684, opened 2020-10-09
fix reshape boolean bug

mtar:
## Description

Fixes the reshape bug on booleans and on floats in some cases. Some tensors were using the dtype of the input by mistake. 

Issue/s resolved: #682 
Issue/s resolved: #654

## Changes proposed:
- fix reshape on booleans

## Type of change

Bug fix

## Due Diligence

no


lenablind (approved):
Looks good to me - Thank you for fixing the bug!

Issue #681, opened 2020-10-08
Implement a `dndarray.balanced` bool attribute


ClaudiaComito:
- it would be set, mostly to `True`, by factories and operations 
- it would save us a ton of `is_balanced()` communication
- `comm.chunk()` would work as usual `if self.balanced`, i.e. in the vast majority of cases, otherwise collect and return information on local data slices
- getters and setters could return the correct slice no matter what (see #668)
- no need for communication for create_lshape_map when `dndarray.balanced == True`

Issue #668, opened 2020-09-09
__getitem__  and __setitem__ assume DNDarray is balanced


ClaudiaComito:
**Description**
`dndarray.__getitem__` and `dndarray.__setitem__` rely on  `comm.chunk()` to provide the information about what slice of the data is on each process. However `comm.chunk()` only calculates such slice, on the assumption that the underlying array is balanced. 

A very common use case for working with unbalanced dndarrays is `ht.where()`. The result will almost always be unevenly distributed, and we want to keep it that way because it matches the relevant data on the node.

Here's an example of the problem (`mpirun -n 2`):
```
ht_mask_matrix = ht.random.randint(0,2,(13,10), split=0)
ht_where = ht.where(ht_mask_matrix > 0)

>> [1,1]<stdout>:ht_where.gshape, ht_where.lshape =  (70, 2) (32, 2)
>> [1,0]<stdout>:ht_where.gshape, ht_where.lshape =  (70, 2) (38, 2) <---

ht_where_x, ht_where_y = ht_where[:, 0], ht_where[:, 1]

>> [1,1]<stdout>:ht_where_x.lshape =  (32,)
>> [1,0]<stdout>:ht_where_x.lshape =  (35,) <---
```

**To Reproduce**
see above

**Expected behavior**
The getters and setters should know know the actual slice of data are on each node. The least disruptive solution for me would be to have `comm.chunk()` check for balance, and collect the actual chunking information (instead of calculating it) if the array is unbalanced.

**Version Info**
0.4. current main branch

ClaudiaComito:
@coquelin77 on MM points to the communication overhead that would arise by checking for balance every time we run comm.chunk().

I'm thinking: how about a `dndarray.balanced` boolean attribute?  
- it would be set, mostly to `True`, by factories and operations (notable exceptions: `ht.where`, `ht.flatten` maybe? @lenablind )
- it would save us a ton of `is_balanced()` communication
- `comm.chunk()` would work as usual `if self.balanced`, i.e. in the vast majority of cases, otherwise collect and return information on local data slices
- getters and setters could return the correct slice no matter what

Thoughts?

lenablind:
> @coquelin77 on MM points to the communication overhead that would arise by checking for balance every time we run comm.chunk().
> 
> I'm thinking: how about a `dndarray.balanced` boolean attribute?
> 
[...]

PR #687, opened 2020-10-26
Features/681 balanced

ClaudiaComito:
## Description

Issue/s resolved: #681  (see discussion in #668 )

## Changes proposed:
- New dndarray property `balanced`, can be`True` or `False`, or `None` if not enough info.
- `balanced` is set internally within the factories and cannot be specified by the user. It is always set to `True` except by `ht.array` where  `is_split is not None`, in which case the balanced status will be assessed (adds one more communication step to `ht.array()`, but eliminates the need for possibly repeated `is_balanced()` checks later).
- the dndarray method `is_balanced` now returns `self.balanced`, unless `self.balanced is None`, in which case it will perform the check as usual and set the `balanced` property.
- One major problem is our extended use of the construct `x.larray = new_local_torch_x` in code and tests. Very often `new_local_torch_x.shape != x.larray.shape` and results into a mess, among other things invalidating the `balanced` information.  I've introduced the following changes:
  - added a "local shape sanitation" to `dndarray.larray.setter`. To me it's still too lax. If not called from all processes, it will produce an imbalanced dndarray with inhomogeneous balanced status.    
  - replaced instances of brute-force larray-setting with `ht.array(new_local_torch_x, is_split=...)` (notably `concatenate`, `argmin` and `argmax`)
  - while I was at it, modified `__reduce_op` to return the correct `argmax`/ `argmin` result directly

- I've also expanded the sanitation module with `sanitize_in_tensor` (verify that input is torch.Tensor), `sanitize_lshape` (for larray.setter, see above), and I've modified `constants.sanitize_infinity` to use torch.finfo/iinfo and moved it to`sanitation.sanitize_infinity`.

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -1804,8 +1804,8 @@ class DNDarray:
         balanced = 1 if test_lshape == self.lshape else 0
 
         out = self.comm.allreduce(balanced, MPI.SUM)
-        self.__balanced = True if out == self.comm.size else False
-        return self.balanced
+        balanced = True if out == self.comm.size else False
+        return balanced
 
     def is_distributed(self):
         """
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -76,6 +76,7 @@ class DNDarray:
         self.__split = split
         self.__device = device
         self.__comm = comm
+        self.__balanced = False
         self.__ishalo = False
         self.__halo_next = None
         self.__halo_prev = None
@@ -89,12 +90,8 @@ class DNDarray:
             self.__array = self.__array.to(devices.sanitize_device(self.__device).torch_device)
 
     @property
-    def halo_next(self):
-        return self.__halo_next
-
-    @property
-    def halo_prev(self):
-        return self.__halo_prev
+    def balanced(self):
+        return self.__balanced
 
     @property
     def comm(self):
@@ -113,6 +110,14 @@ class DNDarray:
         return self.__gshape
 
     @property
+    def halo_next(self):
+        return self.__halo_next
+
+    @property
+    def halo_prev(self):
+        return self.__halo_prev
+
+    @property
     def nbytes(self):
         """
         Equivalent to property gnbytes.
@@ -126,6 +131,16 @@ class DNDarray:
         return self._DNDarray__array.element_size() * self.size
 
     @property
+    def ndim(self):
+        """
+        Returns
+        -------
+        number_of_dimensions : int
+            the number of dimensions of the DNDarray
+        """
+        return len(self.__gshape)
+
+    @property
     def numdims(self):
         """
         Returns
@@ -137,16 +152,6 @@ class DNDarray:
           `numdims` will be removed in HeAT 1.0.0, it is replaced by `ndim` because the latter is numpy API compliant.
         """
         warnings.warn("numdims is deprecated, use ndim instead", DeprecationWarning)
-        return len(self.__gshape)
-
-    @property
-    def ndim(self):
-        """
-        Returns
-        -------
-        number_of_dimensions : int
-            the number of dimensions of the DNDarray
-        """
         return len(self.__gshape)
 
     @property
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -165,7 +165,6 @@ class DNDarray:
                 "might corrupt/invalidate the metadata in a DNDarray instance"
             )
         self.__array = array
-        # TODO: implement __update_metadata() for gshape, lshape, dtype
 
     @property
     def nbytes(self):
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -2628,10 +2628,7 @@ class DNDarray:
                 ]
             self.__balanced = True
         else:
-            if not isinstance(target_map, torch.Tensor):
-                raise TypeError(
-                    "target_map must be a torch.Tensor, currently {}".format(type(target_map))
-                )
+            sanitation.sanitize_in_tensor(target_map)
             if target_map[..., self.split].sum() != self.shape[self.split]:
                 raise ValueError(
                     "Sum along the split axis of the target map must be equal to the "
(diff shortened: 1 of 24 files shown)


Markus-Goetz:
Why do we introduce this property?

`is_balanced` should not have side-effects. It should be a pure informative call.

ClaudiaComito:
> Why do we introduce this property?
> 
> `is_balanced` should not have side-effects. It should be a pure informative call.


Hi Markus, I forgot to link the original problem, I've added it now (#668). Let me know if I should expand more.

heat/core/dndarray.py
```diff
@@ -1769,12 +1792,16 @@ def is_balanced(self):
         balanced : bool
             True if balanced, False if not
         """
+        if self.balanced is not None:
+            return self.balanced
```
coquelin77, line 1796: is it worth doing the check here? perhaps this function should actually check every time. if we want to avoid the check i think we would use `if a.balanced:` instead of `if a.is_balanced()`. although this seems like its splitting hairs

heat/core/factories.py
```diff
@@ -374,10 +377,20 @@ def array(
...
         split = is_split
+        # compare to calculated balanced lshape (cf. dndarray.is_balanced())
+        gshape = tuple(int(ele) for ele in gshape)
+        lshape = tuple(int(ele) for ele in lshape)
+        _, _, chk = comm.chunk(gshape, split)
+        test_lshape = tuple([x.stop - x.start for x in chk])
+        match = 1 if test_lshape == lshape else 0
+        gmatch = comm.allreduce(match, MPI.SUM)
+        if gmatch != comm.size:
+            balanced = False
```
coquelin77, line 388: i like this logic. its clean and neat.

heat/core/tests/test_arithmetics.py
```diff
@@ -511,7 +511,7 @@ def test_prod(self):
         self.assertEqual(shape_noaxis_split_axis_neg_prod.larray.dtype, torch.float32)
         self.assertEqual(shape_noaxis_split_axis_neg_prod.split, 1)
 
-        out_noaxis = ht.zeros((1, 2, 3, 5))
+        out_noaxis = ht.zeros((1, 2, 3, 5), split=1)
```
coquelin77, line 514: was this changed for completeness or another reason?

coquelin77 (commented):
looks good to me. i left a few questions in the code.

also, why did you change some of the test case to add an axis? where the other cases already covered? (i didnt look to deeply into this)

heat/core/dndarray.py
ClaudiaComito: I started out like you say, and then I found myself writing `if a.balanced is None, do a.is_balanced()` too many times.

I do agree though that there should be a way to force the check. I was thinking `a.is_balanced(force=True)`

heat/core/dndarray.py
```diff
@@ -152,6 +165,7 @@ def larray(self, array):
                 "might corrupt/invalidate the metadata in a DNDarray instance"
             )
         self.__array = array
+        # TODO: implement __update_metadata() for gshape, lshape, dtype
```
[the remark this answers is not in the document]
ClaudiaComito: I would absolutely be for restricting larray.setter to same shape/dtype. In which case we can also call it `tensor()` and make users happy.

heat/core/factories.py
ClaudiaComito: yep, it's yours! 😎

heat/core/tests/test_arithmetics.py
ClaudiaComito: This is because with the introduction of the sanitation module, we're a bit stricter on the output buffer, where gshape, dtype, split and device must match the result (`sanitation.sanitize_out`). I haven't been systematic in switching to the sanitation module so far, I do it when I happen to see it can be done quickly. In that case, I often have to modify the tests to define the output buffer accordingly.

heat/core/factories.py
coquelin77: :see_no_evil: i dont remember writing it and i just assumed that you did that

heat/core/dndarray.py
coquelin77: maybe check instead of force? or better yet, `force_check`?

heat/core/dndarray.py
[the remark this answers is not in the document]
coquelin77: lets stick with `larray` so that users get used to the nomenclature used in other places.

would that restriction cause a lot of problems?

heat/core/tests/test_arithmetics.py
coquelin77: ah okay. i just wanted to make sure that the default behaviors still returned the expected results

heat/core/dndarray.py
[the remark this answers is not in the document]
ClaudiaComito: I don't think so, I've removed the brute-force larray.setter code already.

heat/core/dndarray.py
ClaudiaComito: Yes, `force_check` sounds perfect

Markus-Goetz:
@ClaudiaComito I understand the issue now. However, shouldn't setitem and getitem force consistency in this case? Moreover, I am still maintaining that a purely informative check called `is_balanced` should not cause side effects, i.e. occasionally do something behind the scenes.

ClaudiaComito:
> @ClaudiaComito I understand the issue now. However, shouldn't setitem and getitem force consistency in this case? Moreover, I am still maintaining that a purely informative check called `is_balanced` should not cause side effects, i.e. occasionally do something behind the scenes.

Oh, I see what you mean re: is_balanced. Yes, you're right. I'll remove the property setting then, but are you OK with the `force-check`solution? (might be redundant if we get stricter about what `larray.setter` is allowed to do, see discussion above).

Regarding setitem and getitem. My idea was to have `chunk` check the `balanced` property and, if False, return the "measured" result instead of the theoretical one based on the `balanced=True` assumption. In that case, setitem and getitem would automatically get the correct result from `chunk` no matter what.

coquelin77 (approved):

PR #693, opened 2020-11-06
Fix clustering example not compilable

sebimarkgraf:
## Description
Changes the clustering example of the documentation to compile.

## Changes proposed:
- Introduce variable R for radius of the circle
- Use triple string

## Type of change
- Documentation update

## Due Diligence
Doc update. Seems to be not included in changelog so far

no

skip ci


mtar:

Cdebus (approved):
Nice catch, thank you

Issue #692, opened 2020-10-28
support for complex numbers


mtar:
Heat does not work with complex numbers.

**Additional context**
```
ht.array([1j])
Traceback (most recent call last):
  File "/h/m/P/h/o/lib/python3.8/site-packages/heat/core/types.py", line 304, in canonical_heat_type
    return __type_mappings[a_type]
KeyError: torch.complex64

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/h/m/P/h/o/lib/python3.8/site-packages/heat/core/factories.py", line 317, in array
    dtype = types.canonical_heat_type(obj.dtype)
  File "/h/m/P/h/o/lib/python3.8/site-packages/heat/core/types.py", line 306, in canonical_heat_type
    raise TypeError("data type {} is not understood".format(a_type))
TypeError: data type torch.complex64 is not understood
```

PR #699, opened 2020-11-17
complex data type

mtar:
## Description

This PR adds `complex64` and `complex128` as new dtypes for complex numbers. It also adds `real()`, `imag()`, angle(), and `conjunction()`.

Issue/s resolved: #692 

## Changes proposed:
- complex64 & complex128 and aliases
- real(), imag()
- conjunction(), conj()
- angle()

## Type of change

## Due Diligence

yes / no


heat/core/communication.py
```diff
@@ -61,6 +61,8 @@ class MPICommunication(Communication):
         torch.int64: MPI.LONG,
         torch.float32: MPI.FLOAT,
         torch.float64: MPI.DOUBLE,
+        torch.complex64: MPI.COMPLEX,
+        torch.complex128: MPI.DOUBLE_COMPLEX,
```
coquelin77, line 65: does torch have half complex?

coquelin77 (approved):

heat/core/communication.py
mtar: pytorch has complex32, but we don't support 16 bit floats

coquelin77: we could add that to the comms package, its a small change. but that is not related to this PR

coquelin77 (approved):

PR #720, opened 2021-02-03
modify code coverage

mtar:
## Description

Issue/s resolved: #

## Changes proposed:
-
-
-
-

## Due Diligence

yes / no

skip ci


mtar:
sd

mtar:
b

bhagemeier (approved):
I generally agree with the changes. However, we may need to adjust the margin to be somewhat tighter than 5%. I assume that the drop by 2.75% in this case is due to the fact that all tests, which are covered to 100% by nature, have been removed from the coverage. OTOH, an addition of roughly 5% untested code is difficult to achieve the larger the project grows.

PR #722, opened 2021-02-08
Lanczos Bugfix in Solver

Cdebus:
Reverted inplace calculation from commit 80577f70cf7c91822569288817296bfd858cff73, because this was causing incorrect eigenvalue calculations

## Description
Spectral clustering was yielding negative Eigenvalues, which cannot be due to the Laplacian being a positive semidefinite matrix. It turns out that a change in the solver class to inplace operations for updating the w vector caused wrong calculations. 

## Changes proposed:
-```w -= alpha * vi - beta * V[:, i - 1]``` -->  ``` w = w - alpha * vi - beta * V[:, i - 1]``` 

## Type of change

 no


mtar:
run tests

coquelin77 (approved):

PR #723, opened 2021-02-09
Enhancement/ci venv

mtar:
This change uses the same python virtual environment for all branches. This step was necessary due to the large size of the torch package (~1.5GB) which filled too much space before when the environment was created separately.


mtar:
rerun tests

Issue #721, opened 2021-02-08
Reduce number of warnings in unit tests


mtar:
The test report is cluttered with warning messages. Many come directly from the functions that are tested. 

Use assertWarns (https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns) where a warning is expected.

PR #725, opened 2021-02-15
Enhancement/721 warnings

mtar:
## Description

Reduces the number of expected warnings on unit tests

Issue/s resolved: #721 

## Changes proposed:
-
-
-
-

## Type of change

Bug-fix

## Due Diligence

no


heat/core/types.py
```diff
@@ -286,7 +286,7 @@ def char(cls):
     "c8": complex64,
     "c16": complex128,
     # numpy types
-    np.bool: bool,
+    np.bool_: bool,
```
coquelin77, line 289: what is the difference with this vs `np.bool`?

mtar: np.bool is deprecated. builtin `bool` or `np.bool_`should be used now.

coquelin77: okay cool. this im fine with all of this

coquelin77 (approved):

Issue #730, opened 2021-03-08
Wrong split axis after summing along axis or indexing


ben-bou:
**Description**
The split axis is not updated after sum along axis or indexing with an int.

**To Reproduce**
1. Which module/class/function is affected?
     - `DNDarray`
2. What are the circumstances under which the bug appears?
    - sum along axis
    - indexing with an int
3. What is the exact error message / erroneous behavior?
    - wrong split axis 

**Illustrative**
If applicable, add screenshots or minimal examples to help explain your problem.

For summing along axis 0:
```python
In [1]: import heat as ht
                                                                                                                                                                                
In [2]: a = ht.ones((2,2), split=1)
                                                                                                                                                                     
In [3]: a.sum(0).shape, a.sum(0).split  # wrong split axis; dimension 1 does not exist, should be 0                                                                                                                                                                    
Out[3]: ((2,), 1)
```

Regarding the indexing:
```python
In [1]: import heat as ht                                                                                                                                                                                   

In [2]: a = ht.ones((1,2,3), split=1)                                                                                                                                                                       

In [3]: a.shape, a.split  # a is split along the dimension with two elements                                                                                                                                                                                  
Out[3]: ((1, 2, 3), 1)

In [4]: a[0,:,:].shape, a[0,:,:].split  # this is wrong; a[0] is not split along the dimension with three elements                                                                                                                                                                     
Out[4]: ((2, 3), 1)

In [5]: a[:,0,:].shape, a[:,0,:].split  # wrong again                                                                                                                                                                    
Out[5]: ((1, 3), 1)

In [6]: a[:,:,0].shape, a[:,:,0].split  # correct                                                                                                                                                                    
Out[6]: ((1, 2), 1)
```

**Expected behavior**

For example the `mean` method works fine.
```python
In [4]: a.mean(0).shape, a.mean(0).split  # legitimate  split axis                                                                                                                                                                
Out[4]: ((2,), 0)
```

**Version Info**
 `0.5.1`

**Additional comments**
The issue does **not** occur when using `split=0`.
Am I doing something wrong?

ClaudiaComito:
Thanks @ben-bou for bringing this up!

ben-bou:
Thanks @ClaudiaComito  and @coquelin77 for handling this!

PR #732, opened 2021-03-09
DNDarray getitem: corrected split logic when ints in key

coquelin77:
## Description

Addressed getitem split bug

Issue/s resolved: #730 

## Changes proposed:
- change in logic for the new split calculation in getitem when ints are in the key

## Type of change

## Due Diligence

It might need change things which relied upon the buggy code. however the tests run clean, so hopefully this just wasnt noticed previously.


mtar (approved):

ClaudiaComito (approved):
Great, thanks for this.
I've fixed the problem for reductions operations in #744.

ClaudiaComito (changes requested):
@coquelin77 I just realized  both #730 fixes  (#732 and former #744) must branch off 0.5.x, not off the main branch. We can then release the bug fixes right away without waiting for 0.6.0.

ClaudiaComito:
> @coquelin77 I just realized both #730 fixes (#732 and former #744) must branch off 0.5.x, not off the main branch. We can then release the bug fixes right away without waiting for 0.6.0.

Alright, I think I'm taking this back. The #744 fix already uses some of the 0.6 features and I'm not sure walking the code back to 0.5 state is worth the time. I'll reopen #744 as is and reapprove this one as is as well.

ClaudiaComito (dismissed):

coquelin77:
the master merge cleared the approval from @ClaudiaComito but i dont think that we need to go through it again. i will merge it once the tests run throguh

Issue #737, opened 2021-03-12
implement isreal / iscomplex


mtar:
Check for real or complex elements

PR #738, opened 2021-03-12
Feature/737 isreal iscomplex

mtar:
## Description

Add iscomplex() and isreal() functionality that checks element-wise for complex or real values.

Issue/s resolved: #737 

## Changes proposed:
- New feature iscomplex
- New feature isreal

## Type of change

## Due Diligence

no


heat/core/types.py
```diff
@@ -607,6 +611,42 @@ def can_cast(from_, to, casting="intuitive"):
...
+    --------
+    >>> ht.iscomplex(ht.array([1+1j, 1]))
+    DNDarray([ True, False], dtype=ht.bool, device=cpu:0, split=None)
+    """
+    sanitation.sanitize_in(x)
+
+    if issubclass(x.dtype, _complexfloating):
+        return x.imag != 0
+    else:
+        return factories.zeros(x.shape, bool, split=x.split, device=x.device, comm=x.comm)
```
coquelin77, line 632: why not use `torch.is_complex`?

coquelin77 (changes requested):

heat/core/types.py
mtar: I did not find it :mag: :smile:

mtar: torch.is_complex returns the result for the whole tensor, not element-wise

mtar:
run tests

heat/core/types.py
coquelin77: wow. that is a really annoying difference. but that is what it is

coquelin77 (approved):

PR #749, opened 2021-03-24
Features/unique sort distributed

### LINKED ISSUE #363: vectorized sorting
**Related**
When specifying an axis for the `unique` function, the array is searched for unique vectors along this axis. However a sorted result is difficult to compute when `split != axis`.
Because this feature might also be interesting on its own it would make sense to implement it as a standalone function.

**Feature functionality**
Compared to the classic sort the vectorized sorting should keep the values in one vector consistent and only sort the vectors along the sorting axis. The sorting should give the highest priority for the first value, if on this level there are equal values the next value should be compared and so on.

**Additional context**
Numpy uses some kind of vectorized sorting in their `unique` function when specifying the `axis` argument.

### LINKED ISSUE #564: Add clarification to documentation of unique()
**Feature functionality**
Clarify that the order in res is either unstable or how it relates to split/axis.
Not sure whether this is a bug or works as intended; the order of res is different depending on the split. 

https://github.com/helmholtz-analytics/heat/blob/a1c3792e6e32b3bf7ca4953b6afbc1d53759a031/heat/core/manipulations.py#L1319

**Additional context**
Example: 
>>> import heat as ht
>>> a = ht.array([1,2,3])
>>> a
tensor([1, 2, 3])
>>> ht.unique(a)
tensor([3, 2, 1])
>>> b = ht.array([1,2,3], split=0)
>>> b
tensor([1, 2, 3])
>>> ht.unique(b)
tensor([1, 2, 3])

### LINKED ISSUE #621: Unique() inconsistencies
**Description**
While looking into #564, I found a number of inconsistencies in ht.unique().

- #564 is no documentation issue. 
   - The kwarg `sorted` is set to False by default for ht.unique(), but it's True by default for torch.unique(). Currently,  `ht.unique(a)` results in a sorted DNDarray if `a.split=None` (pure torch implementation: `ht.unique(a) = torch.unique(a._DNDarray__array)`), whereas if `a.split` is not None, the result will not be sorted (heat implementation).
  - Replacing `ht.unique(a) = torch.unique(a._DNDarray__array)` with 
`ht.unique(a) = torch.unique(a._DNDarray__array, sorted=sorted)` doesn't help, because `sorted=False` means different things for heat and torch: 
      - heat interpretation: leave result unsorted; 
      - torch interpretation: leave result REVERSE SORTED. See discussion on #564 for an example.
   I propose setting `sorted=True` by default in ht.unique() as at the moment it's the only way to prevent inconsistencies with torch, **although** I'm aware that the sorting comes with significant overhead. Incidentally, numpy.unique() returns the "sorted unique elements of an array" and not sorting is not even an option. 

- if `return_inverse=True`, ht.unique() by design returns a list of one DNDarray (the unique elements) and one torch tensor (the inverse indices). Should be two DNDarrays.

- it is currently not possible to run `ht.unique(a, sorted=True, axis=axis)` if `axis != split`. Error message: 
   > Sorting with axis != split is not supported yet. See https://github.com/helmholtz-analytics/heat/issues/363

   This needs to be followed up.

**To Reproduce**
1. Which module/class/function is affected?
manipulations.unique()

2. What are the circumstances under which the bug appears?
see above

3. What is the exact error message / erroneous behavior?
see above

**Version Info**
 current main branch

ClaudiaComito:
## Description

This PR introduces major changes in the `ht.unique()` implementation, fixing some bugs/inconsistencies along the way (see below).

## Changes proposed:
Distributed `unique` requires two passes: 
1) find local sorted unique elements, 
2) find global sorted unique elements. 

The current (v0.5.1) implementation solves step 2. by running `torch.unique` again on the gathered local unique elements. This might turn into a memory bottleneck for very large data.
 
The main implementation change in this PR is that, in the distributed case,  `ht.unique` now recycles the "pivot sorting" implementation (see `ht.sort()`, `manipulations._pivot_sorting()`) to perform an `Alltoallv`-based sorted unique operation that doesn't require "gathering". 

The main user-side changes are as follows:
- `ht.unique` now, like numpy, always returns the SORTED unique elements.

- "sparse" vs. "dense" unique. If the collective size of the local uniques (from step 1) above) is smaller than the size of the local data, then `ht.unique` gathers everything and runs the operation locally. In this case, the unique elements array will have `split=None`. Otherwise, distributed `unique` via `_pivot_sorting()` (`Alltoallv`) returning a distributed DNDarray.
- inverse indices are now a DNDarray and distributed like the input data. Note that  inverse indices are used  to recreate the original data shape from the unique elements. However, the sorted unique elements corresponding to a given inverse index might be on a different process. Eventually, `setitem` should be able to deal with this, at the moment `unique[inverse]` requires a `unique.resplit_(None)` first.

As an aside:
- get `gethalo` to work on imbalanced DNDarrays
- get `create_lshape_map` to only require communication for imbalanced DNDarrays
- resolve race condition in `test_qr` that has been  popping up on and off for ages.
- **ADDED 24 NOV 2021**:  `factories.array` behaviour when `copy=False` now closer to `np.array` (https://numpy.org/doc/stable/reference/generated/numpy.array.html).  

Issue/s resolved: #363, #564, #621

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected):
   - `ht.unique()` always returns the sorted unique elements, kwarg`sorted` has been removed
   - inverse indices are no longer torch tensors, they're now DNDarrays and distributed like the input
   - `unique.resplit_(None)` might be required before applying inverse indices 
   - **NEW:** `factories.array(copy=False)` does not copy slices of original data unless absolutely necessary (dtype, order etc.) 

## Due Diligence

- the possibility to leave the data "unsorted" is not available any longer. 
- operations expecting inverse indices to be a  local torch tensor will fail.
- operations expecting `ht.unique()` to return a non-distributed DNDarray may fail in some cases.


heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
         rcounts = recv_matrix
 
         shape = (partition_matrix[rank].max(),) + transposed.size()[1:]
-        first_result = torch.empty(shape, dtype=local_sorted.dtype)
+
+    first_result = torch.empty(shape, dtype=local_sorted.dtype)
```
coquelin77, line 2087: device

heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
...
+            idx_slice = [slice(None)]
+        else:
             idx_slice = [slice(None)] + [slice(ind, ind + 1) for ind in idx]
 
-            send_count = scounts[idx_slice].reshape(-1).tolist()
-            send_disp = [0] + list(np.cumsum(send_count[:-1]))
-            s_val = local_sorted[idx_slice].clone()
-            s_ind = actual_indices[idx_slice].clone().to(dtype=local_sorted.dtype)
+        send_count = scounts[idx_slice].reshape(-1).tolist()
+        send_disp = [0] + list(np.cumsum(send_count[:-1]))
```
coquelin77, line 2104: why not torch? i think that `scounts` is already a torch tensor,

heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
...
+        s_val = local_sorted[idx_slice].clone()
 
-            recv_count = rcounts[idx_slice].reshape(-1).tolist()
-            recv_disp = [0] + list(np.cumsum(recv_count[:-1]))
-            rcv_length = rcounts[idx_slice].sum().item()
-            r_val = torch.empty((rcv_length,) + s_val.shape[1:], dtype=local_sorted.dtype)
-            r_ind = torch.empty_like(r_val)
+        recv_count = rcounts[idx_slice].reshape(-1).tolist()
+        recv_disp = [0] + list(np.cumsum(recv_count[:-1]))
```
coquelin77, line 2108: see above

heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
...
+        rcv_length = rcounts[idx_slice].sum().item()
+        r_val = torch.empty((rcv_length,) + s_val.shape[1:], dtype=local_sorted.dtype)
```
coquelin77, line 2110: device

heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
...
-                    # Taking just enough elements from the last element to fill the current processes tensor
-                    send_vec[idx][last][proc] = int(target_cumsum[proc] - current_cumsum[last - 1])
-                    current_counts[last] -= int(target_cumsum[proc] - current_cumsum[last - 1])
-                else:
-                    # process doesn't need more values
-                    send_vec[idx][proc][proc] = (
-                        partition_matrix[proc][idx] - send_vec[idx][proc].sum()
+    if sort_op is torch.unique:
+        # early out for unique
+        return first_result
```
coquelin77, line 2122: question for the future: should `unique` also return the indices?

heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
...
+
+    # The process might not have the correct number of values therefore the tensors need to be rebalanced
+    send_vec = torch.zeros(local_sorted.shape[1:] + (size, size), dtype=torch.int64)
+    target_cumsum = np.cumsum(counts)
```
coquelin77, line 2126: device + why numpy again

heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
...
+    for idx in np.ndindex(local_sorted.shape[1:]):
+        idx_slice = [slice(None)] + [slice(ind, ind + 1) for ind in idx]
+        current_counts = partition_matrix[idx_slice].reshape(-1).tolist()
+        current_cumsum = list(np.cumsum(current_counts))
```
coquelin77, line 2130: same as before. partition matrix is already a torch tensor. here it goes: 
torch tensor -> list -> numpy array -> cumsum -> list
it could go
torch tensor -> cumsum -> list

coquelin77: also, cumsum will need the `dim=0` parameter

heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
...
+        for proc in range(size):
+            # process has to many values which will be sent to higher ranks
+            if current_cumsum[proc] > target_cumsum[proc]:
+                first = next(i for i in range(size) if send_vec[idx][:, i].sum() < counts[i])
+                last = next(
+                    i
+                    for i in range(size + 1)
+                    if i == size or current_cumsum[proc] < target_cumsum[i]
+                )
```
coquelin77, line 2139: doesnt this just take the first value of the generator?

```python
>>> next(i for i in range(10))
0
```

heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
...
+                    send_vec[idx][first + i][proc] = int(x - send_vec[idx][first + i].sum())
+                    current_counts[first + i] = 0
+                # Taking just enough elements from the last element to fill the current processes tensor
+                send_vec[idx][last][proc] = int(target_cumsum[proc] - current_cumsum[last - 1])
+                current_counts[last] -= int(target_cumsum[proc] - current_cumsum[last - 1])
+            else:
+                # process doesn't need more values
+                send_vec[idx][proc][proc] = partition_matrix[proc][idx] - send_vec[idx][proc].sum()
+            current_counts[proc] = counts[proc]
+            current_cumsum = list(np.cumsum(current_counts))
```
coquelin77, line 2173: numpy again. im not sure if it makes a large difference to put it in torch though. that can be said about most of these things to be fair, but if something uses it on a GPU then it needs to be transferred there

heat/core/manipulations.py
```diff
@@ -2074,115 +2083,188 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
...
+
+        send_count = send_vec[idx][rank]
+        send_disp = [0] + list(np.cumsum(send_count[:-1]))
+
+        recv_count = send_vec[idx][:, rank]
+        recv_disp = [0] + list(np.cumsum(recv_count[:-1]))
+
+        end = partition_matrix[rank][idx]
+        s_val, indices = first_result[0:end][idx_slice].sort(descending=descending, dim=0)
+        r_val = torch.empty((counts[rank],) + s_val.shape[1:], dtype=local_sorted.dtype)
```
coquelin77, line 2189: device

heat/core/manipulations.py
```diff
@@ -2779,146 +2903,96 @@ def unique(
...
-                    end = steps[i + 1]
-                    for num, x in enumerate(inverse_buf[begin:end]):
-                        inverse_indices[begin + num] = g_inverse[begin + x]
-
+        lres = torch.unique(local_data, sorted=True, return_inverse=False, dim=unique_axis)
+    gres = factories.array(lres, dtype=a.dtype, is_split=0, device=a.device)
+
+    # calculate size (bytes) of local unique. If less than local_data, gather and run everything locally
+    _, data_max_lshape, _ = a.comm.chunk(a.gshape, a.split, rank=0)
+    data_max_lbytes = torch.prod(torch.tensor(data_max_lshape)) * a.larray.element_size()
```
coquelin77, line 2912: dont need to put everything in tensors here (im not sure about the overhead) but this can also be done with the max `lshape * element_size` i believe that `element_size` returns an int but i am not sure

heat/core/manipulations.py
```diff
@@ -2779,146 +2903,96 @@ def unique(
...
+        gres.larray = lres
+
+    if axis is not None and axis != 0:
+        # transpose back to original
+        gres = linalg.basics.transpose(gres, (axis, 0))
 
-    return_value = result
     if return_inverse:
-        return_value = [return_value, inverse_indices.to(a.device.torch_device)]
+        return (gres, global_inverse)
```
coquelin77, line 2988: unnecessary parentheses

heat/core/manipulations.py
```diff
@@ -3037,7 +3111,14 @@ def resplit(arr: DNDarray, axis: int = None) -> DNDarray:
...
         )
-        counts, displs, _ = arr.comm.counts_displs_shape(arr.shape, arr.split)
+        if arr.is_balanced():
+            counts, displs, _ = arr.comm.counts_displs_shape(arr.shape, arr.split)
+        else:
+            counts = arr.create_lshape_map()[:, arr.split]
+            displs = torch.cumsum(
+                torch.cat((torch.tensor([0], device=counts.device), counts[:-1])), dim=0
+            )
+            counts, displs = tuple(counts.tolist()), tuple(displs.tolist())
```
coquelin77, line 3121: is this undone in another PR?

coquelin77 (changes requested):

coquelin77:
rerun tests

heat/core/manipulations.py
```diff
@@ -2798,146 +2922,96 @@ def unique(
...
+                incoming_offset = 0
+            else:
+                origin = (rank - p) % size
+                incoming_offset = gres_offsets[origin]
+                tmp = torch.empty(
+                    gres_map[0].tolist(), dtype=local_data.dtype, device=local_data.device
+                )
+            # loop through unique elements, find matching position in data
+            for i, el in enumerate(lres):
+                counts = torch.zeros_like(local_data, dtype=torch.int8, device=local_data.device)
```
coquelin77, line 2983: is int8 too small here? can we spare the space for a larger dtype? i think this only goes to 128, which should be fine for a while, but int16/32 shouldnt cost that much

coquelin77 (changes requested):

heat/core/manipulations.py
```diff
@@ -1980,116 +1980,113 @@ def shape(a):
...
-            else torch.empty((0,) + local_sorted.size()[1:], dtype=local_sorted.dtype)
+    elif sort_op is torch.unique:
+        local_sorted = sort_op(transposed, dim=0, **kwargs)[0]
+        local_shape = local_sorted.shape
+        if 0 in local_shape:
+            local_shape = transposed.shape
+        lshape_map = torch.empty(
+            (size, transposed.ndim), dtype=torch.int64, device=transposed.device
         )
+        a.comm.Allgather(torch.tensor(local_shape), lshape_map)
```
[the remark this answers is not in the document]
ClaudiaComito: This is a relic from when we didn't have the `dndarray.counts_displs()` method. Replaced

heat/core/manipulations.py
```diff
@@ -1942,101 +1942,111 @@ def shape(a: DNDarray) -> Tuple[int, ...]:
...
-        # sorting is affected by split, processes need to communicate results
-        # transpose so we can work along the 0 axis
-        transposed = a.larray.transpose(axis, 0)
-        local_sorted, local_indices = torch.sort(transposed, dim=0, descending=descending)
-
-        size = a.comm.Get_size()
-        rank = a.comm.Get_rank()
+    size = a.comm.Get_size()
+    rank = a.comm.Get_rank()
+    transposed = a.larray.transpose(axis, 0)
```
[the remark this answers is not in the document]
ClaudiaComito: > this assumes a 2D torch tensor i believe. is this the case at this point?

No it doesn't, `torch.transpose` works like `np.swapaxes`, this will just transpose axis 0 with axis `axis` and leave the rest be.

heat/core/manipulations.py
```diff
@@ -1942,101 +1942,111 @@ def shape(a: DNDarray) -> Tuple[int, ...]:
...
-        length = local_sorted.size()[0]
-
-        # Separate the sorted tensor into size + 1 equal length partitions
-        partitions = [x * length // (size + 1) for x in range(1, size + 1)]
-        local_pivots = (
-            local_sorted[partitions]
-            if counts[rank]
-            else torch.empty((0,) + local_sorted.size()[1:], dtype=local_sorted.dtype)
+    elif sort_op is torch.unique:
+        local_sorted = sort_op(transposed, dim=0, **kwargs)[0]
```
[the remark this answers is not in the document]
ClaudiaComito: Yes,  `axis` must match `split`, and we transpose everything so that `split` is always 0.

heat/core/manipulations.py
```diff
@@ -1942,101 +1942,111 @@ def shape(a: DNDarray) -> Tuple[int, ...]:
...
+        counts, displs = tuple(counts.tolist()), tuple(displs.tolist())
 
-        # Only processes with elements should share their pivots
-        gather_counts = [int(x > 0) * size for x in counts]
-        gather_displs = (0,) + tuple(np.cumsum(gather_counts[:-1]))
+    unique_along_axis = True if sort_op is torch.unique and axis is not None else False
 
-        pivot_dim = list(transposed.size())
-        pivot_dim[0] = size * sum([1 for x in counts if x > 0])
+    length = local_sorted.size()[0]
```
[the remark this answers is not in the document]
ClaudiaComito: This is from the original `sort` implementation, but interesting, `.shape[0]` is actually fastest, `len(tensor)` slowest (`len` 3x slower than `.shape[0]`).

Changed to `.shape[0]`.

heat/core/manipulations.py
```diff
@@ -1942,101 +1942,111 @@ def shape(a: DNDarray) -> Tuple[int, ...]:
...
+        else torch.empty((0,) + local_sorted.size()[1:], dtype=local_sorted.dtype)
+    )
 
-        pivot_dim[0] = size - 1
-        global_pivots = torch.empty(
-            pivot_dim, dtype=a.dtype.torch_type(), device=a.device.torch_device
-        )
+    # Only processes with elements should share their pivots
+    gather_counts = [int(x > 0) * size for x in counts]
+    gather_displs = (0,) + tuple(np.cumsum(gather_counts[:-1]))
```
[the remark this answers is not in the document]
ClaudiaComito: legacy code, fixed

heat/core/manipulations.py
ClaudiaComito: Yes, and it does, but  `unique` returns the inverse indices.
```
x[sort_indices] == sorted_x
unique_x[unique_indices] == x
```
so the inverse indices for `unique`  are dealt with in `unique()`, not within `__pivot_sorting()`. At least at the time I couldn't think of a way of exploiting `__pivot_sorting()` for that, I'll think about it some more.

heat/core/manipulations.py
ClaudiaComito: legacy code, fixed

heat/core/manipulations.py
ClaudiaComito: Interesting. I haven't looked into the `indices` part of `__pivot_sorting` at all. I'll check this out when tackling #777 if that's ok with you.

heat/core/manipulations.py
ClaudiaComito: legacy code, fixed

heat/core/manipulations.py
ClaudiaComito: modified, now using `torch.prod(a.lshape[0])* a.larray.element_size()`

heat/core/manipulations.py
ClaudiaComito: you're right, I started this off as an array containing only 1s and 0s, but that evolved. Great you noticed, now int32

ClaudiaComito:
Needs #938

Issue #656, opened 2020-08-21
Indexing does not preserve the lshape of elements


Inzlinger:
**Description**
When using list slicing the local shape of MPI-Ranks that are left empty cause issues with other functions, as their shape is always (0,).

**To Reproduce**
Run with n>2:
```
import heat as ht

a = ht.array([[1,1],[2,2],[3,3]], split=0)
index = ht.array([0, 1], split=0)
b = a[index]
b.balance_()
print(b.lshape, b.gshape)
ht.spatial.cdist(a,b)
```

**Expected behavior**
Lshape consistent over all ranks, even when they are left empty.

**Version Info**
HeAT on branch master, f2ca0021cf8c9ecefaa0be384e7e83ade53bd663

ClaudiaComito:
@Inzlinger thanks for reporting this and sorry it took so long to get back to you. If you can, check out [this branch](https://github.com/helmholtz-analytics/heat/tree/bug/754-getitem-indexing) - I hope we have a more consistent indexing there. Let me know how it goes.

Issue #754, opened 2021-04-08
Wrong shape after indexing a single element with a slice


ben-bou:
**Description**
As seen in the code below, after a \_\_getitem__ that pulls only one element from one process and leaves the other processes empty, the gshape is wrong.

**To Reproduce**
1. Which module/class/function is affected?
\_\_getitem__
2. What are the circumstances under which the bug appears?
one process has one element, the other process is empty.
3. What is the exact error message / erroneous behavior?
wrong `gshape` and `lshape`

**Illustrative**
```python
import heat as ht

arr = ht.ones((1,10), split=1)

res = arr[:,0:1]

print('rank:', ht.MPI_WORLD.rank, 'gshape:', res.gshape, 'lshape:', res.lshape, 'split:', res.split)
```
Run with one process, the output is as expected:
```
rank: 0 gshape: (1, 1) lshape: (1, 1) split: 1
```

Run with 2 processes:
```
rank: 0 gshape: (1,) lshape: (1, 1) split: 1
rank: 1 gshape: (1,) lshape: (0,) split: 1
/home/b.bourgart/anaconda3/lib/python3.7/site-packages/heat/core/dndarray.py:1598: ResourceWarning: This process (rank: 1) is without data after slicing, running the .balance_() function is recommended
  ResourceWarning,
```
The gshape has only one dimension, but should have two because a `slice` was used for indexing. The lshape of rank 1 should also have two dimensions (#656). 

**Expected behavior**
See above.

**Version Info**
0.5.1; also on master branch

**Additional comments**
Any other comments here.

ben-bou:
Additionally, consider the following code (using the current master branch):
```python
import heat as ht
mask3D = ht.random.randn(2,2,2, split=0)
print('before where |', 'rank:', ht.MPI_WORLD.rank, 'gshape:', mask3D[0].gshape, 'lshape:', mask3D[0].lshape, 'split:', mask3D[0].split, flush=True); ht.MPI_WORLD.Barrier()  # organize output
mask2D = ht.where(mask3D[0] > 0, 1., 0.)
print(' after where |', 'rank:', ht.MPI_WORLD.rank, 'gshape:', mask2D.gshape, 'lshape:', mask2D.lshape, 'split:', mask2D.split, flush=True)
```
With one process, this results in:
```
before where | rank: 0 gshape: (2, 2) lshape: (2, 2) split: 0
 after where | rank: 0 gshape: (2, 2) lshape: (2, 2) split: 0
[...]

PR #755, opened 2021-04-08
make create_lshape_map safe for empty processes

ben-bou:
## Description

By @coquelin77 [comment](https://github.com/helmholtz-analytics/heat/issues/656#issuecomment-678262744) on issue #656  should individual functions handle the case that a process has no data. This PR adds that functionality to the `create_lshape_map` function (which currently throws an exception every time it encounters a process with no data).

Is this actually needed? Because it's so small I figured that just creating the PR is faster than asking beforehand. 

Additional note: Once the lshape is (0,) (#656), it is impossible to know in which dimension(s) the zero is, therefore I assume it to be in the split dimension if the dndarray is split. Another possibility might be to check the gshape for zeros as well.

Issue/s resolved: #656

## Changes proposed:
- check if the local process is empty
- If so, use gshape instead of lshape
- Check if split is not None
- Add zero in the split-dimension

## Type of change
Non-breaking

## Due Diligence

no

skip ci

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -1142,7 +1142,16 @@ class DNDarray:
         lshape_map = torch.zeros(
             (self.comm.size, len(self.gshape)), dtype=torch.int, device=self.device.torch_device
         )
-        lshape_map[self.comm.rank, :] = torch.tensor(self.lshape, device=self.device.torch_device)
+        if self.lnumel:
+            lshape_map[self.comm.rank, :] = torch.tensor(
+                self.lshape, device=self.device.torch_device
+            )
+        else:  # process is empty -> lshape is always (0,)
+            lshape_map[self.comm.rank, :] = torch.tensor(
+                self.gshape, device=self.device.torch_device
+            )
+            if self.split is not None:  # otherwise the global array has a zero in its shape
+                lshape_map[self.comm.rank, self.split] = 0
         self.comm.Allreduce(MPI.IN_PLACE, lshape_map, MPI.SUM)
         return lshape_map
 
--- a/heat/core/tests/test_dndarray.py
+++ b/heat/core/tests/test_dndarray.py
@@ -261,6 +261,13 @@ class TestDNDarray(TestCase):
         data.balance_()
         self.assertTrue(data.is_balanced())
 
+        data = ht.zeros((2, 3, 4, 100), split=3)
+        data = data[:, :, :, :2]  # only rank 0 has data
+        lshape_map = data.create_lshape_map()
+        lshape_map = lshape_map[:, : data.split]  # remove split-axis
+        for lshape in lshape_map:  # all others should be the same
+            self.assertTrue(all(lshape == lshape_map[0]))
+
         data = ht.zeros((70, 20), split=0, dtype=ht.float64)
         data = data[:50]
         data.balance_()


mtar:

ClaudiaComito:
@ben-bou  thanks for this. I ran into a similar roadblock and made some changes to `create_lshape_map` in  #749 , do you want to check it out and see if it solves your problem? That would be branch `features/unique-sort-distributed`. 

Apologies for spamming on a Sunday. Weekdays are crazy with the whole family in home-office/schooling. Needless to say, I'm not expecting replies on the weekend.

Cheers!

ben-bou:
@ClaudiaComito thanks for your reply, no need to apologise!
I looked at the `create_lshape_map` in #749 and I don't think the changes address the issue #656 because in the case of a distributed and unbalanced array it still tries to write the `lshape` into the `lshape_map`. However, if e.g. only one out of 4 processes has data (so the array is distributed and unbalanced), the `lshape` of the 'empty' processes is always `(0,)` and cannot be written into the `lshape_map`. Anyway that's how I understand it.
Cheers!

ClaudiaComito:
> @ClaudiaComito thanks for your reply, no need to apologise!
> I looked at the `create_lshape_map` in #749 and I don't think the changes address the issue #656 because in the case of a distributed and unbalanced array it still tries to write the `lshape` into the `lshape_map`. However, if e.g. only one out of 4 processes has data (so the array is distributed and unbalanced), the `lshape` of the 'empty' processes is always `(0,)` and cannot be written into the `lshape_map`. Anyway that's how I understand it.
> Cheers!

Are you on version 0.5.1? 

We have fixed the local shapes for empty processes to be consistent with the global shape, I think it was  in #687, but that was after the last release I'm afraid. We will put out another release soon.

ClaudiaComito:
> > @ClaudiaComito thanks for your reply, no need to apologise!
> > I looked at the `create_lshape_map` in #749 and I don't think the changes address the issue #656 because in the case of a distributed and unbalanced array it still tries to write the `lshape` into the `lshape_map`. However, if e.g. only one out of 4 processes has data (so the array is distributed and unbalanced), the `lshape` of the 'empty' processes is always `(0,)` and cannot be written into the `lshape_map`. Anyway that's how I understand it.
> > Cheers!
> 
> Are you on version 0.5.1?
> 
> We have fixed the local shapes for empty processes to be consistent with the global shape, I think it was in #687, but that was after the last release I'm afraid. We will put out another release soon.

OK so already on the main branch (i.e. independently from #749), if I run the following:
```
data = torch.zeros(3, 5)
if rank == 1:
   data = torch.zeros(0, 5)
data = ht.array(data, is_split=0)
lshape_map = data.create_lshape_map()
print(lshape_map)
```

I get the following (on 4 processes):
```
[1,0]<stdout>:tensor([[3, 5],
[1,0]<stdout>:        [0, 5],
[1,0]<stdout>:        [3, 5],
[1,0]<stdout>:        [3, 5]], dtype=torch.int32)
[1,1]<stdout>:tensor([[3, 5],
[1,1]<stdout>:        [0, 5],
[1,1]<stdout>:        [3, 5],
[1,1]<stdout>:        [3, 5]], dtype=torch.int32)
[1,2]<stdout>:tensor([[3, 5],
[1,2]<stdout>:        [0, 5],
[1,2]<stdout>:        [3, 5],
[1,2]<stdout>:        [3, 5]], dtype=torch.int32)
[1,3]<stdout>:tensor([[3, 5],
[1,3]<stdout>:        [0, 5],
[1,3]<stdout>:        [3, 5],
[1,3]<stdout>:        [3, 5]], dtype=torch.int32)
```

ben-bou:
Ok thanks. Can you please try the code in #754? I just ran that on the master branch and got the lshape `(0,)`, as stated in the example.
This then shows that this PR is actually not needed and just part of issue #754, right?

ClaudiaComito:
> Ok thanks. Can you please try the code in #754? I just ran that on the master branch and got the lshape `(0,)`, as stated in the example.
> This then shows that this PR is actually not needed and just part of issue #754, right?

I just checked #754 - yes I have the same problem. Thanks for bringing it up. Indeed I think it's a `getitem` indexing glitch and has nothing to do with` create_lshape_map`. I'll look into it.

ben-bou:
Many thanks! I'll close this PR

Issue #703, opened 2020-12-07
Indexing not Stable


Inzlinger:
**Description**
When using indices to acces array elements the order in the result is not stable for different splits.

**To Reproduce**

`````
import heat as ht

a = ht.array([0,1], split = 0)

indices = ht.array([0, 1, 0, 1], split=0)

b = a[indices]
b.balance_()

print(b._DNDarray__array)

print(b)
``````
With 1 process this gives [0, 1, 0, 1] as expected
Run this with n=2 and it changes the order to [0, 0, 1, 1]. 

**Expected behavior**
The same ordering in the result for all splits.

ClaudiaComito:
Thanks @Inzlinger , I'm looking into getitem for other reasons so I'll check this out as well if @coquelin77 agrees.

coquelin77:
i am pretty sure that this is caused by the assumption if the index is a split `DNDarray`, it is assumed to be non-repeating. when the indices are gathered they are then sorted. this is to determine which process has the requisite data. this could also be done with a mask on the index instead. then the local key which is generated would use this mask based on where the mask == the local rank number. or something along these lines.

@ClaudiaComito I have no objections to you looking into this one. i dont think that I will get to this for a little while otherwise. when i have more time available i will look at this again.

ClaudiaComito:
@Inzlinger solving this issues requires communication, i.e. a distributed getitem. This would be needed and awesome and is on our list. If you want to start, you're very welcome! If you want we can organize a chat about how to proceed.

ClaudiaComito:
Working on it  in #914 

---
Reviewed within #1109

ClaudiaComito:
solved by #938

PR #758, opened 2021-04-15
Fix indexing inconsistencies

ClaudiaComito:
## Description

Fixing a few inconsistencies in the slicing of DNDarrays. 

1.  lshapes after slicing, esp. with no data on process.
2. ~~number of dimensions after slicing, esp.  if the slice only contains 1 element, or if the slice is along the split axis. We cannot be 100% consistent with numpy/torch here, as we need to keep the split dimension no matter what.~~  Slicing one single element along the split axis results in the loss of the split dimension, i.e. in a  `DNDarray` with `split=None`. The rank containing the element broadcasts it to the others.

Issue/s resolved: #656 #754 #770

## Changes proposed:
- local shapes are now always consistent with the global shape  of the sliced `DNDarray`, incl. when local tensors are empty
- always keep "sliced" dimension when key is `slice()`
- ~~always keep "sliced" dimension if it correspond to the split axis (N.B. different from numpy/torch)~~ 
- advanced indexing now works with lists, dndarrays, torch tensors and ndarrays
- distributed advanced indexing: valid (i.e. present on rank) local indices of `key[split]`  must be applied to all other dimensions as well  
- tests updated 
## Also:
- implemented `DNDarray.counts_displs()`, returns actual counts and displacements (items and offsets) without assuming that the DNDarray is balanced, leading to...
- `getitem` no longer assumes that the DNDarray is balanced, partially solving #668 

## Type of change

- Bug fix and breaking change: getting 1 item along the split dimension  now results in the loss of the split axis, hence in a non-distributed DNDarray (the 1 item is "Bcast"ed to the other processes).

## Due Diligence

no


ClaudiaComito:
@ben-bou I can't ask you to review but will be happy to hear feedback

Inzlinger:
Looks like the problems i had in #656 are fixed with this, thanks.
Do you have time to look into #703 as well? Else i might take a shot at it, or try to find a workaround as i also need it for #760.

heat/core/dndarray.py
```diff
@@ -1563,54 +1563,64 @@ def __getitem__(self, key):
...
-                    self.device,
-                    self.comm,
-                    self.balanced,
-                )
+            arr = self.__array[key].reshape(gout_full)
+            return DNDarray(
+                arr, tuple(gout_full), self.dtype, new_split, self.device, self.comm, self.balanced
+            )
+
         # else: (DNDarray is distributed)
```
ben-bou, line 1624: Because the `comm.chunk` is used, the DNDarray has to be distributed and balanced, right? 
For distributed and unbalanced, the `lshape_map`  should be used instead of `comm.chunk`

ClaudiaComito: Yes, this problem is the reason why we implemented the `balanced` property, see #668 and #681. I've now implemented a `DNDarray.counts_displs` method to provide the actual counts and offsets for the DNDarray, which in turn uses `create_lshape_map`, which I have modified to only need communication when the DNDarray is not balanced.

heat/core/dndarray.py
```diff
@@ -566,11 +584,23 @@ def create_lshape_map(self) -> torch.Tensor:
         Generate a 'map' of the lshapes of the data on all processes.
         Units are ``(process rank, lshape)``
         """
+        if not self.is_distributed:
+            return torch.tensor(self.gshape).reshape(1, self.ndim)
+
```
coquelin77, line 589: this should be the lshape for every rank. so the shape would be `(self.comm.size, len(self.gshape))` with the gshape as all rows

heat/core/dndarray.py
```diff
@@ -818,28 +871,46 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
...
+            active_rank = torch.where(key[self.split] >= chunk_starts)[0][-1].item()
+            # slice `self` on `active_rank`, allocate `arr` on all other ranks in preparation for Bcast
+            if rank == active_rank:
+                key[self.split] -= chunk_start.item()
+                arr = self.__array[tuple(key)].reshape(tuple(lout))
+            else:
+                arr = torch.empty(tuple(lout), dtype=self.larray.dtype, device=self.larray.device)
+            # broadcast result
+            self.comm.Barrier()
```
coquelin77, line 907: needed?

coquelin77 (changes requested):

heat/core/dndarray.py
ClaudiaComito: you're absolutely right!

heat/core/dndarray.py
```diff
@@ -818,28 +871,46 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
...
         if 0 in arr.shape:
-            # arr is empty
-            # gout is all 0s as is the shape
-            warnings.warn(
-                "This process (rank: {}) is without data after slicing, "
-                "running the .balance_() function is recommended".format(self.comm.rank),
-                ResourceWarning,
-            )
+            # no data on process
+            lout[new_split] = 0
```
ben-bou, line 912: This is not always the expected behaviour.
e.g.
```python
arr = ht.ones((10,1,10), split=2)
res = arr[:,:-1,:]
print('rank:', ht.MPI_WORLD.rank, 'gshape:', res.gshape, 'lshape:', res.lshape)
# rank: 0 gshape: (10, 0, 10) lshape: (10, 0, 0)
# rank: 1 gshape: (10, 0, 10) lshape: (10, 0, 0)
```
Expected would be an lshape of `(10, 0, 5)` for both.

heat/core/dndarray.py
```diff
@@ -778,16 +823,24 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
                 if not isinstance(lkey[self.split], torch.Tensor)
                 else lkey[self.split]
             )
-
             loc_inds = torch.where((inds >= chunk_start) & (inds < chunk_end))
+            # if there are no local indices on a process, then `arr` is empty
```
ben-bou, line 827: > if there are no local indices on a process, then `arr` is empty

`arr` is empty in the `split` dimension -> `lout[self.split]` should be set to 0 here

ClaudiaComito: You're right, done.

ben-bou:
I got this after indexing a 2D dndarray with (0,0), so receiving a scalar (0d).
```  
File "/p/project/cslts/local/juwels/HeAT/experimental_HeAT/lib/python3.8/site-packages/heat/core/dndarray.py", line 887, in __getitem__
    arr = self.comm.bcast(arr, root=active_rank)
  File "mpi4py/MPI/Comm.pyx", line 1257, in mpi4py.MPI.Comm.bcast
  File "mpi4py/MPI/msgpickle.pxi", line 630, in mpi4py.MPI.PyMPI_bcast
  File "mpi4py/MPI/msgpickle.pxi", line 631, in mpi4py.MPI.PyMPI_bcast
mpi4py.MPI.Exception: Message truncated, error stack:
PMPI_Bcast(448)...............: MPI_Bcast(buf=0x7ffc2d95e9b0, count=1, MPI_INT, root=0, comm=MPI_COMM_WORLD) failed
PMPI_Bcast(434)...............: 
MPID_Bcast(76)................: 
MPIR_Bcast_impl(310)..........: 
MPIR_Bcast_intra_auto(223)....: 
MPIR_Bcast_intra_binomial(112): 
(unknown)(): Message truncated
```

ClaudiaComito:
> I got this after indexing a 2D dndarray with (0,0), so receiving a scalar (0d).
> 
> ```
> File "/p/project/cslts/local/juwels/HeAT/experimental_HeAT/lib/python3.8/site-packages/heat/core/dndarray.py", line 887, in __getitem__
>     arr = self.comm.bcast(arr, root=active_rank)
>   File "mpi4py/MPI/Comm.pyx", line 1257, in mpi4py.MPI.Comm.bcast
>   File "mpi4py/MPI/msgpickle.pxi", line 630, in mpi4py.MPI.PyMPI_bcast
>   File "mpi4py/MPI/msgpickle.pxi", line 631, in mpi4py.MPI.PyMPI_bcast
> mpi4py.MPI.Exception: Message truncated, error stack:
> PMPI_Bcast(448)...............: MPI_Bcast(buf=0x7ffc2d95e9b0, count=1, MPI_INT, root=0, comm=MPI_COMM_WORLD) failed
> PMPI_Bcast(434)...............: 
> MPID_Bcast(76)................: 
> MPIR_Bcast_impl(310)..........: 
> MPIR_Bcast_intra_auto(223)....: 
> MPIR_Bcast_intra_binomial(112): 
> (unknown)(): Message truncated
> ```

@ben-bou I can't reproduce this locally. Can you post your code? Thanks a lot!

heat/core/dndarray.py
ClaudiaComito: thanks, not really, was left over from tests.

heat/core/dndarray.py
ClaudiaComito: Thanks, fixed

ben-bou:
> @ben-bou I can't reproduce this locally. Can you post your code? Thanks a lot!

@ClaudiaComito So it's actually not reproducible but rather the same problem you encountered in the tests: in a loop, the first 4 bcasts fail, after that it works. How did you figure out where those 'wrong' broadcasts were happening?

ClaudiaComito:
> > @ben-bou I can't reproduce this locally. Can you post your code? Thanks a lot!
> 
> @ClaudiaComito So it's actually not reproducible but rather the same problem you encountered in the tests: in a loop, the first 4 bcasts fail, after that it works. How did you figure out where those 'wrong' broadcasts were happening?

Could it be that you have a loop calling something like `a[i]` where `i` depends on `rank`?

ben-bou:
> > 
> > @ClaudiaComito So it's actually not reproducible but rather the same problem you encountered in the tests: in a loop, the first 4 bcasts fail, after that it works. How did you figure out where those 'wrong' broadcasts were happening?
> 
> Could it be that you have a loop calling something like `a[i]` where `i` depends on `rank`?

I have no idea what is happening, but something is wrong. You changed the broadcasting in `__getitem__` from `Bcast` to `bcast`. Is there a reason for that?
I changed it back to `Bcast` and applied `torch.contiguous` to get rid of differing strides. That seemed to help (although I don't know why, there was an error regarding the strides in `communication.mpi_type_and_elements_of` so I just got rid of the strides completely).

However, using ParaStationMPI, it gave me
 ```
MPIR_Bcast_intra_binomial(123): message sizes do not match across processes in the collective routine: Received 8 but expected 40
```
at a `__getitem__` even earlier in my script than the original error;

using OpenMPI, it gave me
 ```
EXCEPTION MPI_ERR_TRUNCATE: message truncated
[1621271546.551622] [jwc08n231:27585:0]    ucp_context.c:735  UCX  WARN  transports 'cuda_ipc','gdr_copy','cuda_copy' are not available, please use one or more of: cma, dc, dc_mlx5, dc_x, ib, mm, posix, rc, rc_mlx5, rc_v, rc_verbs, rc_x, self, shm, sm, sysv, tcp, ud, ud_mlx5, ud_v, ud_verbs, ud_x
[1621271547.260514] [jwc08n231:27585:0]      tag_match.c:61   UCX  WARN  unexpected tag-receive descriptor 0x55f40c0 was not matched
[1621271547.260525] [jwc08n231:27585:0]      tag_match.c:61   UCX  WARN  unexpected tag-receive descriptor 0x55f8240 was not matched
[1621271547.260530] [jwc08n231:27585:0]      tag_match.c:61   UCX  WARN  unexpected tag-receive descriptor 0x55f6180 was not matched
[1621271547.260533] [jwc08n231:27585:0]      tag_match.c:61   UCX  WARN  unexpected tag-receive descriptor 0x55f2000 was not matched
```
when doing `ht.float64(1.0) + distributed_dndarray`, although later than the original error;

and using IntelMPI, it went way further than all the others but eventually just hanged.

Any ideas?

.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
+ignore = C901, D402, E203, E266, E402, E501, W503, F403, F401
+max-line-length = 100
+max-complexity = 18
+select = B,C,E,F,W,T4,B9
```
coquelin77, line 5: why is this in a new file? why not in the `setup.cfg` file?

CHANGELOG.md
```diff
@@ -58,6 +58,22 @@
...
+
+## Bug fixes
+- [#709](https://github.com/helmholtz-analytics/heat/pull/709) Set the encoding for README.md in setup.py explicitly.
+- [#716](https://github.com/helmholtz-analytics/heat/pull/716) Bugfix: Finding clusters by spectral gap fails when multiple diffs identical
+- [#732](https://github.com/helmholtz-analytics/heat/pull/732) Corrected logic in `DNDarray.__getitem__` to produce the correct split axis
+- [#734](https://github.com/helmholtz-analytics/heat/pull/734) Fix division by zero error in `__local_op` with out != None on empty local arrays.
+- [#735](https://github.com/helmholtz-analytics/heat/pull/735) Set return type to bool in relational functions.
+- [#744](https://github.com/helmholtz-analytics/heat/pull/744) Fix split semantics for reduction operations
+- [#756](https://github.com/helmholtz-analytics/heat/pull/756) Keep track of sent items while balancing within `sort()`
+- [#758](https://github.com/helmholtz-analytics/heat/pull/758) Fix indexing inconsistencies in `DNDarray.__getitem__`
```
coquelin77, line 70: this should be moved into the pending additions section correct? Judging by line number, its down in the `1.0.0` section. also, why are there so many additions here? shouldnt all of these already be in the file?

heat/core/dndarray.py
```diff
@@ -655,24 +683,23 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
         """
         key = getattr(key, "copy()", key)
         l_dtype = self.dtype.torch_type()
-        kgshape_flag = False
+        advanced_ind = False
+        # kgshape_flag = False
```
coquelin77, line 687: dead code?

heat/core/tests/test_dndarray.py
```diff
@@ -1112,13 +1136,13 @@ def test_setitem_getitem(self):
         a[1:4, 1] = 1
         self.assertTrue((a[1:4, 1] == 1).all())
         self.assertEqual(a[1:4, 1].gshape, (3,))
-        self.assertEqual(a[1:4, 1].split, 0)
+        self.assertEqual(a[1:4, 1].split, None)
```
coquelin77, line 1139: so this is a more major change. can you add something to the changelog which mentions that slicing with an int in the split dimension will result in a split=None?

perhaps we should have a file which contains general usage notes for things like this?

coquelin77 (changes requested):
looks good to me. only minor structural changes needed as far as i can tell. the actual code proposed looks good to me

CHANGELOG.md
ClaudiaComito: Correct, because I'm syncing with the `release/1.0.x` branch, I hadn't imported the new changelog headers. Fixed it now.

heat/core/tests/test_dndarray.py
ClaudiaComito: Added under `Breaking Changes`, and yes, I'll write something up for this one.

ClaudiaComito:
> > Could it be that you have a loop calling something like `a[i]` where `i` depends on `rank`?
> 
> @ClaudiaComito You changed the broadcasting in `__getitem__` from `Bcast` to `bcast`. Is there a reason for that?
> 
> I tried changing it back to `Bcast`, but got this error:
> 
> ```
>   File "/p/project/cslts/local/juwels/HeAT/experimental_HeAT/lib/python3.8/site-packages/heat/core/dndarray.py", line 892, in __getitem__
>     self.comm.Bcast(arr, active_rank)
>   File "/p/project/cslts/local/juwels/HeAT/experimental_HeAT/lib/python3.8/site-packages/heat/core/communication.py", line 704, in Bcast
>     ret, sbuf, rbuf, buf = self.__broadcast_like(self.handle.Bcast, buf, root)
>   File "/p/project/cslts/local/juwels/HeAT/experimental_HeAT/lib/python3.8/site-packages/heat/core/communication.py", line 689, in __broadcast_like
>     return func(self.as_buffer(srbuf), root), srbuf, srbuf, buf
>   File "/p/project/cslts/local/juwels/HeAT/experimental_HeAT/lib/python3.8/site-packages/heat/core/communication.py", line 326, in as_buffer
>     mpi_type, elements = cls.mpi_type_and_elements_of(obj, counts, displs)
>   File "/p/project/cslts/local/juwels/HeAT/experimental_HeAT/lib/python3.8/site-packages/heat/core/communication.py", line 281, in mpi_type_and_elements_of
>     strides[0] = obj.stride()[-1]
> ```
> 
> Which I believe shouldn't happen. The `mpi_type_and_elements_of` should be able to handle all kinds of Tensors.
> 
> As a side note, `bcast`, as is currently used in this branch, pickles the entire object. I'm unsure if this is efficient, especially given the strides. Wouldn't this bcast the entire memory footprint of the Tensor to all processes and then generate a view to only a single slice only afterwards?
> 
> EDIT2: the error had nothing to do with `bcast` or `__getitem__`; the error messages just made it seem that way. Thanks anyway. The questions about the strides in `mpi_type_and_elements_of` and the efficiency of `bcast` still stand but are probably out of the context of this PR

@ben-bou  indeed I switched to `bcast` because `Bcast` was giving me problems, thanks for reminding me I needed to look into this. We seem to have a bug when broadcasting slices of an object (#784 ), so for the time being I'm going to keep `bcast` and will replace it with `Bcast` later.

coquelin77:
rerun tests

coquelin77:
rerun tests

coquelin77:
rerun tests

coquelin77:
rerun tests

coquelin77:
rerun tests

coquelin77 (approved):

PR #761, opened 2021-04-20
Feature/713 result type

mtar:
## Description

This PR provides an implementation of result_type. Additionally, some changes in __binary_op on the output are introduced.

Issue/s resolved: #713 

## Changes proposed:
- add result_type()
- __binary_op calls result_type() for type promotions
- __binary_op uses scalars if an operand is a number
- new entries type mappings

## Type of change

## Due Diligence

yes, the result type of __binary_op follows result_type and scalars are returned when adding scalars and numbers.


heat/core/types.py
```diff
@@ -854,6 +861,87 @@ def promote_types(
...
+            else:
+                prec1 = 2
+        except Exception:
+            try:
+                # type
+                if isinstance(arg, np.dtype):
+                    arg = arg.char
+                type1 = canonical_heat_type(arg)
+                prec1 = 1
+            except Exception:
```
coquelin77, line 907: why such a broad exception?

heat/core/types.py
```diff
@@ -854,6 +861,87 @@ def promote_types(
...
+
+                tc1 = __type_codes[type1]
+                tc2 = __type_codes[type2]
+
+                if tc1 < tc2:
+                    return type2, min(prec1, prec2)
+                if tc1 > tc2:
+                    return type1, min(prec1, prec2)
+
+                raise RuntimeError("Something went wrong")  # pragma: no cover
```
coquelin77, line 937: can drop these down some tabs, the if loops end with returns

coquelin77 (changes requested):
looks good. minor changes and questions

heat/core/types.py
mtar: Yes, I flattened the else block.

heat/core/types.py
coquelin77: what is raised when this happens? `TypeError`?

mtar: Ah - I first thought you meant the except block and not the caught exception :facepalm:

coquelin77 (approved):

coquelin77:
will merge once tests finish

Issue #762, opened 2021-04-21
dndarray repr method output


mtar:
**Description**
A clear and concise description of the bug and the associated functionality.

The repr method in dndarray uses the one from the underlying pytorch tensor.

**To Reproduce**
1. Which module/class/function is affected?
dndarray.\_\_repr\_\_()

**Expected behavior**

**Illustrative**
If applicable, add screenshots or minimal examples to help explain your problem.
```
>>> ht.array([1])
tensor([1])
```

**Version Info**
 master c62cd072479efac8d850b85fc9e426e75fafd344

PR #764, opened 2021-04-27
restore repr

mtar:
## Description

This PR fixes the wrong output of `DNDarray.__repr__()` introduced in e88a489e4980e35632b7f2aed503025e3887aad5
Issue/s resolved: #762 

## Changes proposed:
- fix __repr__

## Type of change

## Due Diligence

no


Markus-Goetz:
repr() and str() tend to do different things. I am not sure that a forward is sufficient

mtar:
In my experience, they are doing the same thing in pytorch. In numpy, they are different., [`repr`](https://numpy.org/doc/stable/reference/generated/numpy.array_repr.html) has information on the kind of array and its data type, [`str`](https://numpy.org/doc/stable/reference/generated/numpy.array_str.html) not.

Markus-Goetz:
Fair enough. In this case, we can go with the simple forward.

Markus-Goetz (approved):

Issue #786, opened 2021-06-07
import error not using hdf5


mtar:
**Description**
https://github.com/helmholtz-analytics/heat/blob/3dc9acaea8b21b382671779b3cd233185fc00374/heat/core/io.py#L26
An import error was introduced to Heat in `io.py` 65129c464f7a63acfab6c3a6e03ce1c09ce51f6e
Remove "save_hdf5" from the list at this line.

**Version Info**
 Which version are you using?
master, 1.x

Issue #697, opened 2020-11-11
Output a more descriptive error when file format dependencies are not installed


sebimarkgraf:
**Related**
When loading the data in a format that is not installed, the error message just tells me that the format is not supported.

**Feature functionality**
Give a meaningful error message that the dependencies are not installed for e.g. HDF5

**Additional context**
This helps a lot in debugging why the file format could not be loaded especially when using different environments.

PR #787, opened 2021-06-08
IO import issues

mtar:
## Description

Fixes an import error in io.py when optional dependencies are not available and changes error message when saving or loading a file that requires an optional dependency.

Issue resolved: #786 
Issue resolved: #697 

## Changes proposed:
- remove save_hdf5 from general _all_
- skip tests using optional dependencies if not available
- more descriptive error message for files that need an optional dependency.

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/io.py
+++ b/heat/core/io.py
@@ -691,10 +691,16 @@ def load(
 
     if extension in __CSV_EXTENSION:
         return load_csv(path, *args, **kwargs)
-    elif supports_hdf5() and extension in __HDF5_EXTENSIONS:
-        return load_hdf5(path, *args, **kwargs)
-    elif supports_netcdf() and extension in __NETCDF_EXTENSIONS:
-        return load_netcdf(path, *args, **kwargs)
+    elif extension in __HDF5_EXTENSIONS:
+        if supports_hdf5():
+            return load_hdf5(path, *args, **kwargs)
+        else:
+            raise ImportError("hdf5 is required for file extension {}".format(extension))
+    elif extension in __NETCDF_EXTENSIONS:
+        if supports_netcdf():
+            return load_netcdf(path, *args, **kwargs)
+        else:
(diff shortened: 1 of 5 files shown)


mtar:
run tests

coquelin77:
also i am fine with the PR not hitting the coverage target. this a fix that simply wont unless we test it with multiple venvs on the system, which seems like overkill

mtar:
rerun tests

mtar:
rerun tests

mtar:
run tests

mtar:
rerun tests

heat/core/io.py
```diff
@@ -691,10 +691,16 @@ def load(
...
         return load_csv(path, *args, **kwargs)
-    elif supports_hdf5() and extension in __HDF5_EXTENSIONS:
-        return load_hdf5(path, *args, **kwargs)
-    elif supports_netcdf() and extension in __NETCDF_EXTENSIONS:
-        return load_netcdf(path, *args, **kwargs)
+    elif extension in __HDF5_EXTENSIONS:
+        if supports_hdf5():
+            return load_hdf5(path, *args, **kwargs)
+        else:
+            raise ImportError("hdf5 is required for file extension {}".format(extension))
```
coquelin77, line 698: i think that these should be either `RunTimeErrors` or `TypeErrors`. the issue that this is talking about is that the package is not installed.

coquelin77 (changes requested):
Looks good to me, except that i dont think that it should be an `ImportError`

heat/core/io.py
mtar: I was using `ImportError` because it's an external package that needs to be imported.
I don't know if `TypeError` is good here as the function gets the right argument types. `ValueError` maybe. 
A general `RuntimeError` might be the best here.

coquelin77 (approved):

Issue #784, opened 2021-06-02
`Bcast` sending wrong buffer 


ClaudiaComito:
**Description**
`ht.communication.Bcast` seems to send the wrong portion of a buffer when broadcasting a slice of an array that does not start at index 0. 

**To Reproduce**
```python
a = ht.arange(13*5, split=0).reshape((13, 5))
active_rank = 1
if a.comm.rank == active_rank:
    arr = a.larray[1:2]
else:
    arr = torch.empty((1,5), dtype=a.larray.dtype)
print("before Bcast on rank ", rank, ", arr = ", arr)
a.comm.Bcast(arr, root=active_rank)
print("after Bcast on rank ", rank, ", arr = ", arr)
```

`mpirun -n 2 `

```
before Bcast on rank  1 , arr =  tensor([[40, 41, 42, 43, 44]], dtype=torch.int32)
before Bcast on rank  0 , arr =  tensor([[        1, 423460512,         1,         1,         0]], dtype=torch.int32)
after Bcast on rank  0 , arr =  tensor([[704643072, 721420288, 738197504, 754974720, 771751936]], dtype=torch.int32)
after Bcast on rank  1 , arr =  tensor([[40, 41, 42, 43, 44]], dtype=torch.int32)
```

**Expected behavior**

```python
a = ht.arange(13*5, split=0).reshape((13, 5))
active_rank = 1
if a.comm.rank == active_rank:
    arr = a.larray[0:1]
else:
    arr = torch.empty((1,5), dtype=a.larray.dtype)
print("before Bcast on rank ", rank, ", arr = ", arr)
a.comm.Bcast(arr, root=active_rank)
print("after Bcast on rank ", rank, ", arr = ", arr)
```
`mpirun -n 2 `
```
before Bcast on rank  0 , arr =  tensor([[         0, 1073741824,          0, 1073741824,  465895430]], dtype=torch.int32)
before Bcast on rank  1 , arr =  tensor([[40, 41, 42, 43, 44]], dtype=torch.int32)
after Bcast on rank  0 , arr =  tensor([[40, 41, 42, 43, 44]], dtype=torch.int32)
after Bcast on rank  1 , arr =  tensor([[40, 41, 42, 43, 44]], dtype=torch.int32)
```

**Version Info**
Main branch

ClaudiaComito:
Might be similar or same issue as in #769

ClaudiaComito:
This has been addressed. Closing.

PR #790, opened 2021-06-08
added device call after the bcast in the getitem

coquelin77:
## Description

a minor bug was showing up where the `bcast` operation was not putting the data onto the proper GPU in the `__getitem__` function. the code added may not be hit all the time, but it is safeguard for when this may happen. 

Issue/s resolved: n/a

## Changes proposed:
- move the local array to the proper GPU after the `bcast` when there are integers in the getitem key

## Type of change

## Due Diligence

no


coquelin77:
ignore the diff hit please, im not sure why it only hits it on the master branch.

ClaudiaComito:
Hey @coquelin77 I'm a bit lost, is #784 fixed? In which case, we should be able to switch back to `Bcast`, which will write in the previously allocated buffer on the correct device?

coquelin77:
> Hey @coquelin77 I'm a bit lost, is #784 fixed? In which case, we should be able to switch back to `Bcast`, which will write in the previously allocated buffer on the correct device?

i thought that it was fixed because it solved this problem, but it still caused the same issue that @ben-bou pointed out before. so i left it as`bcast`

ClaudiaComito (approved):
Thanks @coquelin77 , I'm OK with this, maybe just add a note that the fix won't (shouldn't) be necessary after solving #784?

ClaudiaComito (approved):

Issue #783, opened 2021-05-26
Aliases to existing functions: `greater`, `greater_equal`, `less`, `less_equal`, `not_equal`


ClaudiaComito:
**Related**
#774

**Feature functionality**
see [`heat.core.relational`](https://github.com/helmholtz-analytics/heat/blob/master/heat/core/relational.py)

**Additional context**
See [Python array API standard](https://data-apis.org/array-api/latest/index.html)

Issue #774, opened 2021-05-19
Array API compliance


fschlimb:
A consortium is defining an Array API: https://data-apis.org/array-api/latest/index.html

It seems desirable for HeAT to be as compliant as possible.

The following features are missing in HeAT (others are probably incomplete):
 ```
['acosh',
 'argsort',
 'asinh',
 'atanh',
 'bitwise_invert',
 'bitwise_left_shift',
 'bitwise_right_shift',
 'broadcast_arrays',
 'broadcast_to',
 'from_dlpack',
 'greater',
 'greater_equal',
 'less',
 'less_equal',
 'logaddexp',
 'meshgrid',
 'not_equal',
 'result_type',
 'sign',
 'square',
 'uint16',
 'uint32',
 'uint64']
```
The following features are missing in the array itself (others are probably incomplete):
```
['__abs__',
 '__array_namespace__',
 '__dlpack__',
 '__dlpack_device__',
 '__iadd__',
 '__iand__',
 '__iflowdiv__',
 '__ilshift__',
 '__imatmul__',
 '__imod__',
 '__imul__',
 '__ior__',
 '__ipow__',
 '__irshift__',
 '__isub__',
 '__itruediv__',
 '__ixor__',
 '__rand__',
 '__rflowdiv__',
 '__rlshift__',
 '__rmatmul__',
 '__ror__',
 '__rrshift__',
 '__rxor__']
```

ClaudiaComito:
Thanks a lot for this list @fschlimb!

ClaudiaComito:
`bitwise_invert`, `bitwise_left_shift` and  `bitwise_right_shift` are implemented as `ht.invert`, `ht.left_shift` and `ht.right_shift` following the NumPy API reference.

Dhruv454000:
hello @ClaudiaComito , I would like to work on this issue as a part of gsoc. It would be great If you could guide me how can I start. I am aware of numpy but not much aware of parallel computing.

ClaudiaComito:
This is still work-in-progress and being tracked in #1022. 

--- 
Reviewed within #1109

ClaudiaComito:
Check out https://github.com/data-apis/array-api/issues/462 for tracking and signalling array API complicance.

ClaudiaComito:
- [ ] report compliance, how?
- [ ] merge existing PR when CI runs without problems
- [ ] mark namespace experimental
- [ ] create Tracker Issue for missing ops

[...]

PR #792, opened 2021-06-09
Features/783 api relational

lenablind:
## Aliases for comparison functions in  [`heat.core.relational`](https://github.com/helmholtz-analytics/heat/blob/master/heat/core/relational.py)

This PR is motivated by the request in #774.
It is simply an extension of the current API in the corresponding module. 
The following aliases have been implemented:
| Original        | Alias           | 
| ------------- |:-------------:| 
|  gt     | greater |
| ge     | greater_equal      |   
| lt | less      | 
| le     | less_equal     |   
| ne | not_equal      | 

Issue/s resolved: #783 

## Changes proposed:
- Aliases for the above listed functions
- Corresponding tests for the functions

## Type of change

## Due Diligence

no

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,9 @@ Example on 2 processes:
 - [#728](https://github.com/helmholtz-analytics/heat/pull/728) New feature: `nn.DataParallelMultiGPU` which uses `torch.distributed` for local communication (for use with `optim.DASO`)
 - [#728](https://github.com/helmholtz-analytics/heat/pull/728) New feature: `optim.DetectMetricPlateau` detects when a given metric plateaus.
 
+### Relational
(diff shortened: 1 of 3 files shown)


coquelin77:
rerun tests

coquelin77 (changes requested):
why not simply point to the existing functions? there is a bit of extra work going on here

i.e. 
```python
greater_equal = ge
greater_equal.__doc__ = ge.__doc__
```

lenablind:
> why not simply point to the existing functions? there is a bit of extra work going on here
> 
> i.e.
> 
> ```python
> greater_equal = ge
> greater_equal.__doc__ = ge.__doc__
> ```

You are right, that makes absolutely more sense indeed. Thanks for the feedback! I'll change it right away.

coquelin77 (approved):

Issue #789, opened 2021-06-08
performance issues on a single MPI process


coquelin77:
**Description**
Simple operations are showing a massive slowdown vs numpy/torch when run on a single mpi process. 

**To Reproduce**
1. Which module/class/function is affected?
    - so far, it seems to only be the binary ops, however it may effect other things as well. 
2. What are the circumstances under which the bug appears?
    - single mpi process runs on CPU show the behavior most clearly 
3. What is the exact error message / erroneous behavior?
    - the times of all of the processes should be very similar

**Expected behavior**
the times of all three packages should be similar. 

**Illustrative**

```python
  import heat as ht
  import numpy as np
  import time
  arrsize = 100000000
  times = []
  for _ in range(10):
      a = ht.ones(arrsize)#, split=0)
      b = ht.ones(arrsize)#, split=0)
      t1 = time.perf_counter()
      c = a * b + a ** 2 + b ** 2 + a - b
      t2 = time.perf_counter()
      times.append(t2 - t1)
  print(np.mean(times))
```
1 process heat: `2.7430931525974303`
1 process torch: `1.3330379242979689`

with a bit more digging i can see that this might be happening within the `a ** 2` operation.

**Version Info**
Current master (184f112)

**Additional comments**
Thank you to @fschlimb for pointing this out

ClaudiaComito:
Torch binary operations seem to be much more efficient on scalars than on torch tensors. However, Heat always casts scalars to DNDarrays, hence internally Heat's binary ops are always between torch tensors.

Using the script above, but `c = a**2`.   On my laptop, with `mpirun -n 1 `, I get the following:
```
heat times:  0.8049912449333334
numpy times:  0.70067716212
torch times:  0.5644471529142856
```

However, if I replace `a**2` with `a**torch.tensor(2)` (or `a**np.array(2)` for numpy) I get the following:

```
heat times:  0.814409041133333
numpy times:  0.7144210714799999
torch times:  0.7410358746
```
Interesting that for numpy it doesn't make a difference.

I'll think about how to simplify the `__binary_op()` checks and handling of scalars, or @coquelin77 do you want to take over?

ClaudiaComito:
Some single-process test runs:

| Operation  |  Heat  runtime (s) |   NumPy  runtime (s)|  PyTorch runtime (s)|   Heat/PyTorch |
| :--  | :-------------: | :----------: | :-----------: | :-----------: |
[...]

PR #793, opened 2021-06-10
Bug/789 pow binary op performance

coquelin77:
## Description

Optimizations for `pow`

Issue/s resolved: #789 although resolved is a strong word...more work required

## Changes proposed:
- avoid calling `_binary_op` unless absolutely necessary and use a simpler switch instead

## Type of change
Optimization

## Due Diligence

    - this may have an effect on the dtype results. however the tests ran clean with multiple different operations in var which requires this

YES! `comm.chunk` now returns 4 parameters instead of 3!


coquelin77:
rerun tests

heat/core/arithmetics.py
```diff
@@ -776,6 +776,26 @@ def pow(t1: Union[DNDarray, float], t2: Union[DNDarray, float]) -> DNDarray:
     DNDarray([[ 1.,  8.],
             [27., 64.]], dtype=ht.float32, device=cpu:0, split=None)
     """
+    if isinstance(t1, DNDarray):
+        ret = factories.zeros_like(t1)
+        try:
+            t2 = manipulations.resplit(t2, t1.split)
```
ClaudiaComito, line 782: Hey I'm thinking this might backfire. Example: two DNDarrays `t1` and `t2`. 
`t1.shape = (100, 8)` and `t1.split=1`. 
`t2.shape = (8,)`.

The shapes are broadcastable and `ht.pow(t1, t2)` is legitimate. `t2` cannot be split along 1 though, so in the follow-up the assumption is that `t2` is not a DNDarray. Line 786 will fail. 

I'm wondering if we need so many checks at all. All we need to know is if `t1` or `t2` is a scalar (with the other one a DNDarray). How about 
```python
# t1 is a verified DNDarray
if not hasattr(t2, "dtype"):
   ret = torch.pow(t1.larray, t2)
else:
   # call binary_op
```



 or something like that? What am I missing?

ClaudiaComito (changes requested):

heat/core/arithmetics.py
coquelin77: you are correct. that was an issue with the assumption that i made. i updated the logic in this to correct for this now. it should be able to handle anything you can throw at it (i hope).

coquelin77:
rerun tests

coquelin77:
rerun tests

coquelin77:
New numbers:

local tests: 4 procs - `a` = ht.random.random((10000, 10000), dtype=ht.float32, split=split)
numbers are avg of 10 runs 
- pow (new / old)
	- `a ** a`: 	0.2157	/	0.2207
	- `2.5 ** a`: 	0.2099	/	0.2051
	- `a ** 2.5`:	0.1986	/	0.2002
	- `a ** 2`:	0.0709	/	0.2390
- add (new / old)
	- `a + a`: 	0.0531	/	0.0697
	- `2.5 + a`: 	0.0571	/	0.0603
	- `a + 2.5`:	0.0597	/	0.0615
	- `a + 2`:	0.0578	/	0.0592
- sub (new / old)
	- `a - a`: 	0.0590	/	0.0689*
	- `2.5 - a`: 	0.0584	/	0.0570
	- `a - 2.5`:	0.0571	/	0.0583
	- `a - 2`:	0.0587	/	0.0593
- div (new / old)
	- `a / a`: 	0.0582	/	0.0580
	- `2.5  a`: 	0.0582	/	0.0574
	- `a / 2.5`:	0.0615	/	0.0596
	- `a / 2`:	0.0540	/	0.0601
- floordiv (new / old)
	- `a // a`: 	0.1291	/	0.1374
	- `2.5 // a`: 	0.1342	/	0.1371
	- `a // 2.5`:	0.1339	/	0.1406
	- `a // 2`:	0.1886	/	0.1922
- mul (new / old)
	- `a * a`: 	0.0557	/	0.0579
	- `2.5 * a`: 	0.0580	/	0.0572
	- `a * 2.5`:	0.0600	/	0.0595
	- `a * 2`:	0.0600	/	0.0581

coquelin77:
@mtar can you have a look at why this is failing? i dont know why `gpu` isnt recognized. maybe its a but somewhere that im missing

ClaudiaComito:
Some of this PR has been superseded by the latest implementation of `_operations.__binary_op` with distribution sanitation #902 , but a lot of it is still relevant. I will make the necessary changes this week, @coquelin77 please scream if you prefer a review and to introduce the changes yourself.

ClaudiaComito:
Closing this as too stale to update, changes implemented in #1141

PR #794, opened 2021-06-14
meshgrid implementation

mtar:
## Description

Implements meshgrid functionality to Heat. It uses the pytorch function of the same name. That function does not support the indexing parameter. As a hotfix for now, the first and second vector are switched during computation when using 'xy'. If the split is on the first or second axis, they are resplitted beforehand.

Issue/s resolved: #775 

## Changes proposed:
- add meshgrid function

## Type of change

- New feature

## Due Diligence

no

### DIFF
--- a/heat/core/factories.py
+++ b/heat/core/factories.py
@@ -1053,13 +1053,9 @@ def meshgrid(*arrays: Sequence[DNDarray], indexing: str = "xy") -> List[DNDarray
     ----------
     arrays : Sequence[ DNDarray ]
         one-dimensional arrays representing grid coordinates. If exactly one vector is distributed, the returned matrices will
-        be distributed along the axis equal to the index of this vector in the input list. See Note for special case.
+        be distributed along the axis equal to the index of this vector in the input list.
     indexing : str, optional
         Cartesian ‘xy’ or matrix ‘ij’ indexing of output. It is ignored if zero or one one-dimensional arrays are provided. Default: 'xy' .
-
-    Note
-    ----
-    If `indexing == xy` and the vector at position 0 or 1 is distributed, the split axis for the returned arrays will be switched.
 
     Raises
     ------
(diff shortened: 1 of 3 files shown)


mtar:
run tests

mtar:
run tests

mtar:
run tests

heat/core/factories.py
```diff
@@ -1043,6 +1045,76 @@ def logspace(
...
+            arrays[1] = arrays[1].resplit(0)
+
+    grids = torch.meshgrid(*(array.larray for array in arrays))
+
+    # pytorch does not support indexing keyword: switch back
+    if indexing == "xy" and len(arrays) > 1:
+        grids = list(grids)
+        grids[0], grids[1] = grids[1], grids[0]
+
+    return list(asarray(grid, is_split=splitted) for grid in grids)
```
ClaudiaComito, line 1115: Can we use `DNDarray()` here instead of `asarray()`?

heat/core/factories.py
```diff
@@ -1043,6 +1045,89 @@ def logspace(
...
 
 
+def meshgrid(*arrays: Sequence[DNDarray], indexing: str = "xy") -> List[DNDarray]:
+    """
+    Returns coordinate matrices from coordinate vectors.
+
+    Parameters
+    ----------
+    arrays : Sequence[ DNDarray ]
+        one-dimensional arrays representing grid coordinates. If exactly one vector is distributed, the returned matrices will
```
coquelin77, line 1055: "If exactly one vector is distributed..." should be "If one or more vectors are distributed..."

heat/core/tests/test_factories.py
```diff
@@ -2,6 +2,7 @@
 import torch
 
 import heat as ht
+from torch._C import Value
```
coquelin77, line 5: Why is this here? is it used?

coquelin77 (changes requested):
looks good. only minor changes

heat/core/tests/test_factories.py
mtar: No. It was automatically added by the editor, but it is not used.

heat/core/factories.py
mtar: one distributed vector is allowed, no more.

mtar: Otherwise, a `split` parameter will be necessary and the vectors need to be redistributed.

coquelin77 (approved):

Issue #795, opened 2021-06-14
`reshape` issues: no split to `new_split`, method vs. function call, input `*shape` not supported


ClaudiaComito:
**Description**

1. reshaping a non-distributed DNDarray always returns a non-distributed DNDarray no matter what the `new_split` argument is set to:
```python
a = ht.arange(10)
ht.reshape(a,(2,5), new_split=0)
```
```
DNDarray([[0, 1, 2, 3, 4],
          [5, 6, 7, 8, 9]], dtype=ht.int32, device=cpu:0, split=None)
```

2.  kwarg for `reshape()`  is `new_split` for the function call, `axis` for the method call
```python
a = ht.arange(10, split=0)
a.reshape((2,5), new_split=1)
```
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: <lambda>() got an unexpected keyword argument 'new_split'
```
3. the method call does not allow the elements of the shape to be passed in as separate arguments:

```python
import numpy as np
import torch
import heat as ht
numpy = np.arange(10)
torch = torch.arange(10)
heat = ht.arange(10)
numpy.reshape(2,5)
```
```
array([[0, 1, 2, 3, 4],
       [5, 6, 7, 8, 9]])
```
```python
torch.reshape(2,5)
```
```
tensor([[0, 1, 2, 3, 4],
        [5, 6, 7, 8, 9]])
```
```python
heat.reshape(2,5)
```
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/c.comito/HAF/heat/heat/core/manipulations.py", line 1830, in <lambda>
    DNDarray.reshape = lambda self, shape, axis=None: reshape(self, shape, axis)
  File "/Users/c.comito/HAF/heat/heat/core/manipulations.py", line 1768, in reshape
    raise TypeError("shape must be list, tuple, currently {}".format(type(shape)))
TypeError: shape must be list, tuple, currently <class 'int'>
```

**To Reproduce**
1. Which module/class/function is affected?
`manipulations.reshape()` and `dndarray.DNDarray.reshape()`

2. What are the circumstances under which the bug appears?
see above

3. What is the exact error message / erroneous behavior?
see above

**Expected behavior**
1.
```python
a = ht.arange(10)
ht.reshape(a,(2,5), new_split=0)
```
```
DNDarray([[0, 1, 2, 3, 4],
[...]

PR #796, opened 2021-06-14
Bug/795-reshape split none to new split

ClaudiaComito:
## Description

Issue/s resolved: #795 

## Changes proposed:
- `reshape(a, shape, new_split)` now returns a distributed `DNDarray` if `new_split is not None` even if `a` is not distributed
- update keyword argument for `reshape()` method to `new_split` from old `axis`
- allow shape elements to be passed in as separate arguments when calling the `reshape()` method
- replace unnecessary `factories.array()` calls with `DNDarray()` for a significant improvement in performance
- update tests

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -1685,7 +1685,7 @@ def repeat(a: Iterable, repeats: Iterable, axis: Optional[int] = None) -> DNDarr
     return repeated_array
 
 
-def reshape(a: DNDarray, shape: Tuple[int, ...], new_split: Optional[int] = None) -> DNDarray:
+def reshape(a: DNDarray, *shape: Tuple[int, ...], **kwargs) -> DNDarray:
     """
     Returns an array with the same data and number of elements as `a`, but with the specified shape.
 
@@ -1723,6 +1723,10 @@ def reshape(a: DNDarray, shape: Tuple[int, ...], new_split: Optional[int] = None
     """
     if not isinstance(a, DNDarray):
         raise TypeError("'a' must be a DNDarray, currently {}".format(type(a)))
+
+    # check for nested sequence
+    if isinstance(shape[0], (tuple, list)):
+        shape = shape[0]
 
     tdtype, tdevice = a.dtype.torch_type(), a.device.torch_device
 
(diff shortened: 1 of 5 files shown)


mtar:
rerun tests

heat/core/manipulations.py
```diff
@@ -1724,6 +1724,10 @@ def reshape(a: DNDarray, shape: Tuple[int, ...], new_split: Optional[int] = None
     if not isinstance(a, DNDarray):
         raise TypeError("'a' must be a DNDarray, currently {}".format(type(a)))
 
+    # check for nested sequence
+    if isinstance(shape[0], (tuple, list)):
+        shape = shape[0]
+
```
ClaudiaComito, line 1730: This is so slow, please @Markus-Goetz @coquelin77 can you suggest a more efficient way?

heat/core/manipulations.py
```diff
@@ -1824,10 +1848,18 @@ def reshape_argsort_counts_displs(
...
+        dtype=a.dtype,
+        split=new_split,
+        device=a.device,
+        comm=a.comm,
+        balanced=True,
+    )
 
 
-DNDarray.reshape = lambda self, shape, axis=None: reshape(self, shape, axis)
+DNDarray.reshape = lambda self, *shape, **kwargs: reshape(self, *shape, **kwargs)
```
ClaudiaComito, line 1862: This works but passing shape elements as single arguments is so much slower than passing a tuple. Need help here @Markus-Goetz @coquelin77

heat/core/manipulations.py
```diff
@@ -1724,6 +1724,10 @@ def reshape(a: DNDarray, shape: Tuple[int, ...], new_split: Optional[int] = None
     if not isinstance(a, DNDarray):
         raise TypeError("'a' must be a DNDarray, currently {}".format(type(a)))
 
+    # check for nested sequence
+    if isinstance(shape[0], (tuple, list)):
```
ben-bou, line 1728: Is `tuple` and `list` enough here? Why not accept `array_like`? It that case, this could be replaced by 
`if not np.isscalar(shape[0]):`

ben-bou (dismissed):

heat/core/manipulations.py
ClaudiaComito: Thanks @ben-bou , I introduced the change you suggested, replaced most of the checks with `try...except` statements and am now allowing for `array_like`

ben-bou (approved):

Issue #746, opened 2021-03-16
Print toggle for local tensor printing


coquelin77:
**Related**
Possibly related to #739 but only tangentially.

**Feature functionality**
Toggle between printing the torch tensors local to each MPI process and gathering the full DNDarray on rank 0 (current).

**Additional context**
Thanks to @fschlimb for the idea in #745

PR #816, opened 2021-06-22
Features/746 print0 print toggle

coquelin77:
## Description

Adds a toggle function to switch between local printing and global printing for easier usability and debugging.

Issue/s resolved: #746 

## Changes proposed:
- added local and global printing options
- added a print0 function to both the heat namespace and the builtins

## Type of change
  - this would only be a breaking change if another function relies on global printing and that is not set explicitly 

## Due Diligence

yes -> the built in print statement will change depending on which mode is selected


coquelin77:
this might fail on coverage but im not sure how to test `print0`. any ideas for how to test this @Markus-Goetz ?

coquelin77:
rerun tests

heat/core/tests/test_printing.py
```diff
@@ -15,7 +15,44 @@ def tearDown(self):
         # reset the default device
         ht.use_device(self.device)
 
+    def test_local_printing(self):
+        ht.global_printing()
```
mtar, line 19: put in setUp() and delete in all tests

mtar (changes requested):
What do you think about printing the local metadata too? Similar to DNDarray. `LNDarray([], dtype, device, split)` or something like that. It can be useful for debugging.

mtar (approved):

Issue #825, opened 2021-06-29
setitem not working with slices and values are DNDarrays


coquelin77:
**To Reproduce**
```python
    x = ht.ones((10, 10), split=0)
    setting = ht.zeros((8, 8), split=0)
    x[1:-1, 1:-1] = setting
    print(x)
```
Traceback:
```
[0] Traceback (most recent call last):
[0]   File "testing.py", line 97, in <module>
[0]     x[1:-1, 1:-1] = setting
[0]   File "/home/daniel/.git/heat/heat/core/dndarray.py", line 1450, in __setitem__
[0]     self.__setter(tuple(key), value[tuple(value_slice)])
[0]   File "/home/daniel/.git/heat/heat/core/dndarray.py", line 846, in __getitem__
[0]     arr = self.__array[tuple(key)].reshape(lout)
[0] RuntimeError: shape '[5, 8]' is invalid for input of size 8
```

**Expected behavior**
```
[0] DNDarray([[1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
[0]           [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[0]           [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[0]           [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[0]           [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[0]           [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[0]           [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[0]           [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[0]           [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[0]           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]], dtype=ht.float32, device=cpu:0, split=0)
[1] 
```

**Version Info**
Current master (as of 29.06.2021)

**Additional comments**
Shout out to @fschlimb for bringing this to my attention

PR #826, opened 2021-06-29
Bug/825 setitem slice dndarrays

coquelin77:
## Description

Fix for setting DNDarray values with a DNDarray which has a different size in the split dimension, could use cleaning.

Issue/s resolved: #825 

## Changes proposed:
- Fix for setting DNDarray values with a DNDarray which has a different size in the split dimension
- added OOP redistribute function
- add lshape_map property (not used everywhere yet)

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -9,7 +9,7 @@ import warnings
 from inspect import stack
 from mpi4py import MPI
 from pathlib import Path
-from typing import List, Union, Tuple, TypeVar
+from typing import List, Union, Tuple, TypeVar, Optional
 
 warnings.simplefilter("always", ResourceWarning)
 
@@ -80,6 +80,7 @@ class DNDarray:
         self.__ishalo = False
         self.__halo_next = None
         self.__halo_prev = None
+        self.__lshape_map = None
 
         # check for inconsistencies between torch and heat devices
         assert str(array.device) == device.torch_device
@@ -275,6 +276,13 @@ class DNDarray:
         Returns the shape of the ``DNDarray`` on each node
         """
         return tuple(self.__array.shape)
+
+    @property
+    def lshape_map(self) -> torch.Tensor:
+        """
+        Returns the lshape map. if it has been previously created then it will be created here
+        """
+        return self.create_lshape_map()
 
     @property
     def real(self) -> DNDarray:
@@ -568,11 +576,21 @@ class DNDarray:
         self.__device = devices.cpu
         return self
 
-    def create_lshape_map(self) -> torch.Tensor:
+    def create_lshape_map(self, recreate: Optional[bool] = True) -> torch.Tensor:
         """
         Generate a 'map' of the lshapes of the data on all processes.
         Units are ``(process rank, lshape)``
+
+        Parameters
+        ----------
+        recreate : bool, optional
+            if False (default) and the lshape map has already been created, use the previous
+            result. Otherwise, create the lshape_map
+            Default: False
         """
+        if not recreate and self.__lshape_map is not None:
+            return self.__lshape_map
+
         lshape_map = torch.zeros(
             (self.comm.size, self.ndim), dtype=torch.int, device=self.device.torch_device
         )
@@ -589,6 +607,7 @@ class DNDarray:
             )
             self.comm.Allreduce(MPI.IN_PLACE, lshape_map, MPI.SUM)
 
+        self.__lshape_map = lshape_map
         return lshape_map
 
     def __float__(self) -> DNDarray:
@@ -1055,9 +1074,6 @@ class DNDarray:
                     )
                 )
         if target_map is None:  # if no target map is given then it will balance the tensor
-            target_map = torch.zeros(
-                (self.comm.size, len(self.gshape)), dtype=int, device=self.device.torch_device
-            )
             _, _, chk = self.comm.chunk(self.shape, self.split)
             target_map = lshape_map.clone()
             target_map[..., self.split] = 0
@@ -1151,6 +1167,8 @@ class DNDarray:
             # sometimes need to call the redistribute once more,
             # (in the case that the second to last processes needs to get data from +1 and -1)
             self.redistribute_(lshape_map=lshape_map, target_map=target_map)
+
+        self.__lshape_map = target_map
 
     def __redistribute_shuffle(
         self,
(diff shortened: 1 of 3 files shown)


mtar (changes requested):
Keep it up! @coquelin77 

I played around a little bit. It doesn't work when the split axes are different. Do you want to handle them too?

Some notes I took when reviewing the code. You wanted to clean it up anyway.

heat/core/dndarray.py
```diff
@@ -568,11 +576,21 @@ def cpu(self) -> DNDarray:
         self.__device = devices.cpu
         return self
 
-    def create_lshape_map(self) -> torch.Tensor:
+    def create_lshape_map(self, recreate: Optional[bool] = True) -> torch.Tensor:
```
ClaudiaComito, line 579: I might be wrong but I seem to remember  `Optional` should only be used when the default value is `None`.

should we replace `recreate` with `force-check`?

ClaudiaComito (changes requested):
I love the new `lshape_map` property. Brilliant.

A problem I'm having with `setitem` is this:
```
>>> import heat as ht
>>> a = ht.arange(10).reshape(2,5)
>>> a[1,ht.array([1])]
DNDarray([6], dtype=ht.int32, device=cpu:0, split=None)
>>> a[1,ht.array(1)]
DNDarray(6, dtype=ht.int32, device=cpu:0, split=None)
a[1,ht.array([1])] = 7
  File "<stdin>", line 1, in <module>
  File "/Users/c.comito/HAF/heat/heat/core/dndarray.py", line 1404, in __setitem__
    return self.__setter(key, value)  # returns None
  File "/Users/c.comito/HAF/heat/heat/core/dndarray.py", line 1570, in __setter
    self.__array.__setitem__(key, value)
  File "/Users/c.comito/HAF/heat/heat/core/dndarray.py", line 976, in __len__
    return self.shape[0] 
```    
This is not specific to this branch but maybe can be addressed here.

Thanks a lot!

heat/core/dndarray.py
coquelin77: i removed the optional there. you are correct

i dont view it as a check per-se. it doesnt correct an already existing one, but it creates a new one instead. but i guess that is a matter of perspective. do we use `force-check` elsewhere already?

coquelin77:
> Keep it up! @coquelin77
> 
> I played around a little bit. It doesn't work when the split axes are different. Do you want to handle them too?
> 
> Some notes I took when reviewing the code. You wanted to clean it up anyway.

do you mean to the `value` or the `key` for the different split axes?

if you mean the `value` is the one with a different split, then that is complicated, if the shapes are the right size, i dont have a problem with it. but that is easier said than done

mtar:
From the example in #825:
```python
x = ht.ones((10, 10), split=0)
setting = ht.zeros((8, 8), split=1)
x[1:-1, 1:-1] = setting
print(x)
```
This doesn't work.

This works however
 ```python
x = ht.ones((10, 10), split=0)
setting = ht.zeros((4, 8), split=1)
x[0:4, 1:-1] = setting
print(x)
```
```console
[1,0]<stdout>:DNDarray([[1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[1,0]<stdout>:          [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[1,0]<stdout>:          [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[1,0]<stdout>:          [1., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
[1,0]<stdout>:          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
[1,0]<stdout>:          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
[1,0]<stdout>:          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
[1,0]<stdout>:          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
[1,0]<stdout>:          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
[1,0]<stdout>:          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]], dtype=ht.float32, device=cpu:0, split=0)
[1,1]<stdout>:
```

coquelin77:
with some digging. the example that works in this case, only works because of an accident where all of the values are gathered onto rank 0. 

I think that setting a split DNDarray with another differently split DNDarray should be handled in another issue

heat/core/dndarray.py
ClaudiaComito: we use `force_check` in `is_balanced()` - not sure the choice of underscore over hyphen was a sound one, in hindsight.

mtar (changes requested):
Can you increase the coverage?

---
array indexing is not available `a[ht.array([0,1])] = ht.ones(2)`, but this is another issue.

heat/core/dndarray.py
```diff
@@ -1355,6 +1367,18 @@ def __setitem__(
...
         """
         key = getattr(key, "copy()", key)
+        try:
+            if value.split != self.split:
+                val_split = int(value.split)
+                sp = self.split
+                warnings.warn(
+                    f"\nvalue.split {val_split} not equal to this DNDarray's split:"
+                    f" {sp}. this may cause errors or unwanted behavior",
+                    category=RuntimeWarning,
```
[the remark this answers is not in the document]
mtar: Do you have a test for the warning?

heat/core/dndarray.py
```diff
@@ -1380,92 +1404,150 @@ def __setitem__(
...
+                if rank not in actives:
+                    return  # non-active ranks can exit here
+
+                chunk_starts_v = target_reshape_map[:, self.split]
+                value_slice = [slice(None, None, None)] * value.ndim
+                step2 = key_step if key_step is not None else 1
+                key_start = (chunk_starts_v[rank] - og_key_start).item()
+
+                if key_start < 0:
+                    key_start = 0
```
mtar, line 1500: no test?

heat/core/dndarray.py
```diff
@@ -1380,92 +1404,150 @@ def __setitem__(
...
+            key[self.split] = slice(key_start, key_stop, key_step)
+
+            # todo: need to slice the values to be the right size...
+            if isinstance(value, (torch.Tensor, type(self))):
+                # if its a torch tensor, it is assumed to exist on all processes
+                value_slice = [slice(None, None, None)] * value.ndim
+                step2 = key_step if key_step is not None else 1
+                key_start = (chunk_starts[rank] - og_key_start).item()
+                if key_start < 0:
+                    key_start = 0
```
mtar, line 1527: no test according to codecov

mtar (approved):

mtar (approved):

Issue #780, opened 2021-05-26
Implement `square`


ClaudiaComito:
**Related**
#774

**Feature functionality**
see https://numpy.org/doc/stable/reference/generated/numpy.square.html

**Additional context**
See [Python array API standard](https://data-apis.org/array-api/latest/index.html)

mtar:
#806

coquelin77:
`square` has already been implemented. it is in `exponential`

mtar:
Which definition of sign for complex numbers should we use?
[Numpy](https://numpy.org/doc/stable/reference/generated/numpy.sign.html):  <img src="https://render.githubusercontent.com/render/math?math=x/\sqrt{x \cdot x}">
[Pytorch](https://pytorch.org/docs/stable/generated/torch.sgn.html): <img src="https://render.githubusercontent.com/render/math?math=x/|x|">

PR #827, opened 2021-06-29
add sign, sgn

mtar:
## Description

This PR adds the `sign ` & `sgn` functions to heat. The former behaves like numpy on complex values, the latter like pytorch.

Issue/s resolved: #780 

## Changes proposed:
- add `sign`
- add `sgn`

## Type of change

- New feature

## Due Diligence

no

### DIFF
--- a/heat/core/rounding.py
+++ b/heat/core/rounding.py
@@ -12,7 +12,19 @@ from . import dndarray
 from . import sanitation
 from . import types
 
-__all__ = ["abs", "absolute", "ceil", "clip", "fabs", "floor", "modf", "round", "trunc"]
+__all__ = [
+    "abs",
+    "absolute",
+    "ceil",
+    "clip",
+    "fabs",
+    "floor",
+    "modf",
+    "round",
+    "sign",
+    "sgn",
+    "trunc",
+]
 
 
 def abs(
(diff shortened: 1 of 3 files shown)


mtar:
test fails on pytorch 1.8.1: nonzero can't be used with complex dtypes

coquelin77:
rerun tests

coquelin77:
it looks like the tests are failing here because of device mismatches

coquelin77:
> test fails on pytorch 1.8.1: nonzero can't be used with complex dtypes

this seems to be the case for torch 1.7 as well

mtar:
Tests updated for GPU.
Two missing lines in coverage because of code for different pytorch versions.

heat/core/rounding.py
```diff
@@ -12,7 +12,19 @@
...
+    "abs",
+    "absolute",
+    "ceil",
+    "clip",
+    "fabs",
+    "floor",
+    "modf",
+    "round",
+    "sign",
+    "sgn",
```
ClaudiaComito, line 25: order

heat/core/rounding.py
```diff
@@ -328,6 +340,90 @@ def round(
 DNDarray.round.__doc__ = round.__doc__
 
 
+def sign(x: DNDarray, out: Optional[DNDarray] = None) -> DNDarray:
+    """
+    Returns an indication of the sign of a number, element-wise. The definition for complex values is equivalent to :math:`x / \\sqrt{x \\cdot x}`.
+
```
ClaudiaComito, line 346: Mention "equivalent to numpy.sign" somewhere in the docs?

heat/core/rounding.py
ClaudiaComito, line 25: order

heat/core/rounding.py
```diff
@@ -328,6 +340,90 @@ def round(
...
+    Parameters
+    ----------
+    x : DNDarray
+        Input array
+    out : DNDarray, optional
+        A location in which to store the results.
+
```
ClaudiaComito, line 353: Mention "equivalent to torch.sgn" somewhere in the docs?

heat/core/rounding.py
```diff
@@ -328,6 +340,90 @@ def round(
...
+            out.larray.copy_(x.larray)
+            data = out.larray
+        else:
+            data = torch.clone(x.larray)
+        # NOTE remove when min version >= 1.9
+        if "1.7" in torch.__version__ or "1.8" in torch.__version__:
+            pos = data != 0
+        else:  # pragma: no cover
+            indices = torch.nonzero(data)
+            pos = torch.split(indices, 1, 1)
```
ClaudiaComito, line 381: I'm missing something for sure so forgive the possibly stupid question, but why do we need to introduce a version dependency for such a simple operation?

The following runs fine on pytorch 1.9.:

```python
pos = data != 0
data[pos] = x.larray[pos] / torch.sqrt(torch.square(x.larray[pos]))
```

ClaudiaComito (changes requested):
👏🏼 @mtar thanks a lot as usual. Just a few comments.

heat/core/rounding.py
mtar: fixed

heat/core/rounding.py
mtar: I don't think that it is really necessary because I do mention the definition for complex values. Also, PyTorch's sign function does not work for complex values.
We can add it, though, if you want to put strong emphasis on that.

heat/core/rounding.py
mtar: I don't think that it is really necessary because I do mention the definition for complex values. Also, NumPy doesn't have this function at all.
We can add it, though, if you want to put strongly emphasis on that.

heat/core/rounding.py
mtar: There's a historical reason. I started writing the function with PyTorch 1.9. and I then added the special code for 1.7/1.8 because `nonzero` is not implemented for complex values there. At the same time, Heat pinned the required version of PyTorch and I've never came back to compare the code on 1.9. again :sweat_smile:

mtar: On my local machine, the code for 1.9 is faster and uses less memory.

Issue #807, opened 2021-06-15
Implement `roll`


mtar:
Related
#774

Feature functionality
see https://numpy.org/doc/stable/reference/generated/numpy.roll.html
https://pytorch.org/docs/stable/generated/torch.roll.html

Additional context
See https://data-apis.org/array-api/latest/API_specification/manipulation_functions.html#roll-x-shift-axis-none

PR #829, opened 2021-07-05
Features/807 roll

mtar:
## Description

This PR adds roll function to Heat. It uses PyTorch if the roll is not along the split axis. It improves on it that it is now possible to use a single shift on multiple axes (int/tuple) similar to NumPy.

Issue/s resolved: #807 

## Changes proposed:
- add `roll`

## Type of change

- New feature

## Due Diligence

no

### DIFF
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -41,6 +41,7 @@ __all__ = [
     "repeat",
     "reshape",
     "resplit",
+    "roll",
     "rot90",
     "row_stack",
     "shape",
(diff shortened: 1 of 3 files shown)


mtar:
test fails on pytorch 1.8.1: RuntimeError: repeats has to be Long tensor

coquelin77:
rerun tests

coquelin77:
> test fails on pytorch 1.8.1: RuntimeError: repeats has to be Long tensor

also fails on torch 1.7 with the same error on the 8 GPU tests

heat/core/manipulations.py
```diff
@@ -1850,6 +1851,174 @@ def reshape_argsort_counts_displs(
...
+    if isinstance(shift, int):
+        if isinstance(axis, int):
+            if x.split is not None and (axis == x.split or (axis + x.ndim) == x.split):
+                # roll along split axis
+                size = x.comm.Get_size()
+                rank = x.comm.Get_rank()
+
+                lshape_map = x.create_lshape_map()[:, x.split]  # local elements along axis
+                cumsum_map = torch.cumsum(lshape_map, dim=0)  # cumulate along axis
+                indices = torch.arange(size)
```
coquelin77, line 1912: device

heat/core/manipulations.py
```diff
@@ -1850,6 +1851,174 @@ def reshape_argsort_counts_displs(
...
+            rolled = torch.roll(x.larray, shift, axis)
+            return DNDarray(
+                rolled,
+                gshape=x.shape,
+                dtype=x.dtype,
+                split=x.split,
+                device=x.device,
+                comm=x.comm,
+                balanced=x.balanced,
+            )
```
coquelin77, line 1958: it would wonderful if these loops could be switched slightly so they are not so nested. That being said, i understand why you did it this way. I am not against keeping this way, its just odd to read in the moment

coquelin77 (changes requested):

heat/core/manipulations.py
```diff
@@ -1908,6 +1909,172 @@ def reshape_argsort_counts_displs(
...
+
+    # inputs are ints
+    if isinstance(shift, int):
+        if isinstance(axis, int):
+            if x.split is not None and (axis == x.split or (axis + x.ndim) == x.split):
+                # roll along split axis
+                size = x.comm.Get_size()
+                rank = x.comm.Get_rank()
+
+                lshape_map = x.create_lshape_map()[:, x.split]  # local elements along axis
```
coquelin77: might be able to save on some communication by turning off the `force_check` parameter in this function

coquelin77 (approved):
looks great, can you add 1 more test? 3d

coquelin77:
> looks great, can you add 1 more test? 3d

sorry, i just put it in myself. i thought that upon refresh this would be deleted

Issue #828, opened 2021-07-01
getitem: type-checking of integers


ben-bou:
**Description**
I'm not sure if the calculation of the new split axis in the getitem is handled correctly. I've got the following error:
```
  File "/p/project/cslts/local/juwels/HeAT/experimental_HeAT/lib/python3.8/site-packages/heat/core/dndarray.py", line 742, in __getitem__
    isinstance(key[i], int) or len(key[i]) == 1
TypeError: object of type 'numpy.int64' has no len()
```
probably the ` isinstance(key[i], int)` is insufficient to check for integers and should be replaced (maybe numpy has something like `isinteger()`? But should also work with torch and heat ints)

edit: also occurs in line 852 and leads to returning an empty array
This probably also occurs outside of the getitem

also line 736 `if self.split is not None and len(gout_full) < self.ndim:` is not compatible with `newaxis` (might be irrelevant)

**To Reproduce**
1. Which module/class/function is affected?
`__getitem__`

**Version Info**
master

**Additional comments**
@ClaudiaComito how about this named tensor stuff to determine the new split axis?

ben-bou:
Numpy suggests using `isinstance(obj, np.integer)` to type-check ints

PR #831, opened 2021-07-06
Bug/828 getitem typechecking

ClaudiaComito:
## Description

Expanded type-checking of `key` within `getitem`, so that 1-element `array-like` objects are allowed. This includes 0-dim arrays.

Issue/s resolved: #828 

## Changes proposed:
- private function `__is_singular()` does the checking within `getitem`, thanks @ben-bou !
-
-
-

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -730,31 +730,9 @@ class DNDarray:
             slices = [slice(None)] * (self.ndim - (len(kst) + len(kend)))
             key = kst + slices + kend
 
-        def __is_singular(obj: any) -> bool:
-            """
-            Checks if obj has size one, and thereby reduces the dimensionalty of
-            the output
-            """
-            try:
-                int(obj)
-                return True
-            except Exception:
-                pass
-            if hasattr(obj, "__len__") and len(obj) == 1:
-                return True
-            if hasattr(obj, "shape") and len(obj.shape) == 0:
-                return True
-            if hasattr(obj, "numel"):
-                return obj.numel() == 1
-            if hasattr(obj, "size"):
-                return obj.size == 1
-            if hasattr(obj, "item"):
-                try:
-                    obj.item()
-                    return True
-                except Exception:
-                    pass
-            return False
+        def __is_singular(key: any, axis: int, self_proxy: DNDarray = self_proxy) -> bool:
+            zeros = tuple([0] * (self.ndim - 1))
+            return self_proxy[(*zeros[:axis], key[axis], *zeros[axis:])].ndim == 0
 
         # calculate new split axis
         new_split = self.split
@@ -764,7 +742,7 @@ class DNDarray:
                 new_split = 0
             else:
                 for i in range(len(key[: self.split + 1])):
-                    if __is_singular(key[i]):
+                    if __is_singular(key, i):
                         new_split = None if i == self.split else new_split - 1
 
         key = tuple(key)
@@ -872,7 +850,7 @@ class DNDarray:
                 lout[new_split] = 0
                 arr = torch.empty(lout, dtype=self.__array.dtype, device=self.__array.device)
 
-        elif __is_singular(key[self.split]):
+        elif __is_singular(key, self.split):
             # getting one item along split axis:
             key = list(key)
             if isinstance(key[self.split], list):
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -730,7 +730,7 @@ class DNDarray:
             slices = [slice(None)] * (self.ndim - (len(kst) + len(kend)))
             key = kst + slices + kend
 
-        def _is_singular(obj: any) -> bool:
+        def __is_singular(obj: any) -> bool:
             """
             Checks if obj has size one, and thereby reduces the dimensionalty of
             the output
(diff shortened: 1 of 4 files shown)


ClaudiaComito:
rerun tests

heat/core/dndarray.py
```diff
@@ -730,6 +730,32 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
             slices = [slice(None)] * (self.ndim - (len(kst) + len(kend)))
             key = kst + slices + kend
 
+        def __is_singular(obj: any) -> bool:
```
ben-bou, line 733: I'm sorry and don't want to cause more work, but honestly we already have the `self_proxy` at this point.
Something like the following could be 100 % torch-compatible (i've got no idea about performance or edge cases though)
```python
zeros = tuple([0] * (self.ndim -1))
is_singular = bool(1 - self_proxy[ *zeros[:i], key[i], *zeros[i:] ].ndim)
```

heat/core/tests/test_dndarray.py
```diff
@@ -1069,7 +1069,7 @@ def test_setitem_getitem(self):
         # slice in 1st dim across 1 node (2nd) w/ singular second dim
         c = ht.zeros((13, 5), split=0)
         c[8:12, 1] = 1
-        b = c[8:12, 1]
+        b = c[8:12, np.array(1)]
```
ben-bou, line 1072: Please also add one with `np.int64` instead of `np.array`

ben-bou (approved):

heat/core/dndarray.py
```diff
@@ -730,6 +730,10 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
             slices = [slice(None)] * (self.ndim - (len(kst) + len(kend)))
             key = kst + slices + kend
 
+        def __is_singular(key: any, axis: int, self_proxy: DNDarray = self_proxy) -> bool:
```
ben-bou, line 733: sanitize axis

ben-bou (approved):

ben-bou:
A similar type-checking could be employed everywhere in the code where `isinstance(obj, int)` is used (e.g. [sanitize_axis](https://github.com/helmholtz-analytics/heat/blob/c62cd072479efac8d850b85fc9e426e75fafd344/heat/core/stride_tricks.py#L100))
There's no reason why a numpy integer should not be used to specify an axis.

heat/core/dndarray.py
```diff
@@ -730,6 +730,10 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
...
+            zeros = tuple([0] * (self.ndim - 1))
```
ben-bou, line 734: `self_proxy.ndim` is better because the way it is, `self_proxy` can be something else (even not actually a proxy of self)

heat/core/dndarray.py
ben-bou: maybe rename `self_proxy` to `torch_proxy`

coquelin77 (changes requested):

heat/core/dndarray.py
```diff
@@ -730,6 +730,32 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
...
+            except Exception:
+                pass
+            if hasattr(obj, "__len__") and len(obj) == 1:
+                return True
+            if hasattr(obj, "shape") and len(obj.shape) == 0:
+                return True
+            if hasattr(obj, "numel"):
+                return obj.numel() == 1
+            if hasattr(obj, "size"):
+                return obj.size == 1
```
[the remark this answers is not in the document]
ben-bou: probably not, but i wanted to be as duck-typing friendly as possible. anyway, utilizing torch is way better because supporting ducktypes and array-likes not supported by torch would only lead to new errors

heat/core/dndarray.py
ben-bou: I think both depend on whether this code will ever be used non-internally. My reasoning is that, theoretically, `self_proxy` can be any torch.Tensor and the function will work, so why should it be named after `self`.

heat/core/dndarray.py
```diff
@@ -887,6 +881,12 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
             balanced=True if new_split is None else None,
         )
 
+    @staticmethod
+    def __getitem_is_key_singular(key: any, axis: int, self_proxy: torch.Tensor) -> bool:
```
ClaudiaComito, line 885: I agree with abstracting this method, I'm confused by the name. Something like `__key_is_singular` would work well I think, I'm sure we'll end up using this in `setitem` as well.

coquelin77:
rerun tests

heat/core/dndarray.py
coquelin77: okay that works for me. ill make the change now. my motivation behind the name was to keep it right below the getitem function

coquelin77: ive changed it to `__is_key_singular` to keep it nearby

coquelin77 (approved):
@ClaudiaComito I approve of your changes, but I think that you should also look at the minor changes that I made (i believe that it was primarily abstraction) before this is merged. I think that you should be the one to hit the button when you are ready to merge this

coquelin77 (approved):

PR #856, opened 2021-08-23
Features/torch proxy

ClaudiaComito:
## Description

I propose to implement a DNDarray method to return a 1-byte `torch.Tensor` strided as the original DNDarray global shape. At @ben-bou 's suggestion, we have been using this "proxy" tensor for sanitation, as it allows us to easily bypass many of the instance checks. It is also quite practical to assess the  shape of the output DNDarray of any operation. See getitem/setitem, or `manipulations.tile()` for examples. 

Issue/s resolved: None. Implemented with an eye on long-term maintenance / minimizing repetitions / simplifying code.

## Changes proposed:
- Implement `DNDarray.__torch_proxy__()`
- Implement tests
- replace "proxy" definitions in `dndarray` and `manipulations` with `DNDarray.__torch_proxy__()` calls
- enforce `DNDarray.size` returning `int` even if `DNDarray.ndim` is 0

## Type of change

## Due Diligence

no

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ## Feature Additions
 
+### DNDarray
+- [#856](https://github.com/helmholtz-analytics/heat/pull/856) New `DNDarray` method `__torch_proxy__`
+
 ### Linear Algebra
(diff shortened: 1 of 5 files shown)


ClaudiaComito:
rerun tests

coquelin77 (approved):

Issue #866, opened 2021-09-14
print binary operations one element nested arrays


mtar:
**Description**
A clear and concise description of the bug and the associated functionality.

Printing the result of an arithmetic operations does not work with DNDarrays when at least one operand contains a nested array.
The problem seems to be the resulting array. 

**To Reproduce**
1. Which module/class/function is affected?
print, _ operations.__binary_op
2. What are the circumstances under which the bug appears?
binary_op on distributed + nested array with one element and print
3. What is the exact error message / erroneous behavior?
RecursionError: maximum recursion depth exceeded while calling a Python object

**Expected behavior**
Print result

**Illustrative**
If applicable, add screenshots or minimal examples to help explain your problem.
```
a = ht.array([[2]], split=0) + ht.array([[1]], split=0)
print(a) # Error
print(a.larray, a.shape) # [1,0]<stdout>:tensor([[3]]) (1, 1); [1,1]<stdout>:tensor([[3]]) (1, 1)
```
**Version Info**
 Which version are you using?
master 45d4a00
**Additional comments**
Any other comments here.

PR #868, opened 2021-09-16
fix binary_op on operands with single element

mtar:
## Description

This PR fixes a known bug in _operations.__binary_op() where the data of an DNDarray is copied to all processors if it only contains one element.

The fix introduces the method 'Split()' to MPICommunicator for splitting it into new ones. These new ones are used for the data transfer in __binary_op()

Issue/s resolved: #866 

## Changes proposed:
- add MPICommunicator.Split()
- fix data transfer in __binary_op()
-
-

## Type of change

- Bug fix / New feature

## Due Diligence

no

### DIFF
--- a/heat/core/_operations.py
+++ b/heat/core/_operations.py
@@ -9,7 +9,6 @@ import warnings
 
 from .communication import MPI, MPI_WORLD
 from . import factories
-from . import devices
 from . import stride_tricks
 from . import sanitation
 from . import statistics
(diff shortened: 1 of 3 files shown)


heat/core/_operations.py
```diff
@@ -111,28 +110,31 @@ def __binary_op(
...
                     # )
-                    if t1.comm.rank > 0:
+                    color = 0 if t1.comm.rank < t2.shape[t1.split] else 1
+                    newcomm = t1.comm.Split(color, t1.comm.rank)
+                    if t1.comm.rank > 0 and color == 0:
                         t1.larray = torch.zeros(
                             t1.shape, dtype=t1.dtype.torch_type(), device=t1.device.torch_device
                         )
-                    t1.comm.Bcast(t1)
+                    newcomm.Bcast(t1)
```
coquelin77, line 124: i dont see why we need a split. can you explain that to me?

Also, splits can cause a little bit of overhead in the general code. If you are splitting for this one communication, it might have other effects down the line

heat/core/communication.py
```diff
@@ -436,6 +436,19 @@ def alltoall_recvbuffer(
 
         return self.as_mpi_memory(obj), (recvcount, recvdispls), recvtypes
 
+    def Split(self, color: int = 0, key: int = 0):
```
coquelin77, line 439: return tag for docs

mtar: fixed

heat/core/_operations.py
mtar: I address the problem described by comment in line 114. At the moment rank 0 broadcasts the data of an operand to all other processes causing problems as seen in the linked issue. I split the communicator such that the broadcast from rank 0 is now restricted to the ones that will need the data. On the other ranks an empty array is broadcast leaving them unchanged.

coquelin77: doing this in the abstracted operation can cause performance issues. lets say that you are doing a lot of arithmetic operations where this is the case, then you would be creating a new mpi split comm for each one. they are not cleaned up by the normal garbage collector, so if we do it this way, the new communicator needs to be Finalized and removed from the MPI runtime.

mtar: Would a call to comm.Free() be enough?

coquelin77: should be

Issue #529, opened 2020-04-06
Clean up Lasso file, make examples work


Markus-Goetz:

coquelin77:
bump, is this done?

PR #876, opened 2021-10-04
Bugfix/529 lasso example

bhagemeier:
## Description

Make examples work. The code examples in the examples folder were not fully aligned with some changes, e.g. paths to example data and Python path setup. Additionally, at least the Lasso example requires matplotlib and scikit-learn, which can now be pulled in as optional dependencies.

Issue/s resolved: #529 

## Changes proposed:
- path adaptation (Lasso and KNN examples)
- optional dependencies to run examples

## Type of change

- Bug fix

## Due Diligence

Only example code has been modified.

no

skip ci


coquelin77:
run tests

coquelin77:
rerun tests

PR #877, opened 2021-10-04
Feature/337 determinant

mtar:
## Description

First implementation of determinant computation using Gaussian elimination with partial pivoting.

Issue/s resolved: #337 

## Changes proposed:
- add linalg.det

## Type of change

- New feature

## Due Diligence

no

### DIFF
--- a/heat/core/linalg/basics.py
+++ b/heat/core/linalg/basics.py
@@ -43,72 +43,111 @@ __all__ = [
 def det(a: DNDarray) -> DNDarray:
     """
     Returns the determinant of a square matrix.
+
+    Parameters
+    ----------
+    a : DNDarray
+        A square matrix or a stack of matrices. Shape = (...,M,M)
+
+    Raises
+    ------
+    TypeError
+        If the dtype of 'a' is not floating-point.
+    ValueError
+        If `a.ndim < 2` or if the length of the last two dimensions is not the same.
+
+    Examples
+    --------
+    >>> a = ht.array([[-2,-1,2],[2,1,4],[-3,3,-1]])
+    >>> ht.linalg.det(a)
+    DNDarray(54., dtype=ht.float64, device=cpu:0, split=None)
     """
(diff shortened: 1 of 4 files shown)


heat/core/linalg/basics.py
```diff
@@ -40,6 +41,116 @@
...
+        If `a.ndim < 2` or if the length of the last two dimensions is not the same.
+
+    Examples
+    --------
+    >>> a = ht.array([[-2,-1,2],[2,1,4],[-3,3,-1]])
+    >>> ht.linalg.det(a)
+    DNDarray(54., dtype=ht.float64, device=cpu:0, split=None)
+    """
+    # no split in the square matrices
+    if a.split is None or a.split < a.ndim - 2:
```
coquelin77, line 67:
```suggestion
    if not a.is_distributed() or a.split < a.ndim - 2:
```
might need to change the error messages to be consistent

coquelin77 (changes requested):
looks great to me. Just a small change

coquelin77:
@mtar i have changed the shape rules to line up with torch's results. can you check this again?

mtar:
@coquelin77 Thanks for addressing the wrong dimension of the result tensor.

I was now able to reduce some lines of code due to a fix in master.

heat/core/linalg/basics.py
```diff
@@ -155,6 +156,90 @@ def cross(
...
+            a.shape[:-2],
+            types.heat_type_of(data),
+            split=sp,
+            device=a.device,
+            comm=a.comm,
+            balanced=a.balanced,
+        )
+
+    if a.ndim < 2:
+        raise RuntimeError("DNDarray must be at least two-dimensional.")
```
mtar, line 196: I'm not sure if we should follow the error types of PyTorch. It's RuntimeError mostly.

coquelin77 (approved):

Issue #880, opened 2021-10-08
Binary operations: Support differing lshape-maps


ben-bou:
**Feature functionality**
A clear and concise description of what the feature should do exactly and any alternative solutions.

Example:
```python
a =  ht.ones(10, split=0)
b = ht.zeros(10, split=0)
c = a[ :-1 ] + b[ 1: ]
```
Globally, `a[ :-1 ]` and `b[ 1: ]`  are compatible and the binop should compute. However, because they are differently distributed, it doesn't work. Because this cannot work without additional communication, maybe a warning should be given. A workaround I'm currently using is:
```python
try:
  binop(op, arr1, arr2)
except:
  arr2.redistribute(arr1.lshape_map)
  result.larray = op(arr1.larray, arr2.larray)
```
(more details: https://github.com/ben-bou/heat/blob/redistribute/heat/core/_operations.py#L33 )

**Additional context**

This feature is important to ensure that scripts running with one process also run when using multiple processes.

PR #887, opened 2021-11-05
Support different lshape maps in binary ops

ClaudiaComito:
## Description

Implementing support for binary operations on distributed DNDarrays  of same shape, same split, but different distribution map, i.e.:
```python
a =  ht.ones(10, split=0)
b = ht.zeros(10, split=0)
c = a[:-1] + b[1:]
```

Issue/s resolved: #880 

## Changes proposed:
- `redistribute_` one of the two DNDarrays to match the distribution map of the other.
- expand tests
- include a warning in the `__binary_op` documentation

## Type of change

## Memory requirements

Memory requirements are consistent with the extra memory requirements of `DNDarray.redistribute_()`.

## Performance

Performance slow-down is consistent with redistributing a DNDarray before a binary operation.

## Due Diligence

no, but it redistributes one of the operands in place.

### DIFF
--- a/heat/core/_operations.py
+++ b/heat/core/_operations.py
@@ -192,10 +192,6 @@ def __binary_op(
     else:
         raise NotImplementedError("Not implemented for non scalar")
 
-    # sanitize output
-    if out is not None:
-        sanitation.sanitize_out(out, output_shape, output_split, output_device)
-
     if t1.split is not None:
         output_split = t1.split
         # TODO: implement `dndarray.create_bulk_lshape_maps`
@@ -207,7 +203,6 @@ def __binary_op(
             if (
                 t2.split is not None
                 and not (t2.lshape_map[:, t2.split] == t1.lshape_map[:, t1.split]).all()
-                and t2.lshape_map[:, t2.split].sum() == t1.lshape_map[:, t1.split].sum()
             ):
                 t2.redistribute_(target_map=t1.lshape_map)
             result = operation(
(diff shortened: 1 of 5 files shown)


heat/core/_operations.py
```diff
@@ -119,29 +154,37 @@ def __binary_op(
...
-                    color = 0 if t1.comm.rank < t2.shape[t1.split] else 1
-                    newcomm = t1.comm.Split(color, t1.comm.rank)
-                    if t1.comm.rank > 0 and color == 0:
-                        t1.larray = torch.zeros(
-                            t1.shape, dtype=t1.dtype.torch_type(), device=t1.device.torch_device
-                        )
-                    newcomm.Bcast(t1)
-                    newcomm.Free()
-
+                    t1.resplit_(None)
```
ben-bou, line 157: Why is this needed? It should work the same without this resplit, no?
Nothing below depends on having at least two elements in the split dimension

heat/core/_operations.py
```diff
@@ -153,23 +196,43 @@ def __binary_op(
...
+        # TODO: implement `dndarray.create_bulk_lshape_maps`
+        t1.create_lshape_map()
+        t2.create_lshape_map()
         if len(t1.lshape) > t1.split and t1.lshape[t1.split] == 0:
             result = t1.larray.type(promoted_type)
         else:
+            if (
+                t2.split is not None
+                and not (t2.lshape_map[:, t2.split] == t1.lshape_map[:, t1.split]).all()
+                and t2.lshape_map[:, t2.split].sum() == t1.lshape_map[:, t1.split].sum()
```
ben-bou, line 210: The last condition is not necessary:
1) `output_shape` is already determined, so shapes are compatible
2) no (shape)broadcasting is happening in the split dimension because of lines 152 to 175
-> elements in split-dimension are already equal

ClaudiaComito:
Changes addressed in #902 by @ben-bou

PR #888, opened 2021-11-07
Allow unbalanced and unequally balanced dndarrays for binops

ben-bou:
## Description

Failed tests:  
1) checks split axis after adding empty dimensions
2) resplitting to avoid (shape) broadcasting in split-dimension
3) binop with np.ndarray results in TypeError instead of NotImplementedError

Issue/s resolved: #880

## Changes proposed:
- check lshape-map before binop instead of after try...except because not all processes necessarily fail so synchronization is necessary anyway
- redistribute OUT-OF-PLACE - binops should not alter their arguments
- add support for unbalanced array when the other one is not split
- check arrays for having the same split axis AFTER (shape)broadcasting added empty dimensions (?)
- restructure input sanitation; reduce nested ifs

## Type of change
--->

## Performance
Checks lshape-maps inside every binop between distributed dndarrays -> more communication
Redistributes out-of-place -> more memory

## Due Diligence

yes / no

skip ci

### DIFF
--- a/heat/core/_operations.py
+++ b/heat/core/_operations.py
@@ -109,77 +109,106 @@ def __binary_op(
     # t2 = t2[tuple([None] * (len(output_shape) - t2.ndim))]
 
     # determine output params, transform to output shape
-    if t1.split is not None and t1.shape[t1.split] == output_shape[t1.split]:
-        # t1 is "dominant"
-        output_split = t1.split
-        output_device = t1.device
-        output_comm = t1.comm
-        if out is None:
-            t2 = sanitation.sanitize_distribution(t2, target=t1)
-            output_balanced = t1.balance
-    elif t2.split is not None and t2.shape[t2.split] == output_shape[t2.split]:
-        # t2 is "dominant"
-        output_split = t2.split
-        output_device = t2.device
-        output_comm = t2.comm
-        if out is None:
-            t1 = sanitation.sanitize_distribution(t1, target=t2)
-            output_balanced = t2.balance
-    elif t1.split is not None and t2.split is not None:
-        if t1.split != t2.split:
-            raise NotImplementedError(
-                "DNDarrays must have the same split axes, found {} and {}".format(
-                    t1.split, t2.split
-                )
-            )
+    # if t1.split is not None and t1.shape[t1.split] == output_shape[t1.split]:
+    #     # t1 is "dominant"
+    #     output_split = t1.split
+    #     output_device = t1.device
+    #     output_comm = t1.comm
+    #     if out is None:
+    #         t2 = sanitation.sanitize_distribution(t2, target=t1)
+    #         output_balanced = t1.balance
+    # elif t2.split is not None and t2.shape[t2.split] == output_shape[t2.split]:
+    #     # t2 is "dominant"
+    #     output_split = t2.split
+    #     output_device = t2.device
+    #     output_comm = t2.comm
+    #     if out is None:
+    #         t1 = sanitation.sanitize_distribution(t1, target=t2)
+    #         output_balanced = t2.balance
+    # elif t1.split != t2.split:
+    #     raise NotImplementedError(
+    #         "DNDarrays must have the same split axes, found {} and {}".format(
+    #             t1.split, t2.split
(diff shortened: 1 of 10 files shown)


mtar:

heat/core/_operations.py
```diff
@@ -62,194 +62,202 @@ def __binary_op(
...
     """
+    # Check inputs
+    if not np.isscalar(t1) and not isinstance(t1, DNDarray):
+        raise TypeError(
+            "Only tensors and numeric scalars are supported, but input was {}".format(type(t1))
+        )
+    if not np.isscalar(t2) and not isinstance(t2, DNDarray):
+        raise TypeError(
+            "Only tensors and numeric scalars are supported, but input was {}".format(type(t2))
+        )
```
mtar, line 75: At some point, we were thinking to allow interchangeable operations with PyTorch tensors / NumPy arrays. We can try to create DNDarrays out of the inputs here.

heat/core/_operations.py
```diff
@@ -62,194 +62,202 @@ def __binary_op(
...
+    sanitation.sanitize_in(t2)
+
+    # Make inputs have the same dimensionality
+    output_shape = stride_tricks.broadcast_shape(t1.shape, t2.shape)
+    # Broadcasting allows additional empty dimensions on the left side
+    # TODO simplify this once newaxis-indexing is supported to get rid of the loops
+    while len(t1.shape) < len(output_shape):
+        t1 = t1.expand_dims(axis=0)
+    while len(t2.shape) < len(output_shape):
+        t2 = t2.expand_dims(axis=0)
```
mtar, line 107: np.expand_dims() allows tuples for 'axis'. We will have to improve our implementation of this function.

heat/core/dndarray.py
```diff
@@ -962,6 +968,12 @@ def __is_key_singular(key: any, axis: int, self_proxy: torch.Tensor) -> bool:
         zeros = tuple([0] * (self_proxy.ndim - 1))
```
mtar, line 968:
```suggestion
        zeros = (0,) * (self_proxy.ndim - 1)
```

heat/core/dndarray.py
```diff
@@ -962,6 +968,12 @@ def __is_key_singular(key: any, axis: int, self_proxy: torch.Tensor) -> bool:
...
         return self_proxy[(*zeros[:axis], key[axis], *zeros[axis:])].ndim == 0
 
+    @staticmethod
+    def ___key_adds_dimension(key: any, axis: int, self_proxy: torch.Tensor) -> bool:
+        # determine if the key adds a new dimension
+        zeros = tuple([0] * (self_proxy.ndim - 1))
```
mtar, line 974:
```suggestion
        zeros = (0,) * (self_proxy.ndim - 1)
```

mtar (commented):
I appreciate this PR very much. I had some issues while I was testing it.

Here is some special case that is not working at the moment:
```python
a = ht.array([0],split=0)
b = ht.array([[1,2,3]],split=0)
a+b
```

One further request: It would be nice to get setitem working with it
```python
a = ht.ones((4,4), split=1)
b = ht.ones((4,4), split=1)
c = ht.zeros((4,4), split=1)
c[1:,1:] = a[1:,1:] + b[1:,1:]

ben-bou:
> 

Thanks for your feedback! I've got one question regarding this:

> Here is some special case that is not working at the moment:
> 
> ```python
> a = ht.array([0],split=0)
> b = ht.array([[1,2,3]],split=0)
> a+b
> ```

Because the resulting error `NotImplementedError: DNDarrays must have the same split axes, found 0 and 1` seems accurate, I'm not sure what you would expect to happen:

- Singular items (`a`) to always work, regardless of the split (i.e. `resplit(None)`)
- Tensors which will be broadcast in their own split-dimension to be resplit:
            Splti-dimension of `a` (1) will be expanded to match dimension 1 of `b` (3) and therefore should be redistributed
- If the split-dimension of the result does not yet exist in a Tensor (will be added by the broadcasting) it should be redistributed

ben-bou:
Moved to #902

Issue #886, opened 2021-11-03
implement `linalg.cholesky`


mtar:
**Feature functionality**
computes the cholesky decomposition
Intended scope: 
- real/complex matrices 
- `split` $\in${ `0,1,None`}
- of course, appropriate scalability of the routine is desired

**Additional context**

https://numpy.org/doc/stable/reference/generated/numpy.linalg.cholesky.html
https://pytorch.org/docs/stable/generated/torch.linalg.cholesky.html
https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?#linalg-cholesky-x-upper-false

After obtaining a triangular matrix by Cholesky, one often uses triangular solvers #1096

mrfh92:
My suggestion: implement "standard" Cholesky (see, e.g., https://en.wikipedia.org/wiki/Cholesky_decomposition) but block-wise using PyTorch for the blocks; in this sense, the formulas from wikipedia can be used but the $L_{i,j}$'s and $A_{i,j}$'s are blocks (i.e. matrices) instead of numbers.

PR #893, opened 2021-12-07
Feature/886 cholesky

mtar:
## Description

A first idea of the cholesky decomposition. It does not check if the input is a Hermitian, positive-definite matrix.

*It doesn't work for distributed matrices with complex values.*

Issue/s resolved: #886

## Changes proposed:
- cholesky decomposition

## Due Diligence

No

**Note**
Some lines are not covered due to code for pytorch versions between 1.7 and 1.10


mtar:
run test

mtar:
run tests

mtar:
rerun tests

mtar:
rerun tests

mtar:
run tests

mtar:
run tests

mtar:
rerun tests

mtar:
run tests

CHANGELOG.md
```diff
@@ -30,7 +28,8 @@
 - [#846](https://github.com/helmholtz-analytics/heat/pull/846) New features `norm`, `vector_norm`, `matrix_norm`
 - [#850](https://github.com/helmholtz-analytics/heat/pull/850) New Feature `cross`
 - [#877](https://github.com/helmholtz-analytics/heat/pull/877) New feature `det`
-- [#875](https://github.com/helmholtz-analytics/heat/pull/875) New feature `inv`
+- [#875](https://github.com/helmholtz-analytics/heat/pull/875) New feature `inv
```
coquelin77, line 31:
```suggestion
- [#875](https://github.com/helmholtz-analytics/heat/pull/875) New feature `inv`
```

heat/core/linalg/basics.py
```diff
@@ -44,6 +45,111 @@
 ]
 
 
+def cholesky(x: DNDarray, upper: bool = False) -> DNDarray:
+    """
+    Returns the Cholesky decomposition of a Hermitian or symmetric positive-definite matrix.
```
coquelin77, line 50: add a note saying this doesnt work for complex values (also change the examples)

mtar:
run tests

mtar:
run tests

mtar:
Can't increase coverage, missing lines require pytorch 1.8/1.9

heat/core/linalg/basics.py
```diff
@@ -44,6 +45,118 @@
...
```
ClaudiaComito, line 50: Inconsistent documentation (decomposition of a Hermitian doesn't work on complex data types)

heat/core/linalg/basics.py
```diff
@@ -44,6 +45,118 @@
...
+
+    Parameters
+    ----------
+    x : DNDarray
+        Symmetric or Hermitian positive-definite matrix. Shape = (...,M,M)
+    upper : bool, optional
+        Whether a lower or upper triangular matrix is returned.
```
ClaudiaComito, line 57: "whether an upper triangular ... is returned"

heat/core/linalg/basics.py
```diff
@@ -44,6 +45,118 @@
...
+    DNDarray([[ (1+0j), (-0-2j)],
+              [     2j,  (5+0j)]], dtype=ht.complex64, device=cpu:0, split=None)
+    >>> L = ht.linalg.cholesky(A)
+    >>> L
+    DNDarray([[(1+0j),     0j],
+              [    2j, (1+0j)]], dtype=ht.complex64, device=cpu:0, split=None)
+    >>> L @ L.T.conj()
+    DNDarray([[(1+0j),    -2j],
+              [    2j, (5+0j)]], dtype=ht.complex64, device=cpu:0, split=None)
+    """
```
ClaudiaComito, line 76: Inconsistent docs

heat/core/linalg/basics.py
```diff
@@ -44,6 +45,118 @@
...
+    if x.ndim < 2:
+        raise RuntimeError("DNDarray must be at least two-dimensional.")
+
+    m, n = x.shape[-2:]
+
+    if m != n:
+        raise RuntimeError("Last two dimensions of the DNDarray must be square.")
```
ClaudiaComito, line 83: How about "Expected a DNDarray of shape (..., M, M), got ..." or "Last two dimensions of DNDarray must be equal, are ..."

heat/core/linalg/basics.py
```diff
@@ -44,6 +45,118 @@
...
+
+    if types.heat_type_is_exact(x.dtype):
+        raise NotImplementedError("Not implemented for {}".format(x.dtype))
```
ClaudiaComito, line 86: Why not cast DNDarray to corresponding float?

heat/core/linalg/basics.py
```diff
@@ -44,6 +45,118 @@
...
+
+        return DNDarray(
+            data, x.shape, types.heat_type_of(data), x.split, x.device, x.comm, x.balanced
+        )
+
+    # NOTE There is some issues with complex values. conj() is omitted in the computation.
+    if types.heat_type_is_complexfloating(x.dtype):
+        raise NotImplementedError(
+            "Not implemented for distributed matrices of type {}".format(x.dtype)
+        )
```
ClaudiaComito, line 109: We need to  diagnose the problem.

heat/core/linalg/basics.py
```diff
@@ -44,6 +45,118 @@
...
+                    * manipulations.expand_dims(b, -2)
+                )
+            else:
+                a[..., i + 1 :, i].larray /= manipulations.expand_dims(a[..., i, i], -1).larray
+                b = a[..., i + 1 :, i]
+                b.balance_()
+                L = tril(
+                    manipulations.expand_dims(b, -1)
+                    * manipulations.expand_dims(manipulations.resplit(b, None), -2)
+                )
```
ClaudiaComito, line 131: I find it problematic that `b` needs to be gathered. This is basically an outer product, isn't it? Distributed `outer` is implemented, and we should be able to expand that functionality to a stack of vectors.

(Also it shouldn't be necessary to balance just before resplitting, if it is let's fix it)

heat/core/linalg/basics.py
```diff
@@ -44,6 +45,118 @@
...
+        else:
+            if upper:
+                a[..., i, i + 1 :].larray /= manipulations.expand_dims(a[..., i, i], -1).larray
+                b = a[..., i, i + 1 :]
+                b.balance_()
+                L = triu(
+                    manipulations.expand_dims(manipulations.resplit(b, None), -1)
+                    * manipulations.expand_dims(b, -2)
+                )
```
ClaudiaComito, line 140: See above

ClaudiaComito (changes requested):
Thanks a lot for addressing this @mtar, and sorry for taking so long to review. 

I think we need to address the complex datatype issue and the need for gathering data before merging this.  Let's discuss during today's PR talk.

ClaudiaComito:
@mtar as far as I can tell, `cholesky()` works perfectly on distributed positive-definite Hermitians as long as they are double precision.

Here's the code I'm running:

```python

# test CHOLESKY
a = ht.arange(100, dtype=ht.complex128, split=0)
a = ht.diag(a)
a[ht.where(a == 0)] = 1j
# a.resplit_(axis=1) # uncomment to test split along columns
b = a@a.conj().T # b is positive-definite Hermitian

np_b = b.numpy()

ht_cho = ht.linalg.cholesky(b)
np_cho = np.linalg.cholesky(np_b)
t_cho =  torch.linalg.cholesky(torch.tensor(np_b))

print("ht == np ?", np.allclose(ht_cho.numpy(), np_cho))
print("ht == t ?", torch.allclose(torch.tensor(ht_cho.numpy()), t_cho))
```

mrfh92 (commented):
**Review of "Feature/886-cholesky"**
The code under review adds Cholesky decomposition to the linalg-module of Heat. For two-dimensional inputs, Cholesky decomposition is computed utilizing a serial algorithm; for three- or more-dimensional inputs the decomposition is computed in batch-wise manor applying `torch.linalg.cholesky` to the last two dimensions of the input. Complex inputs are excluded so far.  

- I do not know applications in which we would have to deal with complex data. Therefore, from my point of view it does not make problems if we exclude (at least for the moment) complex data types for Cholesky although they are possible in general 
- Just a comment: I like the distinction between `dim = 2` and `dim > 2` very much. I guess, we should add this type of “batch-linear-algebra” for `dim > 2` for other routines as well.  
- The code for `dim=2` look very “serial”. Indeed, a scaling test on up to 24 processes (2 threads each) on 2 nodes of HDFML yields rather unexpected scaling behavior (which may also be due to the small size of the test matrices…).
- As far as I see, there is no error message for the case that the matrix turns out not to be positive definite. Moreover, it is not checked whether the matrix is symmetric. Since the code targets non-expert-users this may be useful, even if expensive. 

All in all, I suggest that we decide in the PR meeting together whether this PR should be merged as it is (in order to add the functionality) or whether we want to wait for a distributed implementation of Cholesky (for `dim >2`) possibly originating from GSoC. 

cpu

mrfh92:
Closed due to decision in second-last PR talk.

Issue #537, opened 2020-04-09
Better inclusion of license file


bhagemeier:
**Description**
We currently install README.md and LICENSE in the root of the installation directory, e.g. root of the virtual environment that we use. This originated from a poor example, but must be avoided.

**To Reproduce**
1. pip install heat
2. check virtual environment for README.md and LICENSE.

**Expected behavior**
These files should not be there. REAME.md is not needed in an installation. LICENSE should be put in package metadata, as is done by other packages, too.

**Version Info**
ANY

**Additional comments**
Any other comments here.

PR #894, opened 2021-12-08
Features/537 better inclusion of license file

bhagemeier:
## Description

This PR changes several two aspects of installing and using HeAT. 

1. get rid of README.md and LICENSE files in the install root / PREFIX
2. make the already packaged data sets available as a package resource

Issue/s resolved: #537 

## Changes proposed:
- 

## Type of change
- Bug fix
- New feature

## Memory requirements
n/a

## Performance
n/a

## Due Diligence

no

skip ci


coquelin77 (approved):
very cool. thanks for the fix

coquelin77:
this is only failing because the CI is not running on the GPUs. However, this PR does not do any calculation, thus it does not require the tests to run there. I will merge this

PR #899, opened 2021-12-16
Feature/844 cross axisa axisb

ClaudiaComito:
## Description

Introducing changes to accomodate numpy's `axisa`, `axisb`, `axisc` into @mtar 's `ht.cross` implementation.

Still TODO: documentation update, expand on tests, memory/performance checks.
 
Issue/s resolved: #

## Changes proposed:
-
-
-
-

## Type of change

## Due Diligence

no

### DIFF
--- a/heat/core/linalg/basics.py
+++ b/heat/core/linalg/basics.py
@@ -146,17 +146,8 @@ def cross(
 
         axisc = stride_tricks.sanitize_axis(x1.shape, axisc)
 
-        if axisc != axisa:
-            x1_permute_axes = list(range(len(x1.shape)))
-            x1_permute_axes.remove(axisa)
-            x1_permute_axes = x1_permute_axes[:axisc] + [axisa] + x1_permute_axes[axisc:]
-            x1 = x1.transpose(x1_permute_axes)
-
-        if axisc != axisb:
-            x2_permute_axes = list(range(len(x2.shape)))
-            x2_permute_axes.remove(axisb)
-            x2_permute_axes = x2_permute_axes[:axisc] + [axisb] + x2_permute_axes[axisc:]
-            x2 = x2.transpose(x2_permute_axes)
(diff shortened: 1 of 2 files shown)


heat/core/linalg/basics.py
```diff
@@ -74,45 +82,88 @@ def cross(x1: DNDarray, x2: DNDarray, axis: int = -1) -> DNDarray:
...
-    if not isinstance(axis, int):
-        try:
-            axis = int(axis)
-        except Exception:
-            raise TypeError("'axis' must be an integer.")
+    x1_2d, x2_2d = False, False
+    x1_shape, x2_shape = list(x1.shape), list(x2.shape)
 
-    axis = stride_tricks.sanitize_axis(x1.shape, axis)
+    if not axis == -1 or torch.unique(torch.tensor([axisa, axisb, axisc, axis])).numel() == 1:
```
mtar, line 95: If the default value of `axis` is set to `None`, the check will be shortened.

ClaudiaComito: `axis = None` is not compatible with the [Python array API](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?highlight=cross#linalg-cross-x1-x2-axis-1) though.

If we leave `axis=-1` as default, we're compatible with the array API and `cross` will still work as a numpy drop-in replacement.

ClaudiaComito: > If the default value of `axis` is set to `None`, the check will be shortened. Also, some redundant code can be avoided when the axes parameter are set beforehand to the value of it.
> 
> ```python
> if axis is not None:
>     axis = stride_tricks.sanitize_axis(x1.shape, axis)
>     axisa, axisb, axisc = (axis,) * 3
> ```

You're right about the redundant code though, I replaced it with your suggestion.

Issue #913, opened 2022-02-08
`ht.random.randint` `size` type requirement inconsistent with NumPy API


ClaudiaComito:
**Description**
`int` type should be allowed for `size` argument of `ht.random.randint()` 

```python
>>> import numpy as np
>>> import heat as ht
# numpy
>>> a = np.random.randint(0, 50, 20)
>>> a
array([34, 30, 43, 37, 39, 15,  9, 19, 38, 19, 12, 14, 41, 32, 44,  2,  5,
        0,  6, 45])
>>> a = np.random.randint(0, 50, size=20)
>>> a
array([10,  6, 38, 19, 10, 44,  9, 40, 23, 40, 23, 46, 41, 40,  7,  3, 15,
       25, 39, 46])

#heat
>>> b = ht.random.randint(0, 50, 20)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/c.comito/HAF/heat/heat/core/random.py", line 531, in randint
    shape = tuple(int(ele) for ele in size)
TypeError: 'int' object is not iterable

>>> b = ht.random.randint(0, 50, size=20)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/c.comito/HAF/heat/heat/core/random.py", line 531, in randint
    shape = tuple(int(ele) for ele in size)
TypeError: 'int' object is not iterable

 b = ht.random.randint(0, 50, size=(20,))
>>> b
DNDarray([13, 33,  0, 41,  6, 40, 39,  5, 34, 44,  3, 39,  1, 44, 25, 14, 25, 39, 41,  8], dtype=ht.int32, device=cpu:0, split=None)
>>> 
```

**To Reproduce**

see examples above

**Expected behavior**
it should be possible to pass an `int` as `size` argument, also in accordance with the function's type hints: `size: 'Optional[Union[int, Tuple[int]]]' = None`

**Version Info**
main branch

PR #916, opened 2022-02-08
`randint` accept ints for 'size'

mtar:
## Description

Issue/s resolved: #913 

## Changes proposed:
- allow ints for size in randint

## Type of change

- Bug fix

## Memory requirements

n/a

## Performance

n/a

## Due Diligence

no

### DIFF
--- a/heat/core/random.py
+++ b/heat/core/random.py
@@ -527,13 +527,13 @@ def randint(
 
     # sanitize shape
     if size is None:
-        size = (1,)
+        size = ()
     try:
         shape = tuple(int(ele) for ele in size)
     except TypeError:
         shape = (int(size),)
     else:
-        if not all(ele > 0 for ele in shape):
+        if not all(ele >= 0 for ele in shape):
             raise ValueError("negative dimensions are not allowed")
 
     # sanitize the data type
--- a/heat/core/random.py
+++ b/heat/core/random.py
@@ -528,9 +528,13 @@ def randint(
     # sanitize shape
     if size is None:
         size = (1,)
-    shape = tuple(int(ele) for ele in size)
-    if not all(ele > 0 for ele in shape):
-        raise ValueError("negative dimensions are not allowed")
+    try:
+        shape = tuple(int(ele) for ele in size)
+    except TypeError:
+        shape = (int(size),)
+    finally:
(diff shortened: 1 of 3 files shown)


heat/core/random.py
```diff
@@ -528,9 +528,13 @@ def randint(
...
-    shape = tuple(int(ele) for ele in size)
-    if not all(ele > 0 for ele in shape):
-        raise ValueError("negative dimensions are not allowed")
+    try:
+        shape = tuple(int(ele) for ele in size)
+    except TypeError:
+        shape = (int(size),)
+    else:
+        if not all(ele > 0 for ele in shape):
+            raise ValueError("negative dimensions are not allowed")
```
ClaudiaComito, line 538: I was thinking of addressing this like we do for `shape` in `ht.reshape()`:
´´´python
def reshape(a: DNDarray, *shape: Union[int, Tuple[int, ...]], **kwargs) -> DNDarray:
   ...
´´´

Default `size` should be `0`, not `None`. Here a numpy example:
```bash
>>> d = np.random.randint(0,50)
>>> d
1
>>> d = np.random.randint(0,50, 1)
>>> d
array([0])
>>> e = ht.random.randint(0,50)
>>> e
DNDarray([9], dtype=ht.int32, device=cpu:0, split=None)  # ndim=1 but should be 0!
```

heat/core/tests/test_random.py
```diff
@@ -240,6 +240,10 @@ def test_randint(self):
         b = ht.zeros((10,), dtype=ht.int64, split=0)
         self.assertTrue(ht.equal(a, b))
 
+        # size parameter allows int arguments
+        a = ht.random.randint(1, size=10, split=0, dtype=ht.int64)
+        self.assertTrue(ht.equal(a, b))
+
```
ClaudiaComito, line 246: Add tests where no size is specified and scalar is returned?

ClaudiaComito (changes requested):
Thanks a lot @mtar for looking into this right away! I found another inconsistency vs. numpy, see comments.

mtar:
rerun tests

mtar:
reru tests

mtar:
run tests

PR #923, opened 2022-02-22
Features/914 adv indexing

ben-bou:
## Changes proposed:
- preprocess key to find Advanced indexing, `newaxis` indexing and ellipsis 
- change `torch_proxy` By adding empty dimensions where Advanced indexing should be. In Order to reconstruct the correct shape, Put the shape in the dimensions name. 
- simplify slice indexing by slicing a `range_proxy`. 
-

## Notes 
- DNDarray is Not an instance of Iterable
- `iter(dndarray)` works anyway 
- `iter(0-dim dndarray)` also works, but shouldn‘t
- 

skip ci

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -9,7 +9,7 @@ import warnings
 from inspect import stack
 from mpi4py import MPI
 from pathlib import Path
-from typing import List, Union, Tuple, TypeVar, Optional
+from typing import List, Union, Tuple, TypeVar, Optional, Iterable
 
 warnings.simplefilter("always", ResourceWarning)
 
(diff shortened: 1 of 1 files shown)


heat/core/dndarray.py
```diff
@@ -653,6 +653,84 @@ def fill_diagonal(self, value: float) -> DNDarray:
...
+        - has the same dimensionality as the ``DNDarray`` it indexes
+
+        Parameters
+        ----------
+        key : int, slice, Tuple[int,...], List[int,...]
+            Indices for the tensor.
+        """
+        advanced_indexing = False
+        if isinstance(key, DNDarray):
+            # DNDARRAY CURRENTLY DOES NOT IMPLEMENT the Iterable interface, need to define __iter__()
```
ClaudiaComito, line 671: We could implement the Iterable interface for non-distributed DNDarrays, but we need to give some thought to what to fall back on when `split is not None`.

heat/core/dndarray.py
```diff
@@ -653,6 +653,84 @@ def fill_diagonal(self, value: float) -> DNDarray:
...
+            advanced_indexing = True
+            key = factories.array(key)
+            # DOES NOT WORK FOR SEQUENCE OF TENSORS OR DNDARRAYS, works for sequence of ndarrays though
+            # TODO: get outshape + outsplit; depends on wether key is bool or int and key.ndim
+        elif isinstance(key, tuple):
+            key = list(key)
+            for i, k in enumerate(key):
+                if isinstance(k, Iterable) or isinstance(key, DNDarray):
+                    advanced_indexing = True
+                    key[i] = factories.array(key[i])
```
ClaudiaComito, line 685: `key[i]` might be distributed along `self.split` (or newsplit)

heat/core/dndarray.py
```diff
@@ -653,6 +653,84 @@ def fill_diagonal(self, value: float) -> DNDarray:
...
+                raise ValueError("key can only contain 1 ellipsis")
+            elif ellipsis == 1:
+                expand_key = [slice(None)] * (arr.ndim + add_dims)
+                ellipsis_index = key.index(...)
+                expand_key[:ellipsis_index] = key[:ellipsis_index]
+                expand_key[ellipsis_index - len(key) :] = key[ellipsis_index + 1 :]
+                key = expand_key
+            if add_dims:
+                for i, k in reversed(enumerate(key)):
+                    if k is None:
```
ClaudiaComito, line 701: no `k is None` left in key if `ellipsis`

heat/core/dndarray.py
```diff
@@ -653,6 +653,84 @@ def fill_diagonal(self, value: float) -> DNDarray:
...
+            if ellipsis > 1:
+                raise ValueError("key can only contain 1 ellipsis")
+            elif ellipsis == 1:
+                expand_key = [slice(None)] * (arr.ndim + add_dims)
+                ellipsis_index = key.index(...)
+                expand_key[:ellipsis_index] = key[:ellipsis_index]
+                expand_key[ellipsis_index - len(key) :] = key[ellipsis_index + 1 :]
+                key = expand_key
+            if add_dims:
+                for i, k in reversed(enumerate(key)):
```
ClaudiaComito, line 700: `TypeError: 'enumerate' object is not reversible`...

heat/core/dndarray.py
```diff
@@ -653,6 +653,84 @@ def fill_diagonal(self, value: float) -> DNDarray:
...
+            if key is None:
+                return None
+            start, stop, step = key.start, key.stop, key.step
+            return slice(key.stop - 1, key.start - 1, -1 * key.step)
+
+        _, offsets = self.counts_displs()
+        offset = offsets[self.comm.rank]
+        range_proxy = range(self.lshape[split])
+        local_inds = range_proxy[start - offset : stop - offset]  # only works if stop - offset > 0
+        local_inds = local_inds[max(offset - start, 0) % step :: step]
```
ClaudiaComito, line 728: How about:
```
local_start = max(start-offset, 0)
local_stop = max(min(stop-offset, range_proxy.stop), 0)
local_inds = range_proxy[local_start:local_stop]
local_inds = local_inds[max(offset - start, 0) % step :: step]
```
?

heat/core/dndarray.py
```diff
@@ -653,6 +653,84 @@ def fill_diagonal(self, value: float) -> DNDarray:
...
+        if len(local_inds) and stop > offset:
```
ClaudiaComito, line 729: If I got `local_stop` and `local_start` right, we probably don't need the `stop > offset` check any more

heat/core/dndarray.py
```diff
@@ -684,65 +762,54 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
...
+        # advanced index. Later, replace the empty dimensions with the shape of the advanced index
+        proxy = self
+        names = [
+            "split" if (proxy.split is not None and i == proxy.split) else "_{}".format(i)
+            for i in range(proxy.ndim)
         ]
+        proxy_key = list(key)
+        if advanced_indexing:
+            proxy_key = list(key)
+            for i, k in reversed(enumerate(key)):
```
ClaudiaComito, line 783: reversed(enumerate) returns TypeError

ClaudiaComito (changes requested):
Hi @ben-bou thanks a lot for this! I've reviewed the new private functions, I'll get to the rest later.

heat/core/dndarray.py
ben-bou: What would `newsplit` be?

ClaudiaComito: We don't know yet because we need a sanitized key first. I guess what I'm saying is, we need to be agnostic about the `k.split` at this point. How about:

```
                    advanced_indexing = True
                    k_split = getattr(k, "split", None)
                    key[i] = factories.array(key[i], split=k_split)
```

ClaudiaComito: Oh no I'm taking this back. tensors and ndarrays have `.split`as well

heat/core/dndarray.py
```diff
@@ -653,6 +653,84 @@ def fill_diagonal(self, value: float) -> DNDarray:
...
+                expand_key = [slice(None)] * (arr.ndim + add_dims)
+                ellipsis_index = key.index(...)
+                expand_key[:ellipsis_index] = key[:ellipsis_index]
+                expand_key[ellipsis_index - len(key) :] = key[ellipsis_index + 1 :]
+                key = expand_key
+            if add_dims:
+                for i, k in reversed(enumerate(key)):
+                    if k is None:
+                        key[i] = slice(None)
+                        arr = arr.expand_dims(i - add_dims + 1)  # is the -1 correct?
```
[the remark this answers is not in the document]
ben-bou: Yes it will. The difference would be where the `split` axis is. `key[split]` would not be the index for the `split` axis if there is a `newaxis` before it. `expand_dims` is a convenient way to handle this because it's an in-place operation with correct split semantics.

heat/core/dndarray.py
ben-bou: Good point.
So what you're doing by handling the negative values by hand is actually everything indexing the `range_proxy` would do, right? So there is no reason to sidestep into using a `range` (except the comfort of using `len(local_inds)` to find out if the local slice is empty) and we can directly use 
```python
local_slice = slice(local_start + max(offset - start, 0) % step, local_stop, step)
```

heat/core/dndarray.py
ben-bou: we also don't need the `local_inds` anymore. Checking if the local slice is empty can be done with a 
`if local_stop > 0:`

heat/core/dndarray.py
ben-bou: Well, `iter(...)` is just an Iterator over the first dimension, right? So I think
"""python
def __iter__(self):
    if self.split = 0:
        warnings.warn("Iterating over the split dimension is strongly discouraged")
    for i in range(len(self)):
       yield self[i]
"""
would work as expected

heat/core/dndarray.py
ClaudiaComito: The condition for empty slice is `local_slice.start > local_slice.stop` I think.

ben-bou: One second, is checking for an empty slice needed at all?
```python
In [90]: torch.ones((3, 4, 5))[:, 5:3, :].shape                                                          
Out[90]: torch.Size([3, 0, 5])
In [91]: torch.ones((3, 4, 5))[:, 0:0, :].shape                                                          
Out[91]: torch.Size([3, 0, 5])
```
This is exactly what the `larray` of the empty rank should be.
So if that works, is knowing the `output_shape` beforehand still needed?
At least in the case where `key[i]` is a slice we can even construct the resulting `lshape_map` by calculating `local_start` and `local_stop` for every offset.

ben-bou: And if key[split] is a scalar, the split dimension is collapsed and we also know the output-shape and all the lshapes trivially.

Lastly, if advanced indexing is used, the output shape and lshapes are dependent on the key, at least if it's a dndarray.

So if I'm not mistaken, this allows us to get rid of the torch_proxy.

PR #937, opened 2022-03-23
fix #925: ht.nonzero() returns tuple of 1-D arrays instead of n-D arrays

### LINKED ISSUE #925: `ht.nonzero()` should return tuple of 1-D arrays instead of an n-D array
**Description**
`np.nonzero(a)` returns a tuple of 1-D ndarrays with length `a.ndim`. This can be then used directly and without modifications for advanced indexing across dimensions.
```
>>> a
array([[0, 1, 2],
       [3, 4, 5],
       [6, 7, 8]])
>>> mask
array([[False,  True,  True],
       [False,  True,  True],
       [False,  True,  True]])
>>> a[mask].shape
(6,)
>>> a[np.nonzero(mask)].shape
(6,)
>>> a[mask] == a[np.nonzero(mask)]
array([ True,  True,  True,  True,  True,  True])
```
`ht.nonzero` returns the indices in `torch.nonzero` format, in the example above it returns a DNDarray with shape (6,2). 

This is not what the users expect, and complicates our advanced indexing unnecessarily.

Will be addressed together with #914 

**Expected behavior**
`ht.nonzero` should return a tuple of 1-D dndarrays, for API consistency and to simplify `heat`'s advanced indexing implementation

**Version Info**
`main`

Mystic-Slice:
## Description

Switched out the torch's non-zero function with the NumPy version. Works as described in the issue ticket. Testing to be done.

Issue/s resolved: #925 

## Due Diligence

no

skip ci


mtar:

Mystic-Slice:
Hi @ClaudiaComito. This is the solution that I could come up with at first glance. It works well for the arrays I tested. But I still have to test different edge-cases like 1-D arrays, etc... Just thought I could get your input before proceeding further.
I also don't know if I handled the split, gout, and other things correctly. Let me know what you think.

Mystic-Slice:
My fix failed for 1-D arrays. I *think* I fixed it now!

Mystic-Slice:
I will update the documentation and the tests a little bit later.

heat/core/indexing.py
```diff
@@ -69,18 +70,17 @@ def nonzero(x: DNDarray) -> DNDarray:
...
+        DNDarray(
+            lcl_nonzero,
+            gshape=tuple(gout),
+            dtype=types.canonical_heat_type(lcl_nonzero.dtype),
+            split=is_split,
+            device=x.device,
+            comm=x.comm,
+            balanced=False,
+        )
     )
```
ClaudiaComito, line 84: This would return a tuple with 1 element (a 2-dimensional DNDarray). What we need is a tuple of 1-D DNDarrays, namely the columns of the original `lcl_nonzero` (output of `torch.nonzero` operation)

ClaudiaComito (changes requested):
Hey @Mystic-Slice thanks a lot for diving into this right away! 

Note that we only use torch operations internally. Our tensors might reside on GPUs, and a numpy operation would require copying the tensor to CPU.

heat/core/indexing.py
```diff
@@ -55,12 +56,12 @@ def nonzero(x: DNDarray) -> DNDarray:
 
     if x.split is None:
         # if there is no split then just return the values from torch
-        lcl_nonzero = torch.nonzero(input=x.larray, as_tuple=False)
+        lcl_nonzero = torch.tensor(np.array(np.nonzero(x.larray.numpy())))
```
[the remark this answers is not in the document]
Mystic-Slice: Ahh! I understand. Let me think of another way to do this.

heat/core/indexing.py
Mystic-Slice: No. It does work the way we want.
Let us say we have a list of lists in python (a 2-d array basically). When we do, `tuple(arr)`, we get a tuple of 1-D arrays only not a tuple of 2-D array.

Mystic-Slice: Better yet, let me show some demo output of this implementation.

Mystic-Slice: Consider this code snippet.
image

Mystic-Slice: The output is something like this.
image

Mystic-Slice:
Turns out, it is as simple as `torch.transpose()`!
I think this should work as expected.

Mystic-Slice:
What should be the split configuration of items in the returning Tuple?...should it by default be None?

heat/core/indexing.py
```diff
@@ -69,18 +69,17 @@ def nonzero(x: DNDarray) -> DNDarray:
 
     if x.ndim == 1:
         lcl_nonzero = lcl_nonzero.squeeze(dim=1)
```
ClaudiaComito, line 71: I think 70-71 are no longer needed

heat/core/indexing.py
```diff
@@ -48,19 +46,21 @@ def nonzero(x: DNDarray) -> DNDarray:
...
         # if there is no split then just return the values from torch
-        lcl_nonzero = torch.nonzero(input=x.larray, as_tuple=False)
         gout = list(lcl_nonzero.size())
         is_split = None
     else:
         # a is split
-        lcl_nonzero = torch.nonzero(input=x.larray, as_tuple=False)
         _, _, slices = x.comm.chunk(x.shape, x.split)
         lcl_nonzero[..., x.split] += slices[x.split].start
         gout = list(lcl_nonzero.size())
```
ClaudiaComito, line 66: ll. 65-58 are based on the assumption that the global `nonzero` is distributed along the rows, but now it's been transposed, so these need to be updated - or the transposition needs to happen afterwards

heat/core/tests/test_indexing.py
```diff
@@ -1,4 +1,5 @@
 import heat as ht
+from torch import transpose
```
ClaudiaComito, line 2: needed?

ClaudiaComito (changes requested):
Well done @Mystic-Slice, we're getting close, thanks a lot! The tests will be more work because many existing functions rely on the previous `nonzero` format.

heat/core/indexing.py
```diff
@@ -56,14 +56,14 @@ def nonzero(x: DNDarray) -> DNDarray:
...
         # if there is no split then just return the values from torch
-        lcl_nonzero = torch.nonzero(input=local_x, as_tuple=False)
         gout = list(lcl_nonzero.size())
         is_split = None
     else:
         # a is split
-        lcl_nonzero = torch.nonzero(input=local_x, as_tuple=False)
         # adjust local indices along split dimension
         _, displs = x.counts_displs()
         lcl_nonzero[..., x.split] += displs[x.comm.rank]
```
ClaudiaComito, line 69: ll. 68-69 are based on the assumption that the global nonzero is distributed along the rows, but now it's been transposed, so these need to be updated - or the transposition needs to happen afterwards

ClaudiaComito:
@Mystic-Slice I changed the base branch of this PR, because the change you implemented is needed in our indexing overhaul PR #938.

I've taken care of resolving conflicts (hence the commits). I hope you don't mind.

Mystic-Slice:
Not at all.
I still can make the required changes to this branch right?

heat/core/tests/test_indexing.py
Mystic-Slice: Nope. Will remove.

ClaudiaComito:
> 



> Not at all. I still can make the required changes to this branch right?

Absolutely!

Mystic-Slice:
@ClaudiaComito 
I don't know if we even need to keep track of gout (shape of the output), split, etc...
Because anyways we convert it into a tuple and we lose all this information.
What do you think?

Mystic-Slice:
@ClaudiaComito Hi, I made the fixes. But I am not able to get rid of a few errors that arise in testing. Can you please take a look?

ClaudiaComito:
> 



> @ClaudiaComito I don't know if we even need to keep track of gout (shape of the output), split, etc... Because anyways we convert it into a tuple and we lose all this information. What do you think?

@Mystic-Slice , the DNDarray metadata always have to be correct. The information doesn't get lost, it gets used in all subsequent operations, and the returned tuple is a tuple of DNDarrays whose metadata will be based on the original one.

heat/core/indexing.py
```diff
@@ -13,13 +13,13 @@
 __all__ = ["nonzero", "where"]
 
 
-def nonzero(x: DNDarray) -> DNDarray:
+def nonzero(x: DNDarray) -> Tuple:
```
ClaudiaComito, line 16: this should be `Tuple(DNDarray, ...)`

heat/core/indexing.py
```diff
@@ -73,20 +73,18 @@ def nonzero(x: DNDarray) -> DNDarray:
...
-    return DNDarray(
-        lcl_nonzero,
-        gshape=tuple(gout),
-        dtype=types.canonical_heat_type(lcl_nonzero.dtype),
-        split=is_split,
-        device=x.device,
-        comm=x.comm,
-        balanced=False,
+    lcl_nonzero = lcl_nonzero.transpose(0, 1)
+
```
ClaudiaComito, line 77: `gout` must reflect the transposed global shape

ClaudiaComito (changes requested):
@Mystic-Slice thanks for the changes. Just two more changes are needed as far as I'm concerned.

Don't worry about the tests, they fail because our getitem/setitem is transitioning (among other things) from the old `nonzero` format to the new one, so basically every operation fails. We will take care of this in the parent branch/PR.

Mystic-Slice:
> @Mystic-Slice thanks for the changes. Just two more changes are needed as far as I'm concerned.
> 
> Don't worry about the tests, they fail because our getitem/setitem is transitioning (among other things) from the old `nonzero` format to the new one, so basically every operation fails. We will take care of this in the parent branch/PR.

Ahh okok. I was so lost becoz the changes I made had nothing to do with the errors I got. Thnkx for the clarification.
I have made the changes you requested.

Mystic-Slice:
I made [this](https://github.com/helmholtz-analytics/heat/pull/937/commits/67fcdc88f5c980c27832f18ec9420337b29300bc) change because when the DNDarray is converted into a tuple, the meta-data is not really being passed on to the tuple members. This fix solves it.

heat/core/indexing.py
```diff
@@ -56,39 +56,45 @@ def nonzero(x: DNDarray) -> DNDarray:
...
         # adjust local indices along split dimension
         _, displs = x.counts_displs()
         lcl_nonzero[..., x.split] += displs[x.comm.rank]
         del displs
+
+        lcl_nonzero = lcl_nonzero.transpose(0, 1)
+
         # get global size of split dimension
         gout = list(lcl_nonzero.size())
         gout[0] = x.comm.allreduce(gout[0], MPI.SUM)
```
ClaudiaComito, line 78: I had written a lengthy detailed explanation of what's happening here, and then I lost it. Doh. Anyway.

The original `torch.nonzero` output shape is _(number_of_nonzero_elements,  number_of_dimensions)_. The process-local `lcl_nonzero` only know about how many nonzero elements are on process. But the global output DNDarray must know the total.

That `allreduce` call replaces the local value of `gout[0]` (local number of rows = local number of nonzero elements) with the sum of all `gout[0]`s on all processes, in order to synchronize `gout` to the global (albeit memory-distributed) number of rows of the output.  

Problem: `lcl_nonzero` has been transposed, so it's no longer the rows, but the columns representing the number of nonzero elements.

My suggestion: 
- transpose before the `if split` check
- adjust line 71 `lcl_nonzero[..., x.split] += displs[x.comm.rank]` to correct the row (not the column) corresponding to the split dimension
- the MPI collective call at line 78 should sum along the dimension containing the number of nonzero elements, now the columns

ClaudiaComito, line 78: The original shape of `torch.nonzero` is _(number of nonzero elements, number of x dimensions)_

What the allreduce call does is sum all number of nonzero elements across all processes so that every process knows what the total number is (in order to set the global shape of the output DNDarray).

However, right now the allreduce call doesn't do what was intended anymore... `lcl_nonzero` has been transposed.

ClaudiaComito (changes requested):
I've been losing (lengthy) comments so I'll submit this review now even if it's not quite finished and fix it later. Thanks @Mystic-Slice !

Mystic-Slice:
@ClaudiaComito Thanks for the detailed review. I have made the changes.

heat/core/indexing.py
```diff
@@ -56,39 +56,42 @@ def nonzero(x: DNDarray) -> DNDarray:
...
-        dtype=types.canonical_heat_type(lcl_nonzero.dtype),
-        split=is_split,
-        device=x.device,
-        comm=x.comm,
-        balanced=False,
+    non_zero_indices = list(
+        [
+            DNDarray(
+                dim_indices,
+                gshape=tuple(dim_indices.size()),
```
ClaudiaComito, line 82: `gshape` needs to be the global size of the array (you calculated it with the `allreduce` call), `dim_indices` are local slices of the array.

ClaudiaComito (changes requested):
Thanks @Mystic-Slice, one more change needed and then I think we're good to go.

heat/core/indexing.py
Mystic-Slice: Ahh makes sense....will make the change

ClaudiaComito (changes requested):
@Mystic-Slice can you update the changelog? I'll merge this one into the parent branch afterwards. Thanks a lot!

Mystic-Slice:
> @Mystic-Slice can you update the changelog? I'll merge this one into the parent branch afterwards. Thanks a lot!

Done :)

ClaudiaComito (approved):
Great, thanks a lot @Mystic-Slice , I will merge this into the parent branch and we can take care of the tests with the rest.

Issue #750, opened 2021-03-30
Implement a save_csv function


Markus-Goetz:
**Feature functionality**
Implement a function in io.py that allows a user to store the contents of a DNDarray in a CSV file.

rijusougata13:
can you explain what is this DNDarray ? or any link where i can found what is this about..

Markus-Goetz:
Sure, have a look in our tutorial section at [`Heat Tutorials`](https://heat.readthedocs.io/en/latest/tutorial_30_minutes.html).

Bottom line: it is a multi-dimensional array, like a NumPy `ndarray` or a torch `tensor`, with the possibility to distribute its numerical content across multiple processing nodes.

rijusougata13:
```
def save_csv(data, path, dataset, mode="w", **kwargs):
    if not isinstance(data, dndarray.DNDarray):
                raise TypeError("data must be heat tensor, not {}".format(type(data)))
            if not isinstance(path, str):
                raise TypeError("path must be str, not {}".format(type(path)))
            if not isinstance(dataset, str):
                raise TypeError("dataset must be str, not {}".format(type(path)))

    # we only support a subset of possible modes
    if mode not in __VALID_WRITE_MODES:
        raise ValueError(
            "mode was {}, not in possible modes {}".format(mode, __VALID_WRITE_MODES)
        )
    np.savetxt(path, [data], delimiter=',', fmt='%d' )
```

i tried to write this code. Can it solve this issue? @Markus-Goetz

Markus-Goetz:
It only partially solves the issue. In a multi-node processing environment the data would be corrupted due to parallel access to the same file handle as well as lost due to potential overwrites in race conditions.

You would have to respect the split access of the DNDarray and exchange the data/synchronize the program state via MPI message exchange. 

Furthermore, every feature requires corresponding test cases.

rijusougata13:
what do you mean by test cases .? should i have to give examples in comment line ? if not can you give me a short example of this test cases

Markus-Goetz:
I am referring to test cases for unit tests, i.e. small programs that can be executed to determine whether some standard inputs behave correctly.

PR #941, opened 2022-03-25
750 save csv v2

bhagemeier:
## Description

This feature provides saving of data in CSV format. In order to utilize MPI-IO and write in parallel, the CSV format written is normalized to fixed-width. This is not the same as np.savetxt(), but may serve as a starting point to move in that direction. Supports all possible splits for 2D tensors, so None, 0, and 1. `save_csv` will only ever store at most 2D data, comparable to np.savetxt().

Issue/s resolved: #750 

## Changes proposed:
- add ht.save_csv()

## Type of change

## Memory requirements

Requires a buffer (per process) to contain the row that is currently written. Except for very wide rows, it should be negligible. A 3595x3595 tensor taken from a real example worked well and showed only a small increase in the memory footprint.

## Performance

- split=None: no gain
- split=0 reduces writing time substantially
- split=1

## Due Diligence
- [x] All _supported_ (None, 0, 1) split configurations tested

yes: the generic save function can now save CSVs

skip ci


coquelin77 (dismissed):

bhagemeier:
Just found a bug with split=1, which does not work for nprocs>shape[1]. Need to fix before merging.

bhagemeier:
After thorough testing, finding and resolving 2 more bugs, I am sufficiently confident to merge this PR.

Issue #870, opened 2021-09-20
Implement missing functionalities in `ht.divide` (`out`, `where`)


ClaudiaComito:
**Feature functionality**
Enable usage of kwargs `out`, `where`

see https://numpy.org/doc/stable/reference/generated/numpy.divide.html

neosunhan:
Hi @ClaudiaComito, may I try working on this issue?

ClaudiaComito:
Hi @neosunhan , sure, go ahead

neosunhan:
Hi @ClaudiaComito, I have made a pull request that should fix this issue. Would appreciate your input on any further changes I need to make.

PR #945, opened 2022-03-29
Add `out` and `where` args for `ht.div`

neosunhan:
## Description
Implemention of `out` and `where` functionality for `ht.divide` is fairly straightforward. However, use of both arguments at the same time causes complications due to lack of support for in-place `where` in pytorch. 

Currently, I use a workaround by changing the underlying `torch.Tensor` of the `out` DNDarray. Possible alternatives include updating `indexing.where` to allow in-place modification or implementing a function that allows in-place modification using a mask (similar to [`numpy.copyto`](https://numpy.org/doc/stable/reference/generated/numpy.copyto.html)).

Issue/s resolved: #870

## Changes proposed:
- Use `out` functionality of `true_divide` in pytorch backend
- Use `indexing.where`

## Type of change

## Due Diligence

no


mtar:

neosunhan:
Found a cleaner way to handle indexing of the `where` argument.

ClaudiaComito (changes requested):
@neosunhan thank you so much for taking this on!

I have a few comments, the main thing is that `where` should probably be addressed within `_operations.__binary_op()`. This way, it would be available to all binary operations. And it would be easier to satisfy the condition that `ht.divide(t1, t2)` returns `out`, not `t1`, where `where` is False (which also holds for all numpy binary operations).

Please let me know if you need help with `__binary_ops`!

neosunhan:
@ClaudiaComito Thanks for the feedback! I have taken a look at `_operations.__binary_op` and here is what I have come up with so far.
The main idea is to create a new DNDarray with uninitialized values (using `torch.empty`) in the event that `out=None`. We can then copy over the values from the result tensor (using `torch.where` to index if a condition is provided).

heat/core/_operations.py
```diff
@@ -163,23 +168,24 @@ def __get_out_params(target, other=None, map=None):
     if out is not None:
         sanitation.sanitize_out(out, output_shape, output_split, output_device, output_comm)
         t1, t2 = sanitation.sanitize_distribution(t1, t2, target=out)
-        out.larray[:] = operation(
-            t1.larray.type(promoted_type), t2.larray.type(promoted_type), **fn_kwargs
+    else:
+        out_tensor = torch.empty(output_shape, dtype=promoted_type)
```
ClaudiaComito, line 172: 2 comments here:
- `output_shape` is the global (memory-distributed) shape of the output DNDarray, here you're initializing a potentially huge torch tensor. In this case you should call 
```
factories.empty(output_shape, dtype=..., split=..., device=...)
```
and that will take care of only initializing slices of the global array on each process

- if I understand the numpy docs correctly, this empty `out` DNDarray only needs to be initialized if `where` is not None.

ClaudiaComito (changes requested):
Great job @neosunhan ! :clap: I have left some comments in the code.

neosunhan:
@ClaudiaComito I have modified the code accordingly and added a few more tests.

ClaudiaComito:
run tests

ClaudiaComito (changes requested):
@neosunhan thanks a lot. This looks good, as far as I can tell. Can you update the CHANGELOG? When you're done, we can run the GPU tests.

neosunhan:
Updated CHANGELOG for the new `div` kwargs. I'm not sure which section is appropriate to document the new `where` kwarg for `__binary_op` (or if it should even be included in the changelog).

Also I believe there was a typo in the CHANGELOG where the "Feature Additions" heading was repeated 3 times, which I have fixed.

ClaudiaComito:
run tests

heat/core/_operations.py
```diff
@@ -43,6 +44,8 @@ def __binary_op(
         The second operand involved in the operation,
     out: DNDarray, optional
         Output buffer in which the result is placed
+    where: DNDarray, optional
+        Condition of interest, where true yield the result of the operation else yield original value in out (uninitialized when out=None)
```
ClaudiaComito, line 48: We can use numpy's docs for `where`, I think they are a bit clearer. But we must expand on them a bit, e.g. is `where` supposed/expected to be distributed, and how.

heat/core/tests/test_arithmetics.py
```diff
@@ -362,12 +362,42 @@ def test_div(self):
...
+        )
+
+        a = self.a_tensor.copy()
+        ht.div(a, self.a_scalar, out=a, where=a > 2)
+        self.assertTrue(ht.equal(a, result_where))
+        result_where_broadcasted = ht.array([[1.0, 1.0], [3.0, 2.0]])
+        a = self.a_tensor.copy()
+        ht.div(a, self.a_scalar, out=a, where=ht.array([False, True]))
+        self.assertTrue(ht.equal(a, result_where_broadcasted))
+
```
ClaudiaComito, line 392: Test case where `where` is distributed? Also the case in which `where_broadcasted` is distributed might get complicated.

ClaudiaComito (changes requested):
@neosunhan please bear with me. While we're trying to figure out why the GPU tests fail, it occurred to me that we are not checking for `where`'s distribution scheme and whether it fits the way `out` is distributed.

neosunhan:
@ClaudiaComito I added some tests where `out` and `where` are split differently and they all seem to pass. Not sure if I'm missing something here.

ClaudiaComito:
> 



> @ClaudiaComito I've added several test cases with different split configurations of `out` and `where`, but they all seem to pass. Not sure if I'm missing something here.

@neosunhan If you look at the list of checks below, you will see that the **continuous-integration/jenkins/pr-merge** has failed. Click on details, and you can see that the tests passed on 1 process, but failed on 2. Do you need help setting up your environment for multi-process? If you have installed `openmpi`, you should be able to run the tests on 2 processes with `mpirun -n 2 python -m unittest`.

neosunhan:
@ClaudiaComito Thanks for the help! I have added the check for `where`'s distribution scheme and the corresponding test cases.

ClaudiaComito:
run tests

heat/core/_operations.py
```diff
@@ -38,11 +39,17 @@ def __binary_op(
...
+        The second operand involved in the operation.
     out: DNDarray, optional
-        Output buffer in which the result is placed
+        Output buffer in which the result is placed. If not provided, a freshly allocated array is returned.
+    where: DNDarray, optional
+        Condition to broadcast over the inputs. At locations where the condition is True, the `out` array
+        will be set to the result of the operation. Elsewhere, the `out` array will retain its original
+        value. If an uninitialized `out` array is created via the default `out=None`, locations within
+        it where the condition is False will remain uninitialized. If distributed, must be distributed
+        along the same dimension as the `out` array.
```
ClaudiaComito, line 52: This is only correct if `where` and `out` have the same shape. 

For example if `out` is (100, 10000) and distributed along the columns (`out.split` is 1), `where` is (10000,), their shapes are broadcastable but `where` must be distributed along 0.

ClaudiaComito (changes requested):
I'm ready to approve this, great job @neosunhan . Just a small documentation update needed.

heat/core/_operations.py
neosunhan: >"If distributed, the split axis (after broadcasting if required) must match that of the `out` array."

Would this phrasing be accurate?

ClaudiaComito: yes, sounds accurate

ClaudiaComito:
run tests

mtar (commented):
One check can be saved

ClaudiaComito:
run tests

ClaudiaComito (approved):
thanks a lot @neosunhan !

Issue #947, opened 2022-03-30
Improve save_csv write performance


bhagemeier:
**Related**
PR #941 implementing #750 has a deficiency, consuming significantly more time needed for the string formatting. Tests showed that the rundime reduction can be up to 50%.

**Feature functionality**
Replace str.format() with '%' notation.

**Additional context**
Original str.format() profiling.

```
Wed Mar 30 15:55:57 2022    format.prof

         280709762 function calls (280698834 primitive calls) in 138.694 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   1095/1    0.003    0.000  138.696  138.696 {built-in method builtins.exec}
        1    0.018    0.018  138.696  138.696 write_bigm.py:1(<module>)
        1    0.000    0.000  136.808  136.808 /.../heat/heat/core/io.py:1110(<lambda>)
        1    0.000    0.000  136.808  136.808 /.../heat/heat/core/io.py:1060(save)
        1    0.415    0.415  136.807  136.807 /.../heat/heat/core/io.py:926(save_csv)
    28270    4.954    0.000  111.274    0.004 {method 'join' of 'str' objects}
 40020000   13.808    0.000  106.320    0.000 /.../heat/heat/core/io.py:1040(<genexpr>)
 40001495   14.825    0.000   92.514    0.000 {method 'format' of 'str' objects}
 40000000   27.901    0.000   77.688    0.000 /.../venves/heat/lib/python3.9/site-packages/torch/_tensor.py:556(__format__)
    20000    0.075    0.000   22.353    0.001 /.../venves/heat/lib/python3.9/site-packages/torch/_tensor.py:597(__iter__)
    20000   22.216    0.001   22.216    0.001 {method 'unbind' of 'torch._C._TensorBase' objects}
 40000004   20.571    0.000   20.571    0.000 {method 'item' of 'torch._C._TensorBase' objects}
 40000000   19.572    0.000   19.572    0.000 {method '__format__' of 'float' objects}
 40020001    6.394    0.000    6.394    0.000 {method 'dim' of 'torch._C._TensorBase' objects}
 40000001    3.256    0.000    3.256    0.000 {built-in method torch._C._has_torch_function_unary}
    20000    2.530    0.000    2.530    0.000 {method 'Write_at' of 'mpi4py.MPI.File' objects}
```

Using the % operator:

```
Wed Mar 30 15:58:41 2022    percent.prof

[...]

PR #948, opened 2022-03-31
Improve save_csv string formatting

bhagemeier:
Using .format can take up to 2x as long as using %.

Also add a test covering an additional line of code.

## Description

The title says it all. Improve write performance.

Issue/s resolved: #947 

## Changes proposed:
- Use string interpolation operator rather than str.format()

## Type of change
- performance improvement

## Memory requirements
- no changes

## Performance
Up to 2x as fast in overall runtime than using str.format(), e.g. cutting down from 139s to 66s for the same data. See performance comparison on #947 .

## Due Diligence

no


heat/core/io.py
```diff
@@ -1004,16 +1004,16 @@ def save_csv(
         decimals = 0
         dec_sep = 0
         if sign == 1:
-            fmt = "{: %dd}" % (pre_point_digits + 1)
```
bhagemeier, line 1007: Format strings are different depending on whether you use str.format() or string interpolation. String interpolation is significantly faster than str.format(). Therefore, this PR switches to string interpolation.

heat/core/io.py
```diff
@@ -1033,11 +1033,11 @@ def save_csv(
     for i in range(data.lshape[0]):
         # if lshape is of the form (x,), then there will only be a single element per row
         if len(data.lshape) == 1:
-            row = fmt.format(data.larray[i])
+            row = fmt % (data.larray[i])
```
bhagemeier, line 1036: The format strings have to be used differently now.

heat/core/tests/test_io.py
```diff
@@ -195,6 +195,21 @@ def test_save_csv(self):
                             if data.comm.rank == 0:
                                 os.unlink(filename)
 
+        data = ht.random.randint(0, 100, size=(150,))
```
bhagemeier, line 198: This captures a special case that had not been covered before.

heat/core/io.py
mtar: Have you tried f-strings? How do they compete?

mtar: Have you tried f-strings? How do they compete?

bhagemeier: Yes, tried them. Seems a bit convoluted to create them. f-strings are intended to be in code as literals, whereas I dynamically generate the format string depending on desired precision etc. You can give it a try if you like. For me personally, it wasn't quite worth the effort for an output format that is far from ideal in terms of performance anyway.

ClaudiaComito:
run tests

ClaudiaComito (approved):
Thanks a lot @bhagemeier !

PR #951, opened 2022-04-02
IO: Replaced factories.arraay() with DNDarray

shahpratham:
## Description

Replaced factories.array with DNDarray in `io.py`
Issue/s resolved: https://github.com/helmholtz-analytics/heat/issues/797

## Changes proposed:
- factories.array -> dndarray.DNDarray

## Type of change
  
enhancement 

## Due Diligence

no


mtar:

shahpratham:
@ClaudiaComito @mtar Please tell if any changes are required.

heat/core/io.py
```diff
@@ -895,8 +895,14 @@ def load_csv(
         # In case there are some empty lines in the csv file
         local_tensor = local_tensor[:actual_length]
 
-        resulting_tensor = factories.array(
-            local_tensor, dtype=dtype, is_split=0, device=device, comm=comm
+        resulting_tensor = DNDarray(
+            local_tensor,
+            gshape=tuple(local_tensor.shape),
```
ClaudiaComito, line 900: `gshape` is supposed to be the global shape of the memory-distributed array `resulting_tensor`. Here you're setting it to the shape of `local_tensor`, which is basically a slice of the global array.

If we have all the information we need to calculate `gshape` without communication among processes, then we can call `DNDarray(...)`, otherwise we need to use `factories.array()` and that will take care of the comm

heat/core/io.py
```diff
@@ -895,8 +895,14 @@ def load_csv(
...
+            dtype=dtype,
+            split=0,
+            device=device,
+            comm=comm,
+            balanced=local_tensor.is_balanced,
```
ClaudiaComito, line 905: `local_tensor` is a `torch.Tensor`. It doesn't "know" about being a slice of a larger distributed array.

`is_balanced()` is a method of the `DNDarray` class, it has to do with whether the memory-distributed [DNDarray](https://heat.readthedocs.io/en/latest/autoapi/heat/core/dndarray/index.html) is distributed evenly among the available processes.

If we cannot asses load balance of the output DNDarray (I'm not sure that's the case here, I haven't checked), we set `balanced =  None`.

heat/core/io.py
```diff
@@ -918,7 +924,15 @@ def load_csv(
                 values = line.replace("\n", "").replace("\r", "").split(sep)
                 values = [float(val) for val in values]
                 data.append(values[displs[rank] : displs[rank] + chunk[rank]])
-        resulting_tensor = factories.array(data, dtype=dtype, is_split=1, device=device, comm=comm)
+        resulting_tensor = DNDarray(
+            data,
+            gshape=tuple(data.shape),
```
ClaudiaComito, line 929: see above

heat/core/io.py
```diff
@@ -918,7 +924,15 @@ def load_csv(
...
+            dtype=dtype,
+            split=1,
+            device=device,
+            comm=comm,
+            balanced=data.is_balanced,
```
ClaudiaComito, line 934: see above

ClaudiaComito (changes requested):
Hey @shahpratham thanks a lot for jumping in and well done,  the use cases are correct, we want to save us some communication time that arises with `factories.array()` when `is_split` is not None.

However `gshape` is the global shape of the DNDarray, and you cannot derive `balanced` from the process-local torch tensor. More in the comments below.

heat/core/io.py
shahpratham: Okay, understood. So can I make a copy of that tensor before slicing it and pass `gshape=tuple(local_tensor_copy.shape)` to it?

shahpratham:
Hey @ClaudiaComito, I have made some changes, kindly review.

heat/core/io.py
```diff
@@ -895,8 +895,14 @@ def load_csv(
         # In case there are some empty lines in the csv file
         local_tensor = local_tensor[:actual_length]
 
-        resulting_tensor = factories.array(
-            local_tensor, dtype=dtype, is_split=0, device=device, comm=comm
+        resulting_tensor = DNDarray(
+            local_tensor,
+            gshape=local_shape,
```
ClaudiaComito, line 900: Hi @shahpratham `gshape` is the global shape, not the local shape. 
This function reads an array  that might be, say, _(1billion x 1000)_ in size (just making the size up). The data will be distributed on many processes, i.e. each process will read only a specific subset of lines (if split=0) or columns (if split=1) out of that file, and store them in `local_tensor`. 

So `local_tensor` is the process-local slice of data. Its shape, `local_shape` may vary  (depending on number of processes for example. See `communication.chunk()` ). But the global shape `gshape` will always be _(1billion x 1000)_.

shahpratham: > But the global shape gshape will always be (1billion x 1000).

Yes, so I need to get global shape of the csv file, after reading it, like how its done in `load_hdf5` function(on line 121)  and in `load_netcdf` function (on line 334), right?

ClaudiaComito:
@shahpratham now you know everything about Heat - let's merge this PR, can you update? Thanks!

shahpratham:
yes, sorry I forgot about this. 
I have created a  new PR #1089 , due to some issue with my previous branch.

ClaudiaComito:
This is addressed in #1089, closing

PR #952, opened 2022-04-03
fix #351: Solution 2: Allow weights to be 2D array for tuple axis

### LINKED ISSUE #351: ht.average() should allow weighted average across multiple axes
**Feature functionality**

`ht.average(x, weights=None, axis=None,  returned=False)` (PR #352) has the same constraints as numpy.average(), among them the following:

- if axis is not None, the weights tensor must be 1D and of the same size as x.gshape[axis].

This makes sense as long as axis is not a tuple. numpy.average() supports tuple axis, but as far as I understand, in this case np.average() only works if weights.shape = x.shape (this is not clear from the numpy documentation, it's what I gather from testing).

I see two possible solutions: 
1) enforce weights.shape = x.shape if axis is tuple (and make it clear in the docs)
2) allow weights = tuple of ht.tensors of the appropriate size and in the appropriate order for the respective axis.

At the moment, ht.average() raises a NotImplementedError if axis is a tuple.

Mystic-Slice:
## Description

When the shapes of `x` and `weights` are the same, everything proceeds as it did before. 
When they are different, there are 2 cases possible:
* Case 1: `axis` is of type `int` and `weights` is a 1D array.
* Case 2: `axis` is of type `tuple` and `weights` is a 2D array of weights for each of the dimensions specified in `axis`

Both cases are handled in the same way (by converting `axis` and `weights` in Case 1 to be `int` and 2D array respectively). The average is calculated along one dimension at a time in the order specified in `axis`.

This solution works as expected for the few cases I tested. Different splits are not tested yet.

If the `axis` is a tuple of ints, the `weights` array could be of the same shape as `x` or it could be a 2D array containing weights for each of the dimensions specified in `axis` in the correct order.

**TODO:** Return the `cumwgt` for tuple axis case. 
Should the sum of all elements in the weights array be returned as `cumwgt`?

Issue/s resolved: #351 

## Type of change

New feature

## Due Diligence

yes / no

skip ci


mtar:

PR #953, opened 2022-04-03
fix #351: Solution 1: enforce weights.shape = x.shape for tuple axis

### LINKED ISSUE #351: ht.average() should allow weighted average across multiple axes

Mystic-Slice:
Issue/s resolved: #351

## Type of change

Documentation update

## Due Diligence

yes / no

skip ci


mtar:

ClaudiaComito (changes requested):
@Mystic-Slice thanks a lot for taking this on and  welcome to our legacy code 😬 

I vaguely remember implementing `ht.average`, it seems so overly complicated now.

I agree with your solution 1, we can enforce - just like numpy - that `weights.shape == a.shape` if averaging along more than one axis. They must be distributed along the same split axis as well.

The distributed operation itself can be simplified quite a bit, as it's just:

`(a*weights).sum(axis)/weights.sum(axis)`

`a*weights` calls `_operations.__binary_op()` and `ht.sum()` calls `_operations.__reduce_op()`, those will take care of sanitizing, communication when needed, etc.

Please go ahead and make the changes whenever you feel comfortable. Thanks a lot!

Mystic-Slice:
Hey @ClaudiaComito!
Sorry for the late reply!
I think the changes that you requested are already present in the code.
See lines 286 - 299

Mystic-Slice:
Changes moved to a new branch. See, https://github.com/helmholtz-analytics/heat/pull/1037

PR #961, opened 2022-04-09
Fixes #814

### LINKED ISSUE #814: numpy.loadtxt vs heat.load_csv
**Description**
load_csv is not from numpy API but behaves differently than naive use of numpy.loadtxt.

**To Reproduce**
```
import numpy as np
import heat as ht

a = np.zeros(8)
np.savetxt("a.csv", a, delimiter=',')

a1 = np.loadtxt("a.csv", delimiter=',')  # fix a: ndmin=2
a2 = ht.load_csv("a.csv")                # fix b: .squeeze()

assert a1.shape == a2.shape, f"Expected shape {a1.shape} but got shape {a2.shape}"
print(f"ok: {a1.shape}")
```

**Expected behavior**
Both shapes identical.

**Illustrative**
If applicable, add screenshots or minimal examples to help explain your problem.

**Version Info**
latest at the time

**Additional comments**
Squeezing the HeAT array after reading or adding `ndmin=2` to `loadtxt` makes it the same (see "ndmin" arg here https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html)

@coquelin77 Not sure if this is really an issue, just an observation.

marianna13:
## Description

Issue/s resolved: #814

## Changes proposed:
- Apply NumPy `squeeze` function to data that we read in `load_csv` function. It makes sure we don't get unnecessary dimensions to an array. E,g. array with shape (2,) will be loaded as an array with shape (2,1). This minor fix will make sure that shapes of arrays loaded with NumPy's `loadtxt` function will have equal shapes as arrays loaded with `load_csv`.

## Type of change

## Due Diligence
- [ x] All split configurations tested
- [ x] Multiple dtypes tested in relevant functions

no


mtar:

ClaudiaComito (changes requested):
@marianna13 thanks for contributing and sorry for getting back to you so late.

The arrays we're "squeezing" are actually pytorch tensors, so we should apply `torch.squeeze`. Afterwards we can look into the tests.

ClaudiaComito:
closing this as no distributed test cases

PR #982, opened 2022-06-23
Chunk function for CSR sparse arrays

Mystic-Slice:


mtar:

ClaudiaComito (changes requested):
Thanks a lot @Mystic-Slice for taking this on. 

Before we can apply `chunk`  or `sparse_csr_array`, we have to have a functioning csr_matrix class. I've added a list of missing attributes, let's start building on those.

Thanks a lot!

PR #983, opened 2022-07-01
Signal processing: fully distributed 1D convolution

shahpratham:
## Description
Heat currently supports 1D convolution with only distributed signals. This PR add support for performing 1D convolution with fully distributed signal as well as the kernel.

### Changes done:
- Support added for 1D convolution with all existing modes.
- Pads uneven local weights with zero when input kernel is unevenly distributed.
- Used `bcast` to send kernel/ weights among the processes.
- Applying convolution on local signal and weights.
- Slicing and adding up the calculated signal_filtered 
- Modified and added documentation and correctness tests.

### Approach of solving:
Whenever the kernel `v`, is distributed, we must find the convolution of all distributed kernels with signals and store it in a DNDarray. Later, compute the filtered signal by slicing a chunk of the size of our output's (convolution) shape from each row of the DNDarray and adding them along its column to get our result.

Added the diagram of the same for better understanding [here](https://jamboard.google.com/d/1e1n-Py7drEk1JwmwH7_WKQaeL-gV6DDfz9G6UluJ8cE/edit?usp=sharing)

skip ci


mtar:

heat/core/signal.py
```diff
@@ -45,9 +45,11 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
         Contrary to the original `numpy.convolve`, this function does not
         swap the input arrays if the second one is larger than the first one.
         This is because `a`, the signal, might be memory-distributed,
-        whereas the filter `v` is assumed to be non-distributed,
-        i.e. a copy of `v` will reside on each process.
+        whereas the filter `v` may or may not be distributed,
+        i.e. a copy of `v` may reside on each process.
```
ClaudiaComito, line 49: Now we can deal with distributed kernels, so we can improve consistency with numpy by swapping `a` and `v` if `a` is smaller than `v`, correct?

heat/core/signal.py
```diff
@@ -105,20 +115,32 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
...
         # apply halos to local array
         signal = a.array_with_halos
     else:
         signal = a.larray
 
+    # flip filter for convolution as Pytorch conv1d computes correlations
+    v = flip(v, [0])
+    if v.larray.shape != v.lshape_map[0]:
+        # pads weights if input kernel is uneven
+        target = torch.zeros(v.lshape_map[0][0], dtype=v.larray.dtype)
```
ClaudiaComito, line 133: specify `device`

ClaudiaComito (changes requested):
Thanks a lot @shahpratham , great job! 

I have some minor comments but also am thinking of ways the two communication loops could be merged into one.

heat/core/signal.py
shahpratham: yes, will make the required changes.

heat/core/signal.py
```diff
@@ -105,20 +115,32 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
     a = pad(a, pad_size, "constant", 0)
 
     if a.is_distributed():
-        if (v.shape[0] > a.lshape_map[:, 0]).any():
-            raise ValueError("Filter weight is larger than the local chunks of signal")
+        if (v.lshape_map[:, 0] > a.lshape_map[:, 0]).any():
+            raise ValueError(
+                "Local chunk of filter weight is larger than the local chunks of signal"
+            )
```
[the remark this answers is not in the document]
shahpratham: yes, did it above.  
But when,
```python
a = ht.array([1, 1, 1, 1], split=0)
v = ht.array([1, 1, 1])
```
we should raise this error, because it shouldn't work when no. of processes is more than 2.

shahpratham:
@ClaudiaComito have made all the suggested changes.

heat/core/signal.py
```diff
@@ -77,24 +84,23 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
...
-        raise ValueError("Filter size must not be greater than or equal to signal size")
     if mode == "same" and v.shape[0] % 2 == 0:
         raise ValueError("Mode 'same' cannot be used with even-sized kernel")
 
+    if v.shape[0] > a.shape[0]:
+        a, v = v, a
+
     # compute halo size
-    halo_size = v.shape[0] // 2
+    halo_size = int(v.lshape_map[0][0] / 2)
```
krajsek, line 96: This does not work. The halo size cannot depend on the local size of the first process, as this is arbitrary, e.g. 
one can set the local filter size at process 1 to zero with redistribute_, resulting in halo_size=0, regardless of the actual filter size.

krajsek (changes requested):
1.)Please number the branch

2.)To make it  more compatible with numpy, scalar values should also be allowed, e.g. np.convolve(1,2)

heat/core/signal.py
shahpratham: Basically, what I want is to have `halo_size` be the maximum local size of the `v`. I initially assumed `v` to be balanced. 

> local filter size at process 1 to zero with redistribute_

So, yes this won't work,I'll change `halo_size`  to be the maximum of all local chunk size.

My method works with balanced tensor on process, so if the distribution is uneven, I pad the local chunk with 0 till halo_size is reached. 

Will change this :+1:

shahpratham:
> 1.)Please number the branch

Won't doing close this PR?

image

> 2.)To make it more compatible with numpy, scalar values should also be allowed, e.g. np.convolve(1,2)

Yes, done :+1:

heat/core/signal.py
krajsek: If the maximum local size will work you can do: 
halo_size = torch.max(v.lshape_map[:, 0]).item() // 2

krajsek:
> > 1.)Please number the branch
> 
> Won't doing close this PR?

yes, it seems so. Now it is not so important, but in the future the numbering helps to connect branches and PRs.

> 
> image
> 
> > 2.)To make it more compatible with numpy, scalar values should also be allowed, e.g. np.convolve(1,2)
> 
> Yes, done 👍

heat/core/signal.py
shahpratham: yes, done.

heat/core/signal.py
```diff
@@ -127,22 +149,54 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
         signal = signal.to(float_type)
         weight = weight.to(float_type)
```
ClaudiaComito, line 151: Looks like `v` needs to be cast to float on GPU as well, GPU tests fail later at line 158
 
Screenshot 2022-11-24 at 09 43 43

shahpratham: done, does it works now?

heat/core/signal.py
```diff
@@ -77,24 +88,23 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
...
-        raise ValueError("Filter size must not be greater than or equal to signal size")
     if mode == "same" and v.shape[0] % 2 == 0:
         raise ValueError("Mode 'same' cannot be used with even-sized kernel")
 
+    if v.shape[0] > a.shape[0]:
+        a, v = v, a
+
     # compute halo size
-    halo_size = v.shape[0] // 2
+    halo_size = torch.max(v.lshape_map[:, 0]).item() // 2
```
krajsek, line 100: This seems to only partially solve the problem. It works for some configuration but for others it gives wrong results, for example on two processes:
import heat as ht 
import torch 
signal = ht.ones((7,), split=0)
weight = ht.ones((5,), split=0)
target_map = torch.zeros((weight.comm.size, 1), dtype=torch.int)
target_map[0, 0] = 4
target_map[1, 0] = 1
weight.redistribute_(target_map=target_map)
print(ht.convolve(signal, weight))

leads to:
DNDarray([1., 2., 3., 4., 5., 5., 5., 5., 5., 4., 3.], dtype=ht.float32, device=cpu:0, split=0)

whereas the correct result must be:
DNDarray([1., 2., 3., 4., 5., 5., 5., 4., 3., 2., 1.], dtype=ht.float32, device=cpu:0, split=0)

krajsek (changes requested):

heat/core/signal.py
shahpratham: yes, this convolve only supports weights which are balanced on all nodes, otherwise slicing method doesn't work due to more entries in `global_signal_filtered` as halo_size increased

krajsek: I see, however these cases should be caught in any case, either by an error message or by rebalancing as getting wrong results is dangerous.

shahpratham: yes, even I think error message should be given. will do this asap.

krajsek: o.k., the code seems to be in a good shape now, from my side I would approve a PR. Nice work!

ClaudiaComito: Thanks @shahpratham , @krajsek , I'll approve and merge. Well done @shahpratham ! 👏🏼

ClaudiaComito (approved):

ClaudiaComito (approved):

PR #988, opened 2022-07-11
Features/384 sparse matrices

AmirTouilPU:
### New class: coo_matrix:
coo_matrix class (should be re-named to coo_array). Indices are calculated using a function in the class file 

Issue/s resolved: 
- coo_indices are now working
- gshape calculation from a torch.sparse_coo_tensor in the test.py file: turnt the shape to tuple (couldn't modify the tuple), then to a list, changed the gshape[split] value, then turnt it to a tuple again. 

Should the function indices be part of the class?

### DIFF
--- a/heat/core/communication.py
+++ b/heat/core/communication.py
@@ -208,6 +208,61 @@ class MPICommunication(Communication):
             tuple(slice(0, shape[i]) if i != split else slice(start, end) for i in range(dims)),
         )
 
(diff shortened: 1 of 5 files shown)


ClaudiaComito:
Following

ClaudiaComito:
closing this as it is unfinished. Happy to reopen if somebody else wants to work on it.

Issue #990, opened 2022-07-15
Add `copy` kwarg to `asarray`


neosunhan:
**Feature functionality**
Enables the `copy` kwarg to `asarray`.

**Additional context**
In compliance with the Python array API standard

ClaudiaComito:
closed by #1119

PR #991, opened 2022-07-15
Add `copy` kwarg to `asarray`

neosunhan:
## Description
Enables `copy` kwarg in `asarray`. The keyword defaults to `False` for backward compatibility.

Issue/s resolved: #990

## Type of change

## Due Diligence

no

skip ci


heat/core/factories.py
```diff
@@ -434,6 +434,7 @@ def array(
 def asarray(
     obj: Iterable,
     dtype: Optional[Type[datatype]] = None,
+    copy: bool = False,
```
ClaudiaComito, line 437: According to the array-API standard, `asarray(copy=False)` should return `ValueError` if a copy turns out to be necessary. 
https://data-apis.org/array-api/latest/API_specification/generated/signatures.creation_functions.asarray.html

I propose to set this default to `None` (we won't copy unless necessary), which is basically the default numpy users expect. Then if the users specify `asarray(copy=False)` and copying is necessary, we can return a `ValueError`.

ClaudiaComito (changes requested):
Thanks a lot @neosunhan ! I only have a comment about the array-API behaviour when `copy=False`, let me know what you think.

ClaudiaComito:
@JuanPedroGHM have we got a fitting label for these changes re: auto-generated changelog? I was going to suggest `array-API`. Also, I'm thinking we need some kind of roadblock for when we're about to merge a PR that hasn't been labelled properly.

heat/core/factories.py
neosunhan: I've changed the default and added a check in the event `copy=False`.

heat/core/factories.py
```diff
@@ -482,7 +487,15 @@ def asarray(
...
     """
-    return array(obj, dtype=dtype, copy=False, order=order, is_split=is_split, device=device)
+    if isinstance(copy, bool) and not copy:
+        if not (
+            isinstance(obj, DNDarray)
+            and (dtype is None or dtype == obj.dtype)
+            and (is_split is None or is_split == obj.split)
+            and (device is None or device == obj.device)
+        ):
+            raise ValueError("copy is necessary")
```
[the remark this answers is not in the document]
mtar: I think the check fits better in the array function where the real logic is.

mtar (changes requested):
Now that we have a CI running again, we can finally review some PRs. :smiley:

ClaudiaComito:
Switched base branch from `main` to `release/1.2.x`  so we can release this sooner.

ClaudiaComito:
Closing this as it is addressed in #1119

Issue #992, opened 2022-07-15
[Bug]: Precision loss when using `float64` data type


neosunhan:
### What happened?

Loss of precision is observed in several Heat functions when using the `float64` data type. This is due to pre-conversion to `float32` (the pytorch default floating data type) before converting the resulting tensor back to `float64`.

The following functions are affected:
- `arange`
- `array`
- `linspace`
- `abs`

### Code snippet

```python3
>>> import heat as ht
>>> ht.arange(16777217.0, 16777218, 1, dtype=ht.float64)
DNDarray([16777216.], dtype=ht.float64, device=cpu:0, split=None)
>>> ht.array(16777217.0, dtype=ht.float64)
DNDarray(16777216., dtype=ht.float64, device=cpu:0, split=None)
```

### Version

main (development branch)

### Python version

3.10

### PyTorch version

1.11

PR #993, opened 2022-07-15
Bug/992 precision loss

neosunhan:
## Description
Fixed precision loss in several functions when using `float64` data type.

Issue/s resolved: #992

## Changes proposed:
- Passed in the `dtype` argument to backend torch functions instead of converting to desired data type after using default torch data type.

## Type of change

## Due Diligence

no

skip ci

### DIFF
--- a/heat/core/factories.py
+++ b/heat/core/factories.py
@@ -139,10 +139,12 @@ def arange(
     # compose the local tensor
     start += offset * step
     stop = start + lshape[0] * step
-    data = torch.arange(start, stop, step, device=device.torch_device)
-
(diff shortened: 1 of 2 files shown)


heat/core/linalg/basics.py
```diff
@@ -1179,6 +1179,8 @@ def matrix_norm(
 
     row_axis, col_axis = axis
 
+    dtype = types.promote_types(x.dtype, types.float32)
```
ClaudiaComito, line 1182:
```suggestion
    # dtype = types.promote_types(x.dtype, types.float32)
```

Issue #1004, opened 2022-08-15
[Bug]: Extra dimension added on 0-dimensional arrays


neosunhan:
### What happened?

Several functions return a 1-dimensional array (instead of a 0-dimensional array) when the output is of size 1. Their corresponding numpy and pytorch equivalents return 0-dimensional arrays.

Below is a non-exhaustive list of such functions:
- `all`
- `any`
- `prod`
- `sum`
- `argmax`
- `argmin`
- `max`
- `mean`
- `min`
- `std`
- `var`

### Code snippet triggering the error

```shell
>>> ht.all(ht.array([True])).shape
(1,)
>>> np.all(np.array([True])).shape
()
>>> ht.sum(ht.arange(5)).shape
(1,)
>>> np.sum(np.arange(5)).shape 
()
```

### Version

main (development branch)

PR #1005, opened 2022-08-15
Reduction operations on entire array now return 0-dimensional `DNDarray`

neosunhan:
## Description
Return 0-dimensional (instead of 1-dimensional) singleton arrays for several functions.

Issue/s resolved: #1004

## Changes proposed:
- Remove flattening step after partial reduction operation for 0-dimensional arrays
- EDIT 21.8.2023: adapt tests

## Type of change

## Due Diligence

yes
- `all`
- `any`
- `allclose`
- `prod`
- `sum`
- `argmax`
- `argmin`
- `max`
- `mean`
- `min`
- `std`
- `var`


mrfh92 (approved):
As far as I can judge, the changes look fine to me and also the tests have been adapted accordingly. Therefore, I recommend merging if the CI runs through. 

Thanks for the work :+1:

Issue #1010, opened 2022-08-20
[Bug]: `test_random` fails on GPU 


ClaudiaComito:
### What happened?

`test_random.test_permutation` (and possibly other functions) fail on GPU, `ht.permutation` does not return the same result as the torch version.

Probably relevant:

https://pytorch.org/docs/1.11/notes/randomness.html#controlling-sources-of-randomness

### Code snippet triggering the error

```shell
heat.core.tests.test_random.TestRandom
```

### Error message or erroneous outcome

```shell
======================================================================
FAIL: test_permutation (heat.core.tests.test_random.TestRandom)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/p/project/haf/users/comito1/devel/heat/heat/core/tests/test_random.py", line 72, in test_permutation
    self.assertTrue((a.larray == a_cmp).all())
AssertionError: tensor(False, device='cuda:0') is not true

----------------------------------------------------------------------
```

### Version

1.2.x

### Python version

3.9

### PyTorch version

1.11

PR #1014, opened 2022-08-22
fix: set cuda rng state on gpu tests for test_random.py

JuanPedroGHM:
## Description

On test_random.py (test_permutation and test_randperm), change between torch.cuda and torch.random to get/set_rng_state depending on self.device.torch_device.
 
Issue/s resolved: #1010 

## Type of change

#### Does this change modify the behavior of other functions? If so, which?
no

Tested with:
python 3.8
torch 1.12
cuda 11.6
openmpi 4.0

skip ci


ClaudiaComito (dismissed):
Verified that `test_random` runs on HDFML with:
- OpenMPI/4.1.2
- Python/3.9.6
- mpi4py/3.1.3 
- PyTorch/1.11-CUDA-11.5
- on 1 Node, 1 process & 2 processes
- on 2 Nodes, 1 process each

It works! Thanks a lot @JuanPedroGHM !

ClaudiaComito (approved):
Added python 3.9 and pytorch 1.12 tests to the CI matrix

Issue #1017, opened 2022-08-29
[Bug]: `prod` and `sum` does not work for empty arrays


neosunhan:
### What happened?

Calling `prod` and `sum` on an empty DNDarray throws an error.

### Code snippet triggering the error

```python3
>>> import numpy as np
>>> import heat as ht
>>> ht.sum(ht.array([]))
Traceback (most recent call last):
...
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
>>> np.sum(np.array([]))
0.0
>>> ht.prod(ht.array([]))
Traceback (most recent call last):
...
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
>>> np.prod(np.array([]))
1.0
```

### Version

main (development branch)

### Python version

3.10

### PyTorch version

1.11

PR #1018, opened 2022-08-29
Bug/1017 `prod` / `sum` with empty arrays

neosunhan:
## Description
Fixed a bug where calling `prod` and `sum` on empty DNDarrays resulted in an error.

Issue/s resolved: #1017

## Changes proposed:
- Check for `split` in `__reduce_op` before using split axis to index

## Type of change

## Due Diligence

no

skip ci


heat/core/_operations.py
```diff
@@ -422,7 +422,7 @@ def __reduce_op(
     balanced = x.balanced
 
     # if local tensor is empty, replace it with the identity element
-    if 0 in x.lshape and (axis is None or (x.split in axis)):
+    if split is not None and 0 in x.lshape and (axis is None or split in axis):
```
mtar, line 425:
```suggestion
    if x.is_distributed() and 0 in x.lshape and (axis is None or split in axis):
```
A small suggestion: x.is_distributed() also checks the split

heat/core/tests/test_arithmetics.py
```diff
@@ -646,6 +646,10 @@ def test_prod(self):
         self.assertEqual(shape_split_axis_tuple_prod.split, None)
         self.assertTrue((shape_split_axis_tuple_prod == expected_result).all())
 
+        # empty array
+        empty = ht.array([])
+        self.assertEqual(ht.prod(empty), ht.array([1.0]))
```
ClaudiaComito, line 651: the introduced change affects distributed DNDarrays, but tests are on non-distributed arrays.

heat/core/tests/test_arithmetics.py
```diff
@@ -792,6 +796,10 @@ def test_sum(self):
         self.assertEqual(shape_split_axis_tuple_sum.split, None)
         self.assertTrue((shape_split_axis_tuple_sum == expected_result).all())
 
+        # empty array
+        empty = ht.array([])
+        self.assertEqual(ht.sum(empty), ht.array([0.0]))
```
ClaudiaComito, line 801: see above

ClaudiaComito (changes requested):
Thanks a lot for catching this @neosunhan ! My only comment is  on the test cases. Otherwise this PR is ready as far as I'm concerned. Well done!

heat/core/tests/test_arithmetics.py
neosunhan: Hi Claudia, from what I understand, the change actually affects non-distributed (to be more specific, empty) DNDarrays. In the linked issue #1017, calling `prod` or `sum` on empty arrays would cause them to (erroneously) enter this if clause. This causes the error in `split + 1` since `split == None`. The introduction of the check to verify that the array is distributed fixes this by skipping the if clause for non-distributed arrays.

ClaudiaComito: Got it, thanks!

ClaudiaComito (approved):
Perfect, thanks @neosunhan !

Issue #778, opened 2021-05-26
Implement `broadcast_arrays`, `broadcast_to`


ClaudiaComito:
**Related**
#774

**Feature functionality**
see https://numpy.org/doc/stable/reference/generated/numpy.broadcast_arrays.html
see https://numpy.org/doc/stable/reference/generated/numpy.broadcast_to.html
see #689, #701 

**Additional context**
See [Python array API standard](https://data-apis.org/array-api/latest/index.html)

PR #1020, opened 2022-08-30
Implement `broadcast_arrays`, `broadcast_to`

neosunhan:
## Description
Implement `broadcast_arrays`, `broadcast_to`

Issue/s resolved: #778

## Changes proposed:
- 

## Type of change

## Due Diligence

no

skip ci


heat/core/manipulations.py
```diff
@@ -89,6 +91,54 @@ def balance(array: DNDarray, copy=False) -> DNDarray:
...
+        DNDarray to broadcast.
+    shape : Tuple[int, ...]
+        Array shape. Must be compatible with x.
+    """
+    if not isinstance(x, DNDarray):
+        raise TypeError("'x' must be a DNDarray, currently {}".format(type(x)))
+
+    broadcasted = torch.broadcast_to(x.larray, shape)
+
+    return factories.array(broadcasted, dtype=x.dtype, device=x.device)
```
ClaudiaComito, line 139:
```suggestion
    # figure out the output split axis via dndarray.__torch_proxy__ and named tensors functionality
    try:
        torch_proxy = x.__torch_proxy__
    except AttributeError:
        raise TypeError("'x' must be a DNDarray, currently {}".format(type(x)))

    split_tags = [None]*x.ndim
    if x.split is not None:
        split_tags[x.split] = "split"
    torch_proxy = torch.tensor(torch_proxy, names=split_tags)
    torch_proxy = torch_proxy.broadcast_to(shape)
    output_split = torch_proxy.names.index("split")

    # exploit binary operations broadcasting 
    broadcasted = factories.zeros(shape, dtype=x.dtype, split=output_split, device = x.device, comm=x.comm)
    broadcasted += x

    return broadcasted
```

ClaudiaComito (changes requested):
Thanks a lot @neosunhan and apologies again for the Draft PR mix-up.

These two functions don't return the expected values when the input arrays are distributed. I've suggested a change for `broadcast_to` that leverages all the broadcasting checks in `_operations.__bin_op()`. 

You should be able to reimplement `broadcast_arrays` using the new version of `broadcast_to` so that distributed arrays can be broadcasted even if they are distributed along different axes, e.g. this should be possible:
```python
a = ht.arange(100, split=0)
b = ht.arange(200).reshape(2,100)
b.resplit_(axis=1)
# a.split is 0, b.split is 1
ht.broadcast_arrays(a,b)
```

Thanks a lot!

mrfh92 (changes requested):
**Review for PR #1020 "Implement `broadcast_arrays`, `broadcast_to`"**

This PR adds two broadcasting-routines known from NumPy to Heat. 

As far as I can judge, the code is correct and sufficiently commented. However, I would like to suggest to add some more information into the documentation, either by an example what each of the two routines does for a concrete array or by providing reference to the corresponding NumPy-functionality. 

Once these changes have been done, I recommend merging.

heat/core/manipulations.py
```diff
@@ -89,6 +91,81 @@ def balance(array: DNDarray, copy=False) -> DNDarray:
...
+        raise ValueError(e)
+
+    # extract torch tensors
+    t_arrays = list(array.larray for array in arrays)
+
+    broadcasted = torch.broadcast_tensors(*t_arrays)
+
+    out = []
+    for i in range(len(broadcasted)):
+        out.append(factories.array(broadcasted[i], dtype=arrays[i].dtype, device=arrays[i].device))
```
mrfh92, line 118: Maybe I got that wrong, but wasnt there another PR where we wanted to replace the usage of `factories.array` by `DNDarray` for performance reasons?

ClaudiaComito: you're right, and the split axis was missing here as well. Fixed.

PR #1029, opened 2022-09-22
Fixed code that verifies lshape while using is_split

Mystic-Slice:
## Description

The reuse of gshape array as the receiving buffer made the code a bit too difficult to understand. So, created a new array `neighbour_shape` to hold, well...., the neighbour's shape.

The use of `MPI.SUM`, to find if any process has encountered a mismatch in local shapes with its neighbour, leads to errors because of integer limitations. When two or more processes find a mismatch, the huge negative values are added together and I guess because of integer overflow, they disappear into the unknown and finally, we are left with a positive value. So, it does not create any errors when it should. This can be easily solved by using `MPI.MIN` instead.

One more thing that I don't understand is, why `gshape[i]` is allowed to be one less than `lshape[i]` in line [405](https://github.com/helmholtz-analytics/heat/blob/0949a48f94cf1e3e2a5893064f985e37007bb7ea/heat/core/factories.py#L405). I know there must be a reason for it. I just don't understand why.

### Code snippet:
```
import heat as ht
import torch

rank = ht.communication.MPI_WORLD.rank
dim = rank+1

larray = [[0]*dim]*dim
print("Rank", rank, larray)

d = ht.array(larray, is_split=0)

# literally any operation that involves communication 
# between processes will crash
# like this simple print statement
print(d)
```

## Due Diligence

no

skip ci

### DIFF
--- a/heat/core/factories.py
+++ b/heat/core/factories.py
@@ -406,12 +406,12 @@ def array(
                 for i in range(length):
                     if i == is_split:
                         continue
-                    elif lshape[i] != neighbour_shape[i] and lshape[i] - 1 != neighbour_shape[i]:
+                    elif lshape[i] != neighbour_shape[i]:
                         neighbour_shape[is_split] = np.iinfo(neighbour_shape.dtype).min
 
         # sum up the elements along the split dimension
         reduction_buffer = np.array(neighbour_shape[is_split])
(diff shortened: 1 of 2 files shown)


Mystic-Slice:
Changes moved to a different branch. Refer https://github.com/helmholtz-analytics/heat/pull/1034

PR #1034, opened 2022-10-21
`ht.array`, closed loophole allowing `DNDarray` construction with incompatible shapes of local arrays

Mystic-Slice:
## Description

The reuse of gshape array as the receiving buffer made the code a bit too difficult to understand. So, created a new array `neighbour_shape` to hold, well...., the neighbour's shape.

The use of `MPI.SUM`, to find if any process has encountered a mismatch in local shapes with its neighbour, leads to errors because of integer limitations. When two or more processes find a mismatch, the huge negative values are added together and I guess because of integer overflow, they disappear into the unknown and finally, we are left with a positive value. So, it does not create any errors when it should. This can be easily solved by using `MPI.MIN` instead.

One more thing that I don't understand is, why `gshape[i]` is allowed to be one less than `lshape[i]` in line [405](https://github.com/helmholtz-analytics/heat/blob/0949a48f94cf1e3e2a5893064f985e37007bb7ea/heat/core/factories.py#L405). I know there must be a reason for it. I just don't understand why.

### Code snippet:
```
import heat as ht
import torch

rank = ht.communication.MPI_WORLD.rank
dim = rank+1

larray = [[0]*dim]*dim
print("Rank", rank, larray)

d = ht.array(larray, is_split=0)

# literally any operation that involves communication 
# between processes will crash
# like this simple print statement
print(d)
```

## Due Diligence

no

skip ci


heat/core/factories.py
```diff
@@ -395,19 +399,19 @@ def array(
...
+                    elif lshape[i] != neighbour_shape[i]:
+                        neighbour_shape[is_split] = np.iinfo(neighbour_shape.dtype).min
 
         # sum up the elements along the split dimension
-        reduction_buffer = np.array(gshape[is_split])
-        comm.Allreduce(MPI.IN_PLACE, reduction_buffer, MPI.SUM)
+        reduction_buffer = np.array(neighbour_shape[is_split])
+        comm.Allreduce(MPI.IN_PLACE, reduction_buffer, MPI.MIN)
         if reduction_buffer < 0:
             raise ValueError("unable to construct tensor, shape of local data chunk does not match")
```
ClaudiaComito, line 416: Let's rephrase this. How about  "Unable to construct DNDarray. Local data slices have inconsistent shapes or dimensions." ? Or something similar 

Historical note: in the very early phase, our DNDarray class was called `Tensor`.  Any reference to "tensor"  that doesn't refer to `torch.tensor` is outdated, feel free to update to `DNDarray`  it if you spot one.

heat/core/factories.py
```diff
@@ -395,19 +399,19 @@ def array(
...
-                gshape[is_split] = np.iinfo(gshape.dtype).min
+                neighbour_shape[is_split] = np.iinfo(neighbour_shape.dtype).min
             else:
                 # check whether the individual shape elements match
-                comm.Recv(gshape, source=comm.rank - 1)
+                comm.Recv(neighbour_shape, source=comm.rank - 1)
                 for i in range(length):
                     if i == is_split:
                         continue
-                    elif lshape[i] != gshape[i] and lshape[i] - 1 != gshape[i]:
```
ClaudiaComito, line 405: Good catch @Mystic-Slice !

ClaudiaComito (changes requested):
Thanks a lot @Mystic-Slice for tackling this. For me it's good to go, just 2 small changes please:
- Update PR title so that it makes sense in the automated changelog. 
- Update error message (see in-line comment)

Again, great job and thank you so much!

heat/core/factories.py
Mystic-Slice: Thanks @ClaudiaComito. I have made the change.

Mystic-Slice:
I have made the changes. @ClaudiaComito 
Thanks for the review.

ClaudiaComito (approved):

Issue #1035, opened 2022-10-21
[Bug]: in `linalg.solver.lanczos`, dtype of `v0` does not match dtype of input if `v0 is None`


ClaudiaComito:
### What happened?

triggers an error with `torch.dot()` on tensors of non-matching dtypes 

### Version

1.2.x

### Python version

3.8

### PyTorch version

1.11

PR #1038, opened 2022-10-24
Lanczos decomposition `linalg.solver.lanczos`:  Support double precision, complex data types

ClaudiaComito:
## Description

Issue/s resolved: #1035 
Special thanks to @JedrzejMosiezny for reporting and contributing to this fix.

## Changes proposed:
- initialization of `v0` consistent with data type of input
- added support for complex dtypes
- added `test_lanczos` function, incl. correctness tests for `linalg.solver.lanczos` for dtype float32, float64, complex128, complex64

## Type of change

## Memory requirements

`v0` might be float64 if the input is double precision.

## Performance

Single precision input isn't much faster than double precision and accuracy is a lot worse. 

## Due Diligence

no


heat/core/linalg/solver.py
```diff
@@ -106,15 +106,23 @@ def lanczos(
...
-        vr = ht.random.rand(n, split=V.split)
+        try:
+            vr = ht.random.rand(n, split=V.split, dtype=V.dtype, device=V.device)
+        except ValueError:
+            # V is complex
+            vr_dtype = ht.float64 if V.dtype is ht.complex128 else ht.float32
+            vr = (
+                ht.random.rand(n, split=V.split, dtype=vr_dtype, device=V.device)
+                + ht.random.rand(n, split=V.split, dtype=vr_dtype, device=V.device) * 1j
+            )
```
d1saster, line 125: It's ok for me, but there a two points that come to my mind:
- `ValueError` and then inferring that it must be complex might be a heuristic that currently works, but maybe not in the future. One could alternatively also add a check at function start that the supported dtypes are supplied?
- Considering the scope and regime in which lanczos works, I am wondering whether supporting complex numbers sends the right message to the user. Maybe it is not too bad to flag a ValueError in the complex case to highlight that s.p.d. matrices (and hence real-valued `v0`) should be supplied. One can even tell the user that if he thinks the matrix is s.p.d. that `A.real` would already do the trick. On the plus side, this would save half of the unnecessary computations.

ClaudiaComito: Agreed on the heuristic, updated. I've also updated the docs.

mrfh92 (approved):
Looks fine from my point of view.

ClaudiaComito:
@mrfh92 @d1saster thanks for reviewing, I think I've addressed everything, also brought the CI in line with `main`. Need one more approval to run the CI matrix. Thanks a lot!

heat/core/linalg/tests/test_solver.py
```diff
@@ -29,3 +29,104 @@ def test_cg(self):
...
+        self.assertTrue(ht.allclose(V_inv, ht.conj(V).T))
+        # V T V* must be = B, V conjugate transpose = V inverse
+        lanczos_B = V @ T @ V_inv
+        self.assertTrue(ht.allclose(lanczos_B, B))
+
+        # single precision tolerance
+        if int(torch.__version__.split(".")[1]) == 13:
+            tolerance = 1e-3
+        else:
+            tolerance = 1e-4
```
ClaudiaComito, line 69: @Mystic-Slice re: #1028, here's a example of checking for the torch version, there might be more advanced ways that I don't know about though.

ClaudiaComito:
@d1saster @mrfh92 @mtar I'm done here. 
- I've generalized matrix and vector multiplications where needed to support complex input
- I've extended the test coverage as much as I could, the missing coverage is from the re-inizialization of the random vector `vr` when `beta` is 0, a case I've been too lazy to recreate in the tests.

d1saster (approved):

PR #1042, opened 2022-10-28
hSVD (prescribed rank or tolerance) 

mrfh92:
## Description

Hiearchical SVD (hSVD) computes an approximate truncated SVD of $A \in \mathbb{R}^{m \times n}$ utilizing a distributed hiearchical algorithm; the truncation rank is given by $maxrank$, i.e.  if $A = U diag(\sigma) V^T$, $U \in \mathbb{R}^{m \times m}$, $V \in \mathbb{R}^{n \times n}$, $\sigma \in \mathbb{R}^{\min(m,n)}$ is the true SVD of $A$, this routine computes an approximation for $U[:,:maxrank]$ (and $sigma[:maxrank]$, $V[:,:maxrank]$).

There are three routines:
* `hsvd`: "expert" version with all parameters to be set; potential conflicts are not catched 
* `hsvd_rank`: user-friendly version with appropriate default-parameters to compute hSVD with prescribed truncation rank 
* `hsvd_reltol`: user-friendly version with appropriate default-parameters to compute hSVD with prescribed relative reconstruction accuracy 

#### Selected References:
[1] Iwen, Ong. A distributed and incremental SVD algorithm for agglomerative data analysis on large networks. SIAM J. Matrix Anal. Appl., 37(4), 2016.
[2] Himpe, Leibner, Rave. Hierarchical approximate proper orthogonal decomposition. SIAM J. Sci. Comput., 40 (5), 2018.

### Dependencies
hSVD makes use of the pytorch-SVD and MPI-communication. No dependencies on HeAT linear algebra, except for matmul. 

## Type of change

- New feature: hSVD 

## Memory requirements
TBD

## Performance
TBD

## Due Diligence
- [X] All split configurations tested
- [X] Multiple dtypes tested in relevant functions
- [X] Documentation updated (if needed)
- [X] Title of PR is suitable for corresponding CHANGELOG entry

no

### DIFF
--- a/heat/core/linalg/__init__.py
+++ b/heat/core/linalg/__init__.py
@@ -5,3 +5,4 @@ Import all linear algebra functions into the ht.linalg namespace
 from .basics import *
 from .solver import *
 from .qr import *
+from .svdtools import *
--- a/heat/core/linalg/svdtools.py
+++ b/heat/core/linalg/svdtools.py
@@ -1,0 +1,524 @@
+"""
+distributed hierarchical SVD - first draft
+"""
+import numpy as np
+import collections
+import torch
+from typing import Type, Callable, Dict, Any, TypeVar, Union, Tuple
+
+from ..communication import MPICommunication
+from ..dndarray import DNDarray
+from .. import factories
+from .. import types
+from ..linalg import matmul, vector_norm
+from ..indexing import where
+from ..random import randn
+
+from ..manipulations import vstack, hstack, diag, balance
+
+from .. import statistics
+from math import log, ceil, floor, sqrt
+
+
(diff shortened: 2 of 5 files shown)


ClaudiaComito:
@mrfh92 thanks a lot for this, including the expansion of our test matrices repertoire! 

Do you think you could test the algorithm on @JedrzejMosiezny 's test case (see #1047 )? Maybe even use that code for our unit tests, if it makes sense. Thanks a lot!

mrfh92:
Had to remove the test for QR because it failed... I do not know why, because actually there are only changes (compared to main) in svdtools and the corresponding test.

Issue #1065, opened 2022-12-24
[Bug]: `ht.sparse.sparse_csr_matrix` factory method does not validate input data type


Mystic-Slice:
### What happened?

`ht.sparse.sparse_csr_matrix` method used to construct DCSR_matrix does not validate the input object data type. Crashes when any other data type is passed.

### Version

1.2.x

### Python version

3.9

### PyTorch version

1.12

PR #1066, opened 2022-12-24
fix 1065: Verify input object type and layout + Supporting tests

Mystic-Slice:
## Description

The factory method can take in objects of different types and convert to a `torch.Tensor` with the `torch.sparse_csr` layout. If the input object cannot be converted, TypeError is raised.

Issue/s resolved: #1065 

## Type of change

Bug fix

## Due Diligence

no


JuanPedroGHM (approved):
Code looks good to me!

I would add an example of the method with a non-sparse list/tensor as an argument.

Mystic-Slice:
@JuanPedroGHM Thanks for the review!
I have added the example. Should be good to go now.

ClaudiaComito (approved):
Looks good! 👏🏼

PR #1089, opened 2023-02-05
Speed up array declaration when possible in `io`module

shahpratham:
## Description

Replaced factories.array with DNDarray in io.py 
Issue/s resolved: #797 

## Changes proposed:
- factories.array -> dndarray.DNDarray

## Type of change

enhancement

## Due Diligence

no


ClaudiaComito (changes requested):
Proposing a few small changes. Apologies for the delay @shahpratham !

ClaudiaComito (approved):
Thanks a lot!

PR #1094, opened 2023-02-07
Fix division precision error in `random` module

### LINKED ISSUE #1076: [Bug]: Random/Randn does not work in particular setting
### What happened?

`randn` (and `rand`) produce an error in certain constellations

```shell
# run with 11 MPI-processes 

import heat as ht
A = ht.random.randn(6667,3523,dtype=ht.float64,split=1)
```

```shell
RuntimeError: shape '[6667, 320]' is invalid for input of size 2133439
  File "developing.py", line 77, in <module>
    A = ht.random.randn(6667,3523,dtype=ht.float64,split=1)
  File "/home/hopp_fa/heat/heat/core/random.py", line 643, in randn
    normal_tensor = rand(*args, dtype=dtype, split=split, device=device, comm=comm)
  File "/home/hopp_fa/heat/heat/core/random.py", line 472, in rand
    values = __int64_to_float64(torch.stack([x_0, x_1], dim=1).flatten()[lslice]).reshape(
RuntimeError: shape '[6667, 320]' is invalid for input of size 2133439
```

### Version

1.2.x

### Python version

3.8

### PyTorch version

1.12

### MPI version

```shell
mpi4py 3.1.3
```

Mystic-Slice:
## Description

_**Copied over from issue 1076**_
Apparently, torch has some precision problems during division. The division operation, that takes place when the number of elements in each process is calculated, causes some problem.

**Code:**
```
import torch
print(f"Number of elements in tensor: 6667*3523={23487841}")
print(23487841/3523)
print(torch.tensor(23487841)/3523)
```

**Output:**
```
Number of elements in tensor: 6667*3523=23487841
6667.0
tensor(6666.9995)
```

This small precision error is blown up when the number of elements is huge. That's probably why the tests did not catch the bug because the number of elements were small enough.

Issue/s resolved: #1076 

## Type of change

Bug fix

## Due Diligence
- [ ] Title of PR is suitable for corresponding CHANGELOG entry

no

### DIFF
--- a/heat/core/random.py
+++ b/heat/core/random.py
@@ -114,8 +114,8 @@ def __counter_sequence(
         raise ValueError("Shape is to big with {} elements".format(total_elements))
 
     if split is None:
-        values = torch.ceil(total_elements / 2)
-        even_end = total_elements % 2 == 0
+        values = total_elements.item() // 2 + total_elements.item() % 2
+        even_end = total_elements.item() % 2 == 0
         lslice = slice(None) if even_end else slice(None, -1)
         start = c_1
         end = start + int(values)
--- a/heat/core/random.py
+++ b/heat/core/random.py
@@ -125,7 +125,7 @@ def __counter_sequence(
         counts, displs, _ = comm.counts_displs_shape(shape, split)
 
         # Calculate number of local elements per process
-        local_elements = [total_elements / shape[split] * counts[i] for i in range(size)]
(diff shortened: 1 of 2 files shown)


ClaudiaComito (approved):
Nice catch @Mystic-Slice , awesome!

ClaudiaComito (approved):

PR #1105, opened 2023-02-14

Gaurav17Joshi:
## Description

 Adding  a feature to the ..core/linalg/solver.py file. The function triangular_solves solves a system of equation Ax = B, where A and B are two-dimensional DND arrays, such that A is either upper or lower triangular.

## Changes proposed:

## Type of change
 New feature

No

Sample Output:
upper matrix (split)


Gaurav17Joshi:
@mrfh92 , can you help me understand why the pre-commit.ci-pr is showing error? I think it is because of formatting isuues in the code and the file https://github.com/helmholtzanalytics/heat/pull/1105/commits/053bf552edfbd82c86b57c5d01be27bf93514ab6, shows the places where change has to be made, is that the case ?

PR #1107, opened 2023-02-14

Gaurav17Joshi:
## Description

Adding a feature to the ..core/linalg/solver.py file. The function triangular_solves solves a system of equation Ax = B, where A and B are two-dimensional DND arrays, such that A is either upper or lower triangular.

## Changes proposed:

## Type of change
 New feature

No

Sample Output:


Gaurav17Joshi:
@mrfh92 , please help me mitigate these errors.

PR #1108, opened 2023-02-14

Gaurav17Joshi:
## Description

Adding a feature to the ..core/linalg/solver.py file. The function triangular_solves solves a system of equation Ax = B, where A and B are two-dimensional DND arrays, such that A is either upper or lower triangular.

## Changes proposed:

## Type of change

New feature

No

upper matrix (split)


Gaurav17Joshi:
@mrfh92 , can you please review my code and process the PR.

ClaudiaComito:
@Gaurav17Joshi we try our best but please give us at the very least 24 hours to respond. 

The pre-commit problem should be fixed when you merge the latest base branch.

mrfh92:
@Gaurav17Joshi I am sorry that a proper review will take a few days, most probably til next week, because we have some holidays (Carnival) the next days in Germany. 

Nevertheless, if the machine you are currently working on has more than 1 core (lets say 4 or sth like that) or if you have access to an HPC-system, you could already do some scaling tests: fix some A and B (sufficiently large) and measure the time required for executing triangular_solve when running this on 1,2,3,4,... MPI-processes. (You can use, e.g., MPI.Wtime().) Ideally, you could do this for different combinations of split=0,1,None for A and B. Such scaling tests are important to determine the parallel efficiency of your implementation. Perfect (usually not reached in practice) scaling would mean that the computing time on p processes roughly behaves like c/p with c being a constant. 

(If you do not have suitable hardware, this is no problem. We can run these tests for you on our systems instead.)

Gaurav17Joshi:
Thanks for your reply, I work on your suggestions and some other problems of the codebase.

heat/core/linalg/solver.py
```diff
@@ -85,100 +86,262 @@ def lanczos(
...
+            for j in range(m - 1, -1, -1):
+                value = ((bi - b)[j]) / A[j, j]
+                b += value * (A[:, j])
+                x[j, i] = value
+        else:
+            for j in range(0, m, +1):
+                value = ((bi - b)[j]) / A[j, j]
+                b += value * (A[:, j])
+                x[j, i] = value
+
```
ClaudiaComito, line 346: @Gaurav17Joshi this is not going to work, because our matrices are going to be distributed across nodes (or processes, or ranks in MPI terminology). 

That means, if A.shape is (m,m) and we run the solver of 4 nodes:
- the first node (rank 0) will only hold the slice of data A[m/4, m]
- the second (rank 1) will hold A[m/4:m/2, m]
- and so on, and same with B.

We need to research how to break up this solver in:
- process-local operations that can be applied to the local slice of data (these should be `torch` operations)
plus:
- any communication (see the `communication` module) that needs to happen among processes.

ClaudiaComito (changes requested):
@Gaurav17Joshi thanks for taking this on. You've picked a difficult problem to solve! This PR needs more research, we're happy to work with you if you're still interested in contributing.

mrfh92:
Closed due to inactivity. 

Reopening is possible, of course.

Issue #1117, opened 2023-03-06
Set `factories` default to `copy=None`in line with Python array-API


ClaudiaComito:
`heat.array(obj)` by default makes a copy of the underlying object, in sync with [np.array](https://numpy.org/doc/stable/reference/generated/numpy.array.html). Our users deal with massive arrays by definition, so we should consider deviating from NumPy here.

**Related** 
#991 

**Feature functionality**
I propose to switch to the Python array-API [`asarray`](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.asarray.html) default `copy=None` and only create a copy when strictly necessary.

Let's revise and release this quickly @Markus-Goetz

PR #1119, opened 2023-03-15
`ht.array()` default to `copy=None` (e.g., only if necessary)

ClaudiaComito:
## Description

This PR changes the default of the `ht.array(copy)` attribute to None. Previous default was `copy=True`. We introduce a deviation from the NumPy API to reduce the memory footprint of Heat operations. The changes comply with the [Python array-API standard](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.asarray.html). 

- **Previous** default behaviour when creating a DNDarray: `copy=True`
  - `ht.array` by default would make a copy of the underlying data object.
  - Users could disable this by explicitly setting `copy=False`
  - Pros: we'd be sure that by default, process-local data objects are contiguous
  - Cons: obviously by default we are copying huge arrays
  
- **New** default: following the [Python array-API standard for `asarray`](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.asarray.html), the following options are now available for the `copy` attribute:
      - **copy (Optional[bool])** – boolean indicating whether or not to copy the input. If `True`, the function must always copy. If `False`, the function must never copy for input which supports the [buffer protocol](https://docs.python.org/3/c-api/buffer.html) and must raise a `ValueError` in case a copy would be necessary. If `None`, the function must reuse existing memory buffer if possible and copy otherwise. Default: None.

Issue/s resolved: #1117 #990

## Changes proposed:
- `factories.array()` attribute `copy` now defaults to `None`, i.e. existing memory buffer is reused if possible, copied otherwise.
- `copy=False`  raises a `ValueError` if a copy would be necessary. Note: this only holds for "input which supports buffer protocol", i.e. not for scalars.
- `manipulations.diagonal()` modified to return a contiguous local array
- updated tests

## Type of change

The underlying local data object of a DNDarray (`dndarray.larray`) might now be  a [view of a torch tensor](https://pytorch.org/docs/stable/tensor_view.html). Copy might be necessary before communication.

## Memory requirements

In progress
## Performance

In progress

## Due Diligence

yes

The underlying local data object of a DNDarray (`dndarray.larray`) might now be  a [view of a torch tensor](https://pytorch.org/docs/stable/tensor_view.html). Copy might be necessary before communication.


Mystic-Slice (dismissed):
The changes look good to me!

Mystic-Slice:
The tests' failing has something to do with the numpy type system. After searching around, I found out that: 
They have a separate `dtype` structure.
So, `obj.dtype` returns "int64" which is of type `numpy.dtype[int64]` which even though is equal to `numpy.int64`, is not the same object.
```py
> np.dtype("int64") == np.int64 # True
> np.dtype("int64") is np.int64 # False
```
[Source](https://stackoverflow.com/a/26921882)


#### Workaround
It is to make use of the `type` attribute that the `numpy.dtype` object has, which gives us back the usual data type (here, `numpy.int64`)
Add this snippet after [line 516 in heat/core/type.py](https://github.com/helmholtz-analytics/heat/blob/a5515f08501388f95b571eb28e604d88a7dc86c4/heat/core/types.py#L516)
```py
# Get around the numpy type system (np.dtype[<actual_type_we_want>])
if(hasattr(a_type, "type")):
    a_type = a_type.type
  ```
  This solves the issue for me locally.
  
  PS: Are there any other objects that might also have a `type` attribute?

**Comment by ClaudiaComito** [comment]
> The tests' failing has something to do with the numpy type system. After searching around, I found out that: They have a separate `dtype` structure. So, `obj.dtype` returns "int64" which is actually `numpy.dtype[int64]` which even though is equal to `numpy.int64`, is not the same object.
> 
> ```python
> > np.dtype("int64") == np.int64 # True
> > np.dtype("int64") is np.int64 # False
> ```
> 
> [Source](https://stackoverflow.com/a/26921882)
> #### Workaround
> 
> It is to make use of the `type` attribute that the `numpy.dtype` object has, which gives us back the usual data type (here, `numpy.int64`) Add this snippet after [line 516 in heat/core/type.py](https://github.com/helmholtz-analytics/heat/blob/a5515f08501388f95b571eb28e604d88a7dc86c4/heat/core/types.py#L516)
> 
> ```python
> # Get around the numpy type system (np.dtype[<actual_type_we_want>])
> if(hasattr(a_type, "type")):
>     a_type = a_type.type
> ```
> 
> This solves the issue for me locally.
> 
> PS: Are there any other objects that might also have a `type` attribute?

Thanks a lot for reviewing @Mystic-Slice , and for proposing a solution. I have modified `ht.canonical_heat_types` similarly to what you suggested. `torch.dtype`  have no `type` attribute... yet! We will have to revisit if that changes.

**Review by Mystic-Slice — DISMISSED** [review]

File: heat/core/factories.py [hunk 1]
```diff
@@ -151,7 +151,7 @@ def arange(
 def array(
     obj: Iterable,
     dtype: Optional[Type[datatype]] = None,
-    copy: bool = True,
+    copy: bool = None,
```
JuanPedroGHM, line 154: Probably should be 
```python
copy: Optional[bool] = None,
```

heat/core/factories.py
```diff
@@ -198,6 +199,8 @@ def array(
     ------
     NotImplementedError
         If order is one of the NumPy options ``'K'`` or ``'A'``.
+    ValueError
+        If ``copy`` is False but a copy is necessary.
```
JuanPedroGHM, line 204: `TypeError` should also be listed here

JuanPedroGHM (commented):
The changes look good, and test for the new errors are included. It might be good to add in the documentation why a copy is necessary in the case of the `ValueError`, just to give some clarity on the internal workings.

ClaudiaComito:
@JuanPedroGHM is there a good way of checking for memory usage within the benchmarks?

mrfh92 (approved):
From my point of view the changes look fine. Therefore, I recommend merging if the CI runs through.

Issue #560, opened 2020-04-24
reshape speedup


coquelin77:
**Related**
The current reshape function provides correct results but it is slow.

**Feature functionality**
Faster (ideally close to the speed of Alltoallv) reshaping

**Additional context**
on the branch `reshape_speedup` I have already implemented a jitted version of the main work function within reshape but this only shows a limited benefit. 

current implementation with an example for the following script

```python
import heat as ht
for _ in range(4):
    st = ht.zeros((10000, 1000), split=1)
    t = time.process_time()
    a = ht.reshape(st, (10000000, 1))
    print(a.split, a.gshape, time.process_time() - t)
```
without jit (current):
```
[0] reshape -1
[0] 0 (10000000,) 0.751831633
[0] 0 (10000000,) 0.705800617
[0] 0 (10000000,) 0.6433533120000003
[0] 0 (10000000,) 0.5882456029999998
[2] reshape -1
[2] 0 (10000000,) 0.7038163030000001
[2] 0 (10000000,) 0.7081146759999999
[2] 0 (10000000,) 0.6257028890000003
[2] 0 (10000000,) 0.5925788019999998
[3] reshape -1
[3] 0 (10000000,) 0.691055509
[3] 0 (10000000,) 0.714112522
[3] 0 (10000000,) 0.5806882069999997
[3] 0 (10000000,) 0.5935307019999998
[1] reshape -1
[1] 0 (10000000,) 0.785425587
[1] 0 (10000000,) 0.7074262409999998
[1] 0 (10000000,) 0.6428382629999998
[1] 0 (10000000,) 0.5928067560000003
```
with jit:
```
[2] reshape -1
[2] 0 (10000000,) 0.602714447
[2] 0 (10000000,) 0.6912576529999999
[2] 0 (10000000,) 0.6200092909999999
[2] 0 (10000000,) 0.6013838640000002
[0] reshape -1
[0] 0 (10000000,) 0.602595418
[0] 0 (10000000,) 0.6938838409999999
[0] 0 (10000000,) 0.6186583719999998
[0] 0 (10000000,) 0.6019771760000001
[1] reshape -1
[1] 0 (10000000,) 0.602809302
[1] 0 (10000000,) 0.6920380610000001
[1] 0 (10000000,) 0.6198947539999999
[1] 0 (10000000,) 0.6025693739999998
[3] reshape -1
[3] 0 (10000000,) 0.6025081130000001
[3] 0 (10000000,) 0.693247552
[...]

Issue #1053, opened 2022-11-26
Support Apple's MPS backend


ClaudiaComito:
Since version 1.12, PyTorch supports GPU acceleration via Apple's Metal Performance Shaders (MPS) backend. 
https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/

**Feature functionality**
`device="gpu"` should be a valid argument for `ht.factories` if the MPS backend is available for torch versions >= 1.12.

Ongoing research on branch [`features/1053-support-Apple-silicon-GPUs`](https://github.com/helmholtz-analytics/heat/tree/features/1053-support-Apple-silicon-GPUs), will document here.

ClaudiaComito:
First few tests (PyTorch 1.13):
- only single precision supported
- multi-GPU setup unclear
- PyTorch support does not cover all operations

ClaudiaComito:
In progress #1129

Issue #874, opened 2021-09-27
`reshape` memory requirements


ClaudiaComito:
**Description**
`ht.reshape` memory requirements are quite huge, including in those cases where redistribution is not necessary.

Example 1: `DNDarray.split not None`, but  **single process**:

```
BEFORE RESHAPE: Current memory usage is 0.002501MB; Peak was 0.003077MB
after torch.reshape: Current memory usage is 0.003669MB; Peak was 0.004101MB
after ht.reshape: Current memory usage is 0.372577MB; Peak was 0.382753MB
```

Example 2: `DNDarray.split == new_split` and local tensor sizes are "reshapeable" into new local shapes. **On 2 processes** (addressed in #873) 

```
[1,0]<stdout>:BEFORE RESHAPE: Current memory usage is 0.002501MB; Peak was 0.003077MB
[1,0]<stdout>:after torch.reshape: Current memory usage is 0.003669MB; Peak was 0.004101MB
[1,0]<stdout>:after ht.reshape: Current memory usage is 0.372689MB; Peak was 0.382889MB
[1,1]<stdout>:BEFORE RESHAPE: Current memory usage is 0.002501MB; Peak was 0.003105MB
[1,1]<stdout>:after torch.reshape: Current memory usage is 0.003669MB; Peak was 0.004101MB
[1,1]<stdout>:after ht.reshape: Current memory usage is 0.372806MB; Peak was 0.383006MB
```

Code below covers both examples.

```python
tracemalloc.start()
t_x = torch.arange(100000).reshape(10,-1,10)
x = ht.array(t_x, split=1)
current, peak = tracemalloc.get_traced_memory()
print(f"BEFORE RESHAPE: Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")

t_x = t_x.reshape(10, -1)
current, peak = tracemalloc.get_traced_memory()
print(f"after torch.reshape: Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")

x = x.reshape(10, -1)
current, peak = tracemalloc.get_traced_memory()
print(f"after ht.reshape: Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")
```

**To Reproduce**
1. Which module/class/function is affected?
`manipulations.reshape`
2. What are the circumstances under which the bug appears?
see above
3. What is the exact error message / erroneous behavior?
No error message but show-stopper for large-data calculations

**Expected behavior**
[...]

PR #1125, opened 2023-03-20
Features/560 reshape speedup

ClaudiaComito:
## Description

While testing Heat+torch 2.0 on Apple MPS #1053 I noticed a major bottleneck in `ht.reshape`  even on 1 process. We had been discussing this some time ago, #560, #874

The main slow-down arises in setting up  the `Alltoallv` communication necessary when the input DNDarray is distributed along non-zero axes (`ht.reshape(x, new_shape, new_split)` with `x.split > 0` and potentially `new_split >0`). 

I ran some experiments and realized that resplitting in place (input split to 0, and then if necessary 0 to new_split) while inelegant still performs much better than the  current implementation. This PR implements a pragmatic solution based on `dndarray.resplit_` and `dndarray.redistribute_`.

~~Below some plots from tests on my laptop - matrices are still relatively small and 1-process performance is still better than 2-processes, will replace with scaling tests on HDF ML later.~~

**UPDATED**

Below some tests run on HDF ML, CPU only, 2 CPUs per node, 12 threads per CPU. 
Initial matrix: `A = ht.zeros((1000, size), split=key[0])`
Timed operation: `B = ht.reshape(A, (10000000, -1), new_split=key[1])`
Sampled sizes: `sizes = [500000*n for n in [1, 2, 4, 8, 16, 32, 64, 128, 256]]`
Sampled split combinations: `[[0, 0], [1, 1], [1, 0], [0, 1]]` (only showing plots for  split combi 0-0, 1-1) 
Minimum speed-up: ~8.5 for split 0-0, ~3 for split 1-1 which requires more communication.

reshape_weak

reshape_strong_8x

reshape_strong_64x

Issue/s resolved: #560, #874 

## Changes proposed:
- replace `Alltoallv` call with resplitting in place when necessary
-  updated documentation
-
-

## Type of change

Performance enhancement
## Memory requirements

In progress
## Performance

In progress, also see plots

## Due Diligence

no


mrfh92 (approved):
These changes look fine to me.

mrfh92:
I can run the additional scaling tests

ClaudiaComito:
> 

Hi @mrfh92 , thanks for the review, I actually ran the scaling tests of this implementation. I will run comparison tests with the released `reshape` as well.

mrfh92:
**Strong scaling for a short-fat matrix and split=0,1** 

reshape_main_cpu
reshape_main_gpu
reshape_new_cpu
reshape_new_gpu

**Strong scaling for different shapes from short-fat to tall-skinny (split=1 only)** 

reshape_main_cpu
reshape_main_gpu
reshape_new_cpu
reshape_new_gpu

mrfh92:
It looks like that the "old" reshape is significantly slower than the "new" one (in particular for a small number of processes), but the "old" reshape scales well when increasing the number of processes; the "new" reshape does not seem to scale well (which is expected, because resplit will always have runtime proportional to the number of processes...)

ClaudiaComito:
> It looks like that the "old" reshape is significantly slower than the "new" one (in particular for a small number of processes), but the "old" reshape scales well when increasing the number of processes; the "new" reshape does not seem to scale well (I believe that is due to the usage of resplit which will always have runtime proportional to the number of processes...)
> 
> Nevertheless, this does not necessarily imply that the "new" reshape is not an improvement compared to the "old" one; in contrary, the "new" version is that much more faster for small number of processes that it stil outperforms the "old" one there. However, there we should determine whether there is some break-even point at which the "old" one becomes better.

Thanks a lot @mrfh92 for all the tests. I went back and finished the tests I had started on the cluster, adding test runs on the release (`main` branch) version of reshape. I've updated the PR description with plots and more details on the experiments.

mrfh92 (approved):
The new plots in the PR description show that my worries on possible performance degradation w.r.t. scalability were actually unjustified. Therefore, I now recommend merging if the CI runs through.

PR #1136, opened 2023-03-31
Fixed PyTorch version check in `sparse` module

Mystic-Slice:
## Description

The earlier check for torch version fails in the new major version. This fix solves the issue.

## Type of change

Bug fix

no

### DIFF
--- a/heat/sparse/factories.py
+++ b/heat/sparse/factories.py
@@ -94,7 +94,7 @@ def sparse_csr_matrix(
     (indptr: tensor([0, 1, 3, 4]), indices: tensor([2, 0, 2, 2]), data: tensor([1, 1, 2, 3]), dtype=ht.int64, device=cpu:0, split=None)
     """
     # version check
-    if torch.__version__ < "1.10":
+    if int(torch.__version__.split(".")[0]) <= 1 and int(torch.__version__.split(".")[1]) < 10:
         raise RuntimeError(f"ht.sparse requires torch >= 1.10. Found version {torch.__version__}.")
 
     # sanitize the data type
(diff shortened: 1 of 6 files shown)


heat/sparse/factories.py
```diff
@@ -94,7 +94,7 @@ def sparse_csr_matrix(
     (indptr: tensor([0, 1, 3, 4]), indices: tensor([2, 0, 2, 2]), data: tensor([1, 1, 2, 3]), dtype=ht.int64, device=cpu:0, split=None)
     """
     # version check
-    if int(torch.__version__.split(".")[1]) < 10:
+    if torch.__version__ < "1.10":
```
ClaudiaComito, line 97: oh, I didn't know this was possible! 

I had to go and time these two methods. We do many many checks, and those millionths of seconds pile up. 🤭 

Turns out the string check (this PR) is about 10 times slower than the int check (previous solution) even after having to manually split the string. The int check is even faster for torch >=2, as the first condition is sufficient to make the decision.

Summary: I'd go for 
```python
if int(torch.__version__.split(".")[0]) <= 1 and int(torch.__version__.split(".")[1]) < 10
```
solution in `factories` as well. What do you think?

ClaudiaComito (changes requested):
Thanks a lot @Mystic-Slice for addressing this so soon!

heat/sparse/factories.py
Mystic-Slice: Interesting to know that it is soo much slower.
I have reverted back to the older method of version checking.

ClaudiaComito (approved):

PR #1149, opened 2023-04-30
Fixed bugs due to behaviour change in torch.nn.Module.zero_grad() and torch.Tensor.item() in torch 2.0

Mystic-Slice:
## Description

This PR fixes the following errors in the tests.
**1. `AttributeError: 'NoneType' object has no attribute 'data'`**
Due to a change in default behaviour of `torch.nn.Module.zero_grad()` as stated in the "Backwards Incompatible Changes" in the [2.0 release](https://github.com/pytorch/pytorch/releases/tag/v2.0.0)

**Solution:** Set `set_to_none` to `False` explicitly to preserve previous heat behaviour. 
Can also be modified to follow Pytorch and set the grads to `None` by default to reduce peak memory usage. But it would be a breaking change.

**2. `RuntimeError: a Tensor with * elements cannot be converted to Scalar`**
It is expected to raise a `ValueError`. But, it raises a `RuntimeError`. This change is not mentioned anywhere in the release notes. Even in the torch [source code](https://github.com/pytorch/pytorch/blob/6c934a89a725fd5d171b52a37cbc58e198edf4d6/torch/_refs/__init__.py#L1857), it seems to raise a `ValueError`. 

**Solution:** Changing heat tests and `DNDarray` code to expect a `RuntimeError` fixes the failing tests. But I can't explain why.

**P.S.** The first commit("Support latest PyTorch release") in this PR is not intentional. I can't seem to get rid of it either.
## Type of change

Bug fix

## Due Diligence

yes / no

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -1484,7 +1484,7 @@ class DNDarray:
         for c, k in enumerate(key):
             try:
                 key[c] = k.item()
-            except (AttributeError, ValueError):
+            except (AttributeError, RuntimeError):
                 pass
 
         rank = self.comm.rank
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -1484,7 +1484,7 @@ class DNDarray:
         for c, k in enumerate(key):
             try:
                 key[c] = k.item()
-            except (AttributeError, RuntimeError):
+            except (AttributeError, ValueError, RuntimeError):
                 pass
 
         rank = self.comm.rank
--- a/heat/core/tests/test_dndarray.py
+++ b/heat/core/tests/test_dndarray.py
@@ -579,7 +579,7 @@ class TestDNDarray(TestCase):
         self.assertEqual(type(x.item()), float)
 
         x = ht.zeros((1, 2))
-        with self.assertRaises(ValueError):
+        with self.assertRaises(RuntimeError):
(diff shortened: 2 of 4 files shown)


ClaudiaComito (approved):
Wonderful, thank you so much @Mystic-Slice ! I changed the base branch to the torch 2.0 PR #1134

ClaudiaComito:
>
> **2. `RuntimeError: a Tensor with * elements cannot be converted to Scalar`** It is expected to raise a `ValueError`. But, it raises a `RuntimeError`. This change is not mentioned anywhere in the release notes. Even in the torch [source code](https://github.com/pytorch/pytorch/blob/6c934a89a725fd5d171b52a37cbc58e198edf4d6/torch/_refs/__init__.py#L1857), it seems to raise a `ValueError`.
> 
> **Solution:** Changing heat tests and `DNDarray` code to expect a `RuntimeError` fixes the failing tests. But I can't explain why.

The exception is caught and handled by heat if the torch operation returns the specified errors. The tests too expect a specific exception. Your solution is fine, as far as I can tell.


> 
> **P.S.** The first commit("Support latest PyTorch release") in this PR is not intentional. I can't seem to get rid of it either.

It probably comes from the base branch, don't worry about it. 

Thanks again for these fixes, I'll merge them now.

Issue #694, opened 2020-11-11
Allow the HIP backend


Markus-Goetz:
**Feature functionality**
Since torch 1.7 other backends than CPU and CUDA are supported. Among them is the HIP, aka AMD's CUDA, backend. HeAT should be able to use it. Moreover, when using the 'gpu' device in HeAT, the framework should automatically determine whether CUDA or HIP is meant.

**Additional context**
Requires torch 1.7

PR #1152, opened 2023-05-17
support amd HIP

mtar:
## Description

This Pull Request skips some tests such that it can be run on AMD GPU. Additionally branches starting with 'support/' are tested on push events.

Issue/s resolved: #694

## Changes proposed:
- skip tests using special cuda libraries
- push trigger event on support branch

## Type of change
- testing

## Due Diligence
- [ ] Title of PR is suitable for corresponding CHANGELOG entry

yes / no

### DIFF
--- a/.github/workflows/CIBase.yml
+++ b/.github/workflows/CIBase.yml
@@ -16,6 +16,6 @@ jobs:
          curl -s -X POST \
            --fail \
            -F token=${{ secrets.CB_PIPELINE }} \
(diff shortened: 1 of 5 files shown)


ClaudiaComito (approved):
Thanks a lot @mtar ! Can you create an issue with the HIP QR error message, so we can follow up on it?

ClaudiaComito (approved):

ClaudiaComito (approved):

ClaudiaComito (approved):

Issue #1153, opened 2023-05-18
Implement `__array__()` method for DNDarrays


ClaudiaComito:
**Related**
#1031 

**Feature functionality**
Provide a **view** of the process-local array (a slice of the global array) as a numpy ndarray.

This is very different from the `dndarray.numpy()` method that provides a **copy** of the global array (gathered on each MPI process) as a numpy ndarray.

Potential confusion will be addressed in the documentation. 

**Additional context**
Required for interoperability with the numpy ecosystem including the [xarray](https://docs.xarray.dev/en/stable/user-guide/data-structures.html) library. Missing `__array__` method results in numpy not recognizing the DNDarray as an array object, e.g.:
```
>>> import heat as ht
>>> import numpy as np
>>> a = ht.arange(10)
>>> a
DNDarray([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=ht.int32, device=cpu:0, split=None)
>>> np.asarray(a)
array([DNDarray(0, dtype=ht.int32, device=cpu:0, split=None),
       DNDarray(1, dtype=ht.int32, device=cpu:0, split=None),
       DNDarray(2, dtype=ht.int32, device=cpu:0, split=None),
       DNDarray(3, dtype=ht.int32, device=cpu:0, split=None),
       DNDarray(4, dtype=ht.int32, device=cpu:0, split=None),
       DNDarray(5, dtype=ht.int32, device=cpu:0, split=None),
       DNDarray(6, dtype=ht.int32, device=cpu:0, split=None),
       DNDarray(7, dtype=ht.int32, device=cpu:0, split=None),
       DNDarray(8, dtype=ht.int32, device=cpu:0, split=None),
       DNDarray(9, dtype=ht.int32, device=cpu:0, split=None)],
      dtype=object)
```
whereas interoperability requires:
```
>>> np.asarray(a)
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
```

Issue #1031, opened 2022-09-28
Support distribution of `xarray`


ClaudiaComito:
See https://docs.xarray.dev/en/stable/

If I understand correctly, an `xarray` object is made up of the actual `data` array (np.ndarray), and ~~1-D~~ `coordinates` arrays (dictionaries?) that map `data` dimensions and indices to meaningful physical quantities. 

For example, if `xarray` is  a matrix of coordinates `(date, temperature)`, users will be able to perform
```python
mean_temp = xarray['2010_01_01', '2010_12_31'].mean()
```

**Feature functionality**

Enable distribution of xarray object, allow named dimensions, keep track of `coordinates` arrays, one of which will be distributed.

Example, : 
```python
ht_xarray = ht.array(xarray, split="date")
ht_mean_temp = ht_xarray['2010_01_01', '2010_12_31'].mean() # memory-distr. operation
```

Check out Pytorch's [named tensors](https://pytorch.org/docs/stable/named_tensor.html) functionality.

**Additional context**
Initiating collaboration with N. Koldunov @koldunovn at Alfred Wegener Institute (Helmholtz centre for polar and marine research). 
Also interesting for @kleinert-f, @ben-bou
Tagging @bhagemeier for help with implementation.

koldunovn:
Coordinates are not necessarily 1D arrays. For example for curvilinear grids covering the Earth's surface one would have to describe positions of (e.g. centers) of grid points as 2D arrays (one for `lon` and one for `lat`), and coordinate arrays then will be 2D.

ClaudiaComito:
Got it, thanks @koldunovn, edited original post.

ClaudiaComito:
Implementation discussed during devs meeting Sept 28. **Preliminary** scheme:

- introduce `dndarray.coordinates` attribute (default=`None`)
- `ht.array()` to accept `xarray` input, interpret string `split`, populate `dndarray.coordinates`, chunk data AND split coordinate 
- every operation needs to check `if not coordinates:` or similar, quantify and minimize performance hit

@Markus-Goetz FYI

ClaudiaComito:
also tagging @Mystic-Slice as they have shown interest in working on this :wave:

ClaudiaComito:
Interesting discussion over at [array-api](https://github.com/data-apis/array-api/issues/4#issuecomment-1402323325) about **single-node** parallelism, but xarray also mentioned in distributed execution context. 

[...]

PR #1154, opened 2023-05-19
Introduce `DNDarray.__array__()` method

ClaudiaComito:
## Description

This method returns **a view** of the underlying process-local array as numpy ndarray. If the DNDarray is on GPU, the array is copied to CPU first.

With this change, passing a DNDarray to `np.asarray()` returns a valid np.ndarray. This is needed for interoperability with other libraries in the numpy ecosystem.

Basic example with `xarray`:

```
>>> import xarray as xr
>>> import heat as ht
>>> a = ht.arange(10).reshape(5,2)
>>> a
DNDarray([[0, 1],
          [2, 3],
          [4, 5],
          [6, 7],
          [8, 9]], dtype=ht.int32, device=cpu:0, split=None)
>>> xr_a = xr.DataArray(a)
>>> xr_a
<xarray.DataArray (dim_0: 5, dim_1: 2)>
array([[0, 1],
       [2, 3],
       [4, 5],
       [6, 7],
       [8, 9]], dtype=int32)
Dimensions without coordinates: dim_0, dim_1
```

If the DNDarray is distributed, the example above will return an xarray of the local slice of data on each MPI process.

Related: #1031  

Issue/s resolved: #1153 

## Changes proposed:
- Implement `DNDarray.__array__()` method
- Update documentation of `DNDarray.numpy()` method
- Implement tests
-

## Type of change

## Memory requirements

NA
## Performance

NA

## Due Diligence

no

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -461,9 +461,10 @@ class DNDarray:
 
     def __array__(self) -> np.ndarray:
         """
-        Returns a view of the process-local array as a numpy ndarray.
+        Returns a view of the process-local array as a numpy ndarray, if the array resides on CPU.
+        Otherwise, it returns a copy of the process-local array as numpy ndarray on CPU.
         """
-        return self.larray.__array__()
+        return self.larray.cpu().__array__()
 
     def astype(self, dtype, copy=True) -> DNDarray:
         """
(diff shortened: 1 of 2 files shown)


mrfh92 (approved):
From my point of view the changes look fine 👍 
Therefore I recommend merging if the CI succeeds.

Issue #1157, opened 2023-05-22
[Bug]: Continuous Benchmarking might measure wrong times 


mrfh92:
### What happened?

I had a look into the continuous benchmarking scripts and found that some benchmarks are defined as follows: 
```
@monitor()
def matmul_cpu_split_0(n: int = 3000):
    a = ht.random.random((n, n), split=0, device="cpu")
    b = ht.random.random((n, n), split=0, device="cpu")
    a @ b
```
Therefore, the times for creating the test data are measured together with the time of the actual function to be benchmarked. 
I suggest to create the data always separately and to measure only the time of the actual function. 

### Version

1.2.x

### Python version

None

### PyTorch version

None

PR #1158, opened 2023-05-22
Continuous benchmarks for linalg: try to separate data creation from the actual tests

mrfh92:
I tried to separate the creation of test data from the actual tests in the `linalg`-file of continuous benchmarking 
(see also #1157 )

For `reshape` I have no idea how to proceed so far...

### DIFF
--- a/benchmarks/cb/linalg.py
+++ b/benchmarks/cb/linalg.py
@@ -4,35 +4,52 @@ from perun.decorator import monitor
 
 
 @monitor()
-def matmul_cpu_split_0(n: int = 3000):
-    a = ht.random.random((n, n), split=0, device="cpu")
-    b = ht.random.random((n, n), split=0, device="cpu")
+def matmul_cpu_split_0(a, b):
     a @ b
 
 
 @monitor()
-def matmul_cpu_split_1(n: int = 3000):
-    a = ht.random.random((n, n), split=1, device="cpu")
-    b = ht.random.random((n, n), split=1, device="cpu")
+def matmul_cpu_split_1(a, b):
     a @ b
 
 
 @monitor()
-def qr_cpu(n: int = 2000):
+def qr_cpu_split_0(a):
     for t in range(1, 3):
-        for sp in range(2):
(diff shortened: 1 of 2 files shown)


ClaudiaComito:
@mrfh92 I modified the reshape benchmark, do you think this works?

mrfh92:
@ClaudiaComito I guess that this will work, however I am not sure whether we want to go the way to perform tests with several different shapes at once since in this case we cant know in the end which shape/size/split etc. contributes the most, both w.r.t. time (and possibly lateron memory as well). 

So Im not sure whether we should rather restrict timings always to a single case, i.e. one split and one shape instead of taking times together of a bunch of configurations.

ClaudiaComito:
> @ClaudiaComito I guess that this will work, however I am not sure whether we want to go the way to perform tests with several different shapes at once since in this case we cant know in the end which shape/size/split etc. contributes the most, both w.r.t. time (and possibly lateron memory as well).
> 
> So Im not sure whether we should rather restrict timings always to a single case, i.e. one split and one shape instead of taking times together of a bunch of configurations.

You're right @mrfh92 . I'll not work on this for the next couple days. Let me know if you need me to do something.

mrfh92:
@ClaudiaComito @JuanPedroGHM I think we should discuss this in the next project meeting, more precisely we should address the following question: how to choose a (hopefully) small set of illustrative and representative benchmark examples?

mrfh92:
May be worth waiting for this until #1151 is finished and what kind of benchmarks we can run on the new runners

mrfh92:
Actually, this issue is a bit more complicated: since we are sometimes working with randomly generated test data it might be useful to sample new test data for each measurement (if the runtime of a function depends on the actual data) ...

mrfh92:
Actually, after having migrated the benchmark runs to gitlab I would suggest to use larger test data anyway (e.g. MNIST) ... so probably we should address recreating benchmarks then

mrfh92:
I have just seen that this problem does not appear anymore in the current (main) version of the continuous benchmarks.

PR #1160, opened 2023-05-25
hSVD: Concatenate arrays to reduce communication

ClaudiaComito:
## Description

A proposal to reduce communication in #1126 

## Changes proposed:
- for every iteration, add a row to `U_loc` and store `err_square_loc` in it.
- communicate `U_loc` and `err_square_loc` in one single MPI call instead of one each
- separate tensors again after communication
-

## Type of change

New feature 
## Memory requirements

NA
## Due Diligence

no


ClaudiaComito:
> I conducted tests with the "standard 20GB data set" on a GPU cluster: the modified version significantly outperforms the old ones, with larger advantage for higher maxrank hsvd_gpu_terrabyte

"Significantly" is very polite @mrfh92 . Thanks a lot for running the tests! I assume the high-rank case requires more iterations so the effect is amplified?

I'll clean up the code and submit for your review then.

PR #1175, opened 2023-07-03
Introduce weekly benchmarking 

mrfh92:
As discussed in the PR-Talk today, we are going to restructure our benchmarkings as follows: 

* continuous benchmarking (as we have already): rather small benchmarks that are run after each commit on `main` on gitlab-runners 
* **weekly benchmarking** to be set up on the *****-system: once a week run "larger" benchmarks that are more illustrative 

This PR is to introduce the latter weekly benchmarking. 

Done so far: 
* examples for `hsvd`, `matmul`, `qr` and `lanczos` using the MNIST-data set as test data


mrfh92:
@JuanPedroGHM I have added "larger" benchmarks for `hsvd` in the new `/benchmark/wb`-folder that you might use as examples when setting up the configuration

Issue #1040, opened 2022-10-26
[Feature-384 Sparse] Printing of DCSR_matrix


Mystic-Slice:
**Related**
Pending work in https://github.com/helmholtz-analytics/heat/pull/1028

**Feature functionality**
Decide and implement a format for displaying `DCSR_matrix`

yash2798:
Hello @Mystic-Slice , I want to work on this issue. Can you assign this to me?
Thanks a lot!

Mystic-Slice:
Hi @yash2798. Thanks for your interest.
Yes, please feel free to work on this and create a PR once you are done.

Try looking at how printing works in DNDarray for help in the implementation aspect. 
Also, look at the output of different libraries for their sparse matrix formats (especially scipy and torch) and create a format that you think will fit this library. Let us make any modifications later if needed.

mrfh92:
Stil open and relevant; related PR is ongoing work. 
(Reviewed within #1109)

PR #1176, opened 2023-07-05
Introduced a format for displaying the DCSR_matrix

Sai-Suraj-27:
## Description
Added a `__str__()` method for printing the DCSR_matrix.
PyTorch displays the `sparse tensors` in this format:
image
The introduced format displays the following:
image

Issue/s resolved: #1040

## Changes proposed:
- Added a method for printing the DCSR_matrix.

## Type of change

## Due Diligence

No

### DIFF
--- a/heat/sparse/dcsr_matrix.py
+++ b/heat/sparse/dcsr_matrix.py
@@ -337,3 +337,21 @@ class DCSR_matrix:
         if self.comm.rank != 0:
             return ""
         return print_string
+
+    def __str__(self) -> str:
+        """
+        Computes a string representation of the passed ``DCSR_matrix``.
+        """
+        size = self.__gshape
+        nnz = self.__gnnz
+
+        print_string = (
+            f"indices={self.indices}, "
+            f"data={self.data}, DCSR_matrix(size={size}, "
+            f"nnz={nnz}, split={self.__split})"
+        )
+
+        if self.__comm.rank != 0:
+            return ""
+
+        return print_string
--- a/heat/sparse/dcsr_matrix.py
+++ b/heat/sparse/dcsr_matrix.py
@@ -347,8 +347,8 @@ class DCSR_matrix:
 
         print_string = (
             f"indices={self.indices}, "
-            f"data={self.data}, DCSR_matrix(size={size}, "
-            f"nnz={nnz}, split={self.__split})"
+            f"data={self.data}, "
+            f"size={size}, nnz={nnz}, split={self.__split})"
         )
 
         if self.__comm.rank != 0:
--- a/heat/sparse/dcsr_matrix.py
+++ b/heat/sparse/dcsr_matrix.py
@@ -351,7 +351,7 @@ class DCSR_matrix:
             f"size={size}, nnz={nnz}, split={self.__split})"
         )
 
-        if self.__comm.rank != 0:
+        if self.comm.rank != 0:
             return ""
 
         return print_string
--- a/heat/sparse/dcsr_matrix.py
+++ b/heat/sparse/dcsr_matrix.py
@@ -346,9 +346,9 @@ class DCSR_matrix:
         nnz = self.__gnnz
 
         print_string = (
-            f"indices={self.indices}, "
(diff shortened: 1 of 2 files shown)


mrfh92:
@Sai-Suraj-27 Regarding the introduced format I have a question: your picture of the PyTorch print result shows that indices are given by a 2D-tensor (which makes sense because specifying the position of a non-zero entry in a matrix requires us to state row- and column-index). In your print result for DCSR, the index-tensor is 1D. Is this due to the CSR-format?

Sai-Suraj-27:
> @Sai-Suraj-27 Regarding the introduced format I have a question: your picture of the PyTorch print result shows that indices are given by a 2D-tensor (which makes sense because specifying the position of a non-zero entry in a matrix requires us to state row- and column-index). In your print result for DCSR, the index-tensor is 1D. Is this due to the CSR-format?

@mrfh92 Yes, sir. In PyTorch's sparse tensor representation, the indices are stored as a 2D tensor where each row represents the coordinates (row and column indices) of a non-zero entry in the matrix. On the other hand, In CSR format, the indices are stored as a 1D tensor, where each element represents the column index of a non-zero entry.
We can also add and display the `indptr` array which will tell the number of non-zero entries in each row and from that and the `indices` tensor one can easily find out the (row, col) of all non-zero elements. Please tell me your opinion and if you want to change anything else in this format of displaying the DCSR_matrix.

mrfh92:
Hi @Sai-Suraj-27,
thanks for your fast anwer :) I would like to suggest to include the `indptr` as well such that the output of print gives the user full information needed to "reconstruct" the tensor. However, as I remember @ClaudiaComito has mentored the implementation of the sparse-module so far; so maybe its a better idea to wait until next week when she is back since she can give a more detailed feedback.

Sai-Suraj-27:
> Hi @Sai-Suraj-27, thanks for your fast anwer :) I would like to suggest to include the `indptr` as well such that the output of print gives the user full information needed to "reconstruct" the tensor. However, as I remember @ClaudiaComito has mentored the implementation of the sparse-module so far; so maybe its a better idea to wait until next week when she is back since she can give a more detailed feedback.

Ok, sir. I will wait for the feedback.

Sai-Suraj-27:
> Hi @Sai-Suraj-27, I think it would be the best option to leave `indices` and `data` completely away when displaying a sparse DNDarray. So is suffices to display that it is an object of class `DCSR_matrix`, its `size`, `split` and the number of non-zero entires `nzz`. The reason is that usually sparse arrays are used in a context where the data is so large that displaying single entries does not make sense anymore.

@mrfh92 sir, so do you want me to update the function such that we will be printing like this: 
`DCSR_matrix(size=(3, 3), nnz=8, split=None)` (without `indptr`, `data`, and `indices`)

heat/sparse/dcsr_matrix.py
```diff
@@ -337,3 +337,22 @@ def __repr__(self) -> str:
         if self.comm.rank != 0:
             return ""
         return print_string
+
+    def __str__(self) -> str:
```
Mystic-Slice, line 341: It would be better if this function is moved to the `core/printing` module. There are cases like local print vs global print to be handled. See the `__str__` method of the `DNDarray` class for more information.

And it is usually best to use the printing format used by PyTorch since it gives more information to the user. The information outputted by `__str__` should prioritize user-readability more than anything else. Just the `indptr` and `indices` would be too difficult to understand. I, personally, would prefer to see the exact coordinates being printed out. I say this because the users of Heat shouldn't be expected to know about the CSR matrix and its format. They should just be able to use this more-efficient data structure without having to learn too much about it.

Mystic-Slice: What @mrfh92 says also seems acceptable to me. Maybe the users won't want to see the actual data points. But my only problem with this solution is, the users wouldn't be able to see the data points even if they wanted to. There is no other way for them to get the coordinates if we don't show it in the `print` output.

Mystic-Slice (changes requested):

Issue #889, opened 2021-11-09
ht.allclose should factor in the dtype of the inputs when determining the limits


coquelin77:
**Feature functionality**
see title

**Motivation**
When using ht.allclose with float32, we expect to have lower precision. But the current limits on `ht.allclose` are derived from numpy (https://numpy.org/doc/stable/reference/generated/numpy.allclose.html) which uses 64 bit numbers as default.

NamanKothari5:
Hi, I m Naman Kothari a sophomore in Information Technology. I wish to work on this issue can you help me out in getting started

ClaudiaComito:
This is still open.

ClaudiaComito:
We won't address this issue. The background is this: `atol` and `rtol` would have to be set manually for many operations anyway, the numpy defaults aren't particularly related to 64-bit precision (otherwise they would be much lower), and the numpy defaults are identicaly to the torch defaults. 

`ht.allclose` will stick to the numpy defaults, deviations will have to be addressed in single test functions like we have been doing so far.

Thanks @Sai-Suraj-27 for looking into this.

PR #1177, opened 2023-07-08
Updated the `allclose` function to have less precision according to the dtype of input DNDarrays.

Sai-Suraj-27:
## Description
Updated, the `allclose` function (in `heat/core/logical.py`) to have lesser precision according to the dtype of input DNDarrays.

Issue/s resolved: #889

## Changes proposed:
- Adjusting the tolerance values allows for a looser comparison of float32 arrays.
- Now `ht.allclose` will factor in the dtype of the inputs when determining the limits.

## Type of change

## Due Diligence

Yes.
`ht.allclose`


heat/core/logical.py
```diff
@@ -106,7 +106,11 @@ def local_all(t, *args, **kwargs):
 
 
 def allclose(
-    x: DNDarray, y: DNDarray, rtol: float = 1e-05, atol: float = 1e-08, equal_nan: bool = False
+    x: DNDarray,
+    y: DNDarray,
+    rtol: Optional[float] = None,
```
mrfh92, line 111: I think this needs to be discussed (e.g. in the PR talk): should we keep the NumPy/PyTorch-standard and choose some float limits as default (but adapted to Heats float32-standard datatype) or do we want to proceed differently and let tolerances by default be chosen automatically depending on the input datatypes?

heat/core/logical.py
```diff
@@ -139,17 +143,48 @@ def allclose(
...
+        dtype_precision = torch.finfo(t1.larray.dtype).bits
+    except TypeError:
+        dtype_precision = torch.iinfo(t1.larray.dtype).bits
+
+    adjustment_factor = 1.0
+
+    if dtype_precision != 64:
+        adjustment_factor = 64 / dtype_precision
+
+    if rtol is None:
```
mrfh92, line 161: If I understand this correctly, the default `atol` here is 1e-8 for float64 (as in NumPy). Surprisingly the same default `atol` is chosen in PyTorch as well, where the default datatype is float32, so I am not even sure whether maybe even the PyTorch default is a bit inconsistent. Maybe this needs to be discussed together with the issue mentioned above. 

Nevertheless, I think that `adjustment_factor` needs to be chosen in another way: as a rule of thumb we know that machine precision for float64 is roughly of size 1e-16 and 1e-8 for float32; therefore, I think that just doubling the tolerance when switching from float64 to float32 will not be sufficient.

mrfh92 (commented):
I think we should discussed the addressed issues (in principle only regarding the choice of some default values) in the PR talk next week. Apart from that your changes look fine 👍

heat/core/logical.py
Sai-Suraj-27: Sure...sir, I agree. I think `adjustment_factor` can be changed depending on the choice of some default values we consider. Please tell me after you have discussed this and come to a conclusion.

ClaudiaComito:
What's the status of this PR? I don't remember why it was assigned to me 🙈

ClaudiaComito:
Dear @Sai-Suraj-27 , after many internal discussions we have decided to not merge this PR and not address the original issue #889 . The background is this: `atol` and `rtol` would have to be set manually for many operations anyway, the numpy defaults aren't particularly related to 64-bit precision (otherwise they would be much lower), and the numpy defaults are identicaly to the torch defaults. 

`ht.allclose` will stick to the numpy defaults, deviations will have to be addressed in single test functions like we have been doing so far.

Thanks for your work @Sai-Suraj-27 , and for helping us closing that old issue.

Issue #1073, opened 2023-01-11
Add Hermitian to matrix gallery


ClaudiaComito:
**Related**
#390 #435 

**Feature functionality**
add a function  `utils.data.matrixgallery.hermitian` generating a Hermitian of arbitrary size, dtype, split for testing

**Additional context**
Testing of LA algorithms

mrfh92:
Corresponding PR is marked as ready for review. 

Reviewed within #1109

PR #1086, opened 2023-02-02
Add `hermitian()` to matrix gallery

swaptr:
## Description
Added `hermitian()` to generate random hermitian matrices of size `n`.

Issue/s resolved: #1073 

## Changes proposed:
- Add `hermitian()` to `utils.data.matrixgallery`.

## Type of change
- New feature (non-breaking change which adds functionality)

## Memory requirements
- NA

## Performance
- NA

## Due Diligence
- [ ] All split configurations tested
- [ ] Multiple dtypes tested in relevant functions
- [ ] Documentation updated (if needed)
- [ ] Title of PR is suitable for corresponding CHANGELOG entry

#### Does this change modify the behaviour of other functions? If so, which?
no

PR #1199, opened 2023-08-19
Add `hermitian()` to matrix gallery

ClaudiaComito:
## Due Diligence

- General:
- Implementation:

## Description

Creating a new PR to add tests for the Hermitian test matrix implemented in #1086.

Workflow: 
- merge #1086 into this PR ✅ 
- merge this PR into main

Issue/s resolved: #1073 

## Changes proposed:

-
-
-
-

no

### DIFF
--- a/heat/utils/data/matrixgallery.py
+++ b/heat/utils/data/matrixgallery.py
@@ -25,22 +25,22 @@ def hermitian(
     positive_definite: bool = False,
 ) -> DNDarray:
     """
-    Generates a random Hermitian matrix of size `(n,n)`. A Hermitian matrix is a complex square matrix that is equal to its conjugate transpose; for real data-types this routine 
-    returns a random symmetric matrix of size `(n,n)`. 
+    Generates a random Hermitian matrix of size `(n,n)`. A Hermitian matrix is a complex square matrix that is equal to its conjugate transpose; for real data-types this routine
+    returns a random symmetric matrix of size `(n,n)`.
 
     If `positive_definite=True`, the output is given by ::math`\frac{1}{n} R R^H` with ::math`R \in \mathbb{K}^{n\times n}` having entries distributed according to the standard normal distribution.
-    This corresponds to sampling a random matrix according to the so-called Wishart distribution; see, e.g., [2], and also [3] for additional information regarding the asymptotic distribution of 
-    the singular values. The output matrix will be positive definite with probability 1. 
-    
-    If `positive_definite=False`, the output is ::math`R+R^H` with ::math`R` generated as above. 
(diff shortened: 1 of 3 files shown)


heat/utils/data/matrixgallery.py
```diff
@@ -1,153 +1,186 @@
...
+
+def hermitian(
+    n: int,
+    split: Union[None, int] = None,
+    device: Union[None, str, Device] = None,
+    comm: Union[None, Communication] = None,
+    dtype: Type[datatype] = core.complex64,
+) -> DNDarray:
+    """
+    Generates a Hermitian matrix of size `(n,n)`. A Hermitian matrix is a  complex square matrix that is equal to its conjugate transpose.
```
mrfh92, line 27: With a small modification (devide by 1/size) we could add here that the matrices are sampled according to the "Marchenko-Pastur distribution" (https://en.wikipedia.org/wiki/Marchenko%E2%80%93Pastur_distribution)

heat/utils/data/matrixgallery.py
```diff
@@ -1,153 +1,186 @@
...
+        The desired data-type for the array, defaults to ht.complex64.
+
+    References
+    ----------
+    [1] https://en.wikipedia.org/wiki/Hermitian_matrix
+    """
+    real_dtype = core.float32 if dtype is core.complex64 else core.float64
+    matrix = randn(n, n, dtype=real_dtype, split=split, device=device)
+    matrix += 1j * randn(n, n, dtype=real_dtype, split=split, device=device)
+    return matrix @ core.conj(matrix).T
```
mrfh92, line 49: Two comments: 
* if we scale by 1/matrix.shape[0] the singular values are distributed according to Marchenko Pastur distribution (see above)... also this scaling might remove unwanted effects when choosing larger and larger matrices 
* The formula yields, of course, and they are even positive definite (with probability 1? - Im not an expert in this stuff...). However, doing the matrix-multiplication is quite expensive ... can we add something like an option "positive_definite" that triggers the formula used so far and use 
```
return matrix + core.conj(matrix).T.resplit(matrix.split) 
```
instead for the remaining cases?

mrfh92 (commented):
only small comments on docs and formula used for generation of the random matrices. 
But in prinicple, this PR is ready for merging even without these changes. 

Thanks for the work :+1:

heat/utils/data/matrixgallery.py
ClaudiaComito: Thanks, done

mrfh92:
@ClaudiaComito since I have done some changes here last week after my review it might be more useful if now you review again

mrfh92 (approved):
Looks fine to me, but since the last changes have been done by me, maybe somebody else should review too.

PR #1202, opened 2023-08-22
Separate data and computation in benchmarks

JuanPedroGHM:
## Description
Includes the changes done #1158, and includes updates to accommodate the new version of perun. Created a new branch and pull request because I messed up while rebasing.

* Removed data generation from the benchmarked functions.
* Updated perun api.
* Seed fixed in main script
* Device can be set with ENV variables

Issue/s resolved: #1157 

no

### DIFF
--- a/.perun.ini
+++ b/.perun.ini
@@ -3,6 +3,5 @@ format = bench
 data_out = ./bench_data
 
 [benchmarking]
-bench_enable = True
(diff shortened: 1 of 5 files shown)


ClaudiaComito (approved):
:clap: @JuanPedroGHM

Issue #1054, opened 2022-11-28
[Feature-384 Sparse] `to_sparse()` method for `DNDarray` class


Mystic-Slice:
**Related**
Pending work from https://github.com/helmholtz-analytics/heat/pull/1028

**Feature functionality**
Implement a `to_sparse()` method for the `DNDarray` class to create a `DCSR_matrix`.

Ishaan-Chandak:
Hi @Mystic-Slice I am interested to work on this issue.

Ishaan-Chandak:
Hi @Mystic-Slice in this we basically have to convert DNDarray to DCSR_Matrix

Mystic-Slice:
@Ishaan-Chandak Thanks for your interest!
Yes
We have `torch.Tensor`s inside a `DNDarray`. I think this method can simply just call `ht.sparse.sparse_csr_matrix` on the individual tensors(`dndarray.larray`) with the appropriate `is_split` parameter value.

Also, make sure that the `DNDarray` is balanced before doing this since the sparse module for now does not handle imbalance between processes.

When you make a PR for this, try filling out the form provided so that the PR holds complete information about your change/addition.

Ishaan-Chandak:
Hi @Mystic-Slice Thanks for the answer.
I have one more question that where should I add this function i.e. in which file.

Mystic-Slice:
It would make sense to add this in `heat/core/manipulations.py`

Ishaan-Chandak:
Hi @Mystic-Slice just an update, I  am reading the docs of heat for better understanding and i have managed to write some code. Currently I am not sure whether it is correct or not I will try to run that with some modifications and then send the final results here.

Ishaan-Chandak:
Hi @Mystic-Slice can you help me find some test cases to help me check whether the program I have written is correct or not.

Mystic-Slice:
Hey! @Ishaan-Chandak. Sorry for the late reply.
You can test it using something like 
```py
arr = [ [0, 0, 1, 0, 2]
        [0, 0, 0, 0, 0]
        [0, 3, 0, 0, 0]
        [4, 0, 0, 5, 0]
        [0, 0, 0, 0, 6] ]

A = ht.array(arr, split=0) # DNDarray

B = ht.array(arr, split=0).to_sparse() # DCSR_matrix

assert (A.larray.to_sparse_csr() == B.larray)
```

I of course have not tested this code, but you must be able to verify the `to_sparse()` method this way.

[...]

PR #1206, opened 2023-08-27
Implement `to_sparse()` method for DNDarray conversion to DCSR_matrix

samadpls:
worked on [Feature-384 Sparse] to_sparse() method for DNDarray class 
## Due Diligence

- General:
- Implementation:

## Description
Implemented `to_sparse()` method for DNDarray class conversion to DCSR_matrix representation

Issue/s resolved: #1054 

## Changes proposed:

- Added the to_sparse() method to the DNDarray class for converting to a DCSR_matrix representation.

## Type of change

## Memory requirements

Memory requirements for this change have not been profiled at this time.

## Performance

Performance metrics for this change have not been measured at this time.

no

### DIFF
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -9,6 +9,9 @@ import warnings
 
 from typing import Iterable, Type, List, Callable, Union, Tuple, Sequence, Optional
 
+from heat.sparse.dcsr_matrix import DCSR_matrix
+from heat.sparse.factories import sparse_csr_matrix
+
 from .communication import MPI
 from .dndarray import DNDarray
 
@@ -59,8 +62,43 @@ __all__ = [
     "unique",
     "vsplit",
     "vstack",
+    "to_sparse",
 ]
 
+
+def to_sparse(array: DNDarray) -> DCSR_matrix:
+    """
+    Convert the distributed array to a sparse DCSR_matrix representation.
+
+    Parameters
+    ----------
+    array : DNDarray
+        The distributed array to be converted to a sparse DCSR_matrix.
+
+    Returns
+    -------
+    DCSR_matrix
+        A sparse DCSR_matrix representation of the input DNDarray.
+
+    Notes
+    -----
+    This method allows for the conversion of a DNDarray into a sparse DCSR_matrix representation,
+    which is useful for handling large and sparse datasets efficiently.
+
+    Examples
+    --------
+    >>> dense_array = ht.array([[1, 0, 0], [0, 0, 2], [0, 3, 0]])
+    >>> sparse_matrix = dense_array.to_sparse()
+
+    """
+    array.balance_()
+    csr_matrices = [sparse_csr_matrix(tensor, is_split=array.split) for tensor in array.larray]
+    result = DCSR_matrix.from_csr_matrices(csr_matrices)
+
+    return result
+
+DNDarray.to_sparse: Callable[[DNDarray], DCSR_matrix] = lambda self: to_sparse(self)
+DNDarray.to_sparse.__doc__ = to_sparse.__doc__
 
 def balance(array: DNDarray, copy=False) -> DNDarray:
     """
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -61,6 +61,7 @@ __all__ = [
     "vstack",
 ]
 
+
 def balance(array: DNDarray, copy=False) -> DNDarray:
     """
     Out of place balance function. More information on the meaning of balance can be found in
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -9,9 +9,6 @@ import warnings
 
 from typing import Iterable, Type, List, Callable, Union, Tuple, Sequence, Optional
 
-from heat.sparse.dcsr_matrix import DCSR_matrix
-from heat.sparse.factories import sparse_csr_matrix
-
 from .communication import MPI
 from .dndarray import DNDarray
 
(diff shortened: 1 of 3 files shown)


samadpls:
I will work on writing  test cases once the implementation is approved

heat/core/manipulations.py
```diff
@@ -59,9 +62,46 @@
...
+
+    Examples
+    --------
+    >>> dense_array = ht.array([[1, 0, 0], [0, 0, 2], [0, 3, 0]])
+    >>> sparse_matrix = dense_array.to_sparse()
+
+    """
+    array.balance_()
+    csr_matrices = [sparse_csr_matrix(tensor, is_split=array.split) for tensor in array.larray]
+    result = DCSR_matrix.from_csr_matrices(csr_matrices)
```
Mystic-Slice, line 96: Hi @samadpls!
You are in the right direction in that you have tried to preserve the split configuration using the `is_split` parameter. Kudos for remembering to balance the array first.

However, within a process, the whole data chunk must be treated as a single sparse tensor. So, this won't work,
```py
csr_matrices = [sparse_csr_matrix(tensor, is_split=array.split) for tensor in array.larray]
```
Also, `DCSR_matrix.from_csr_matrices` does not exist. If it is something that you intend to implement, I don't think there is a need for it.

After you balance the array, you can just use the existing factory method `factories.sparse_csr_matrix` with the correct split configuration. I think this would work for us.

Mystic-Slice (changes requested):

Mystic-Slice:
About your issue with imports -> https://github.com/helmholtz-analytics/heat/issues/1054#issuecomment-1696261876

I think you should first move this function to `heat/sparse/manipulations.py` and check if the errors still persist. If you still get some errors, please post screenshots of the error messages.

If you have any other doubts, please feel free to ask here.

ClaudiaComito:
@samadpls and @Mystic-Slice thank you both very much for your contribution. 

Please do implement at least some basic tests before asking for review @samadpls, it makes the review process so much easier - including for you.

Thanks!

samadpls:
Hey @Mystic-Slice and @ClaudiaComito,
Thanks for the reviews. Sorry for the delay. Code is ready for review. Added unit tests, but had to modify them as direct tensor comparison wasn't working with SparseCsr.

heat/sparse/manipulations.py
```diff
@@ -2,16 +2,51 @@
...
+    which is useful for handling large and sparse datasets efficiently.
+
+    Examples
+    --------
+    >>> dense_array = ht.array([[1, 0, 0], [0, 0, 2], [0, 3, 0]])
+    >>> sparse_matrix = dense_array.to_sparse()
+
+    """
+    array.balance_()
+    result = sparse_csr_matrix(array.larray, is_split=array.split)
```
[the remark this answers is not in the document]
samadpls: Hey @Mystic-Slice, I'm not sure but calling sparse_csr_matrix with only an array is giving the following error.
```
  File "/media/samadpls/ubuntu/ML/heat/test.py", line 7, in <module>
    B = A.to_sparse()
  File "/media/samadpls/ubuntu/ML/heat/heat/sparse/manipulations.py", line 42, in to_sparse
    result = sparse_csr_matrix(array, dtype=array.dtype, device=array.device, comm=array.comm, is_split=array.split)
  File "/media/samadpls/ubuntu/ML/heat/heat/sparse/factories.py", line 120, in sparse_csr_matrix
    obj = torch.tensor(
  File "/media/samadpls/ubuntu/ML/heat/heat/core/dndarray.py", line 1116, in __len__
    return self.shape[0]
IndexError: tuple index out of range
```
I checked other test cases and found that sparse_csr_matrix is called with `torch.Tensor` rather than DNDarray, as in this case.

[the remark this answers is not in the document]
Mystic-Slice: Alright. Let's just use `array.larray` then.

samadpls:
Done, @Mystic-Slice,  let me know if there are any further adjustments needed.

Mystic-Slice (approved):
LGTM!

Mystic-Slice:
@samadpls Nice work!

@ClaudiaComito This PR is good to be merged.

ClaudiaComito (changes requested):
Very nice job @samadpls . @Mystic-Slice I'm thinking it would be better to have consistent naming of the functions here (`to_dense`, `to_sparse` ) and alias `to_dense` to `todense` for scipy consistency. What do you both think?

samadpls:
Just to confirm we should also create a `tosparse()` alias, similar to `todense()`, for consistency?

heat/sparse/manipulations.py
```diff
@@ -2,16 +2,53 @@
...
-
+from heat.sparse.factories import sparse_csr_matrix
 from ..core.memory import sanitize_memory_layout
 from ..core.dndarray import DNDarray
 from ..core.factories import empty
 
 __all__ = [
     "todense",
+    "to_sparse",
 ]
```
ClaudiaComito, line 13: @samadpls just to clarify, I think we should name these functions consistently (see below). The scipy `sparse.crs_matrix` has a method `todense()` so we should offer that as an alias. 

Do numpy ndarrays  have a `tosparse()` method? I don't think so but I might be wrong.  

```suggestion
__all__ = [
    "to_dense",
    "to_sparse",
]
```

samadpls: Thanks for the clarification! I've updated the code to keep the naming consistent. 😋

ClaudiaComito (approved):
Perfect, thanks @samadpls !

Mystic-Slice:
> Very nice job @samadpls . @Mystic-Slice I'm thinking it would be better to have consistent naming of the functions here (`to_dense`, `to_sparse` ) and alias `to_dense` to `todense` for scipy consistency. What do you both think? 
> 
> 

Yup. Sounds good!

Issue #1213, opened 2023-09-04
CB Corrections


JuanPedroGHM:
## Problems
- Workflows from PR's not triggered correctly
- Workflows triggered multiple times
- Results are hard to find

PR #1214, opened 2023-09-05
Improve single benchmark execution

JuanPedroGHM:
## Due Diligence
Aiming to improve reporting of CB benchmark results. They are hard to find and unreliable.

## Description
Changes to the two main action files, trigger benchmarks, and report benchmarks. (With changes on codebase side as well)

Issue/s resolved: #1213


ClaudiaComito (approved):
Thanks a lot @JuanPedroGHM !

Issue #1064, opened 2022-12-23
Avoid unnecessary `gather`ing of distributed operand in mixed distributed/non-distributed `logical` functions


ClaudiaComito:
**Feature functionality**

module `heat.core.logical`, functions `allclose(x,y)`, `isclose(x,y)` 

Currently, when only one of the operands is distributed, these functions (via `__sanitize_close_input`) [gather](https://mpitutorial.com/tutorials/mpi-scatter-gather-and-allgather/) the distributed operand and perform the global comparison on each process. 

Modify the relevant functions to carry out the comparison on the relevant slice of non-distributed operand instead.

Ishaan-Chandak:
Hi @ClaudiaComito  is this issue still open ?

mrfh92:
This issue is stil open and useful. 

(Reviewed within #1109)

samadpls:
Hey @ClaudiaComito and @mrfh92, can I work on this issue?

PR #1216, opened 2023-09-08
Enhancement: Avoid unnecessary gathering of distributed operand

samadpls:
## Due Diligence

- General:
- Implementation:

## Description

Updated `__sanitize_close_input` logic to resolve issue #1064 regarding unnecessary gathering of distributed operands.

Issue/s resolved: #1064

## Changes proposed:

- optimized the logic in the `__sanitize_close_input` function to avoid unnecessary gathering of distributed operands.

## Type of change

## Memory requirements

Memory requirements for this change have not been profiled at this time.

## Performance

Performance metrics for this change have not been measured at this time.
yes


heat/core/logical.py
```diff
@@ -516,17 +516,14 @@ def sanitize_input_type(
...
-    elif x.split != y.split:
-        t2 = manipulations.resplit(y, axis=x.split)
-        return x, t2
+    #  If y is distributed, update t1 and t2 accordingly
+    if y.split is not None:
+        t1, t2 = y, x
 
-    else:
-        return x, y
+    return t1, t2
```
mtar, line 526: The output should be distributed along the same split axis, but now the inputs are simply returned. This is not what we want here.

mtar (commented):
Thank you @samadpls. However, your contribution is contrary to the intended behavior of the function and does not address the issue. Perhaps it wasn't written clearly and concisely enough, so I rephrase it again:

What we want for `allclose`/`isclose`: If we have a distributed array and a non-distributed array as inputs, slice the undistributed array such that we can compare it with the corresponding local part of the distributed array on the process.

Hope this helps

heat/core/logical.py
```diff
@@ -519,6 +519,9 @@ def sanitize_input_type(
     # if one of the tensors is distributed, unsplit/gather it
     if x.split is not None and y.split is None:
         t1 = manipulations.resplit(x, axis=None)
+        local_start = x.comm.chunk(t1.gshape, x.split)[0]
```
mtar, line 522: The issue wants to avoid the creation of **t1**

mtar, line 522: x.comm.chunk() gives the offsets for a balanced DNDarray on all processes, but this is not always the case. Calculating the offset from x.lshape_map may give more accurate values of the current distribution.

heat/core/logical.py
```diff
@@ -519,6 +519,9 @@ def sanitize_input_type(
...
+        local_end = local_start + x.lshape[x.split]
+        y = y[local_start:local_end]
```
mtar, line 524: When working with DNDarrays, the arguments to a function should be the same on all processes. For local operations on each process, use the methods _larray()_ or _lloc()_ to work with the underlying PyTorch tensors.

mtar, line 524: Functions working with DNDarrays should always get the same arguments on all processes. For local operations, we use the underlying PyTorch tensor accessible through _larray_ or to a lesser extent the _lloc_ methods.

mtar (commented):
This is a small step into the right direction :slightly_smiling_face:

ClaudiaComito (changes requested):
@samadpls here's what we're trying to achieve with this fix. 

First of all, all Heat operations are implemented to operate on memory-distributed arrays (DNDarrays) . In practice:
- operations are performed on many MPI processes, could be many CPUs or many GPUs, many nodes of a supercomputer;
- if a DNDarray is distributed, each MPI process stores only a slice of that DNDarray (along the `DNDarray.split` dimension)
- users can choose whether to distribute which DNDarrays or not (`DNDarray.split = None` means each process has a copy of the entire DNDarray in memory).

In this PR, you are addressing the case in which  logical operations `all`, `allclose` etc. compare a distributed DNDarray to a non-distributed one.

Example with small arrays for better understanding:

```python
import numpy as np
import heat as ht

x = ht.arange(30, split=0).reshape(-1, 2)  
y = ht.arange(30).reshape(-1, 2)

print(ht.all(x == y))
```
This will return `True` no matter how many MPI processes you run it on. 

Try and print out the local arrays (`DNDarray.larray`) to get a better understanding what the underlying data on each process really are (run this code on more than one process, see [here](https://github.com/helmholtz-analytics/heat#getting-started)):

```python
rank = x.comm.rank 
print(f"On rank {rank}: local x data: {x.larray}")
print(f"On rank {rank}: local y data: {y.larray}")
```

In the current implementation, when a distributed `x` is compared to non-distributed `y`, `x` gets "gathered" so it is no longer distributed. But "gathering" all the slices of a distributed array onto each process is communication- and memory-intensive and, in this case, unnecessary. Each process can simply compare the local slice of x (`x.larray`) to the corresponding slice of `y` and, at the end, only communicate whether the slices on each process match or not.

Feel free to ask if anything is unclear.

heat/core/logical.py
```diff
@@ -519,6 +519,9 @@ def sanitize_input_type(
     # if one of the tensors is distributed, unsplit/gather it
     if x.split is not None and y.split is None:
         t1 = manipulations.resplit(x, axis=None)
```
ClaudiaComito, line 521: @samadpls `resplit(x, axis=None)` collects all distributed slices of `x` and creates a copy of the entire `x` on each process, this is what we want to avoid.

samadpls:
I apologize for the delay in responding. I've been on vacation and traveling for the past 22 days, and I just landed a few hours ago. I'll definitely get back to work and address the PR comments. Thanks for your understanding!

ClaudiaComito:
> I apologize for the delay in responding. I've been on vacation and traveling for the past 22 days, and I just landed few hours ago. I'll definitely get back to work and address the PR comments. Thanks for your understanding!

No problem at all @samadpls, thanks for your time!

ClaudiaComito:
@samadpls have you installed the pre-commit hooks as described [here](https://github.com/helmholtz-analytics/heat/blob/main/quick_start.md#new-contributors)?

samadpls:
@ClaudiaComito Sorry for the oversight! 😢  I've now installed the pre-commit hooks. Thanks for bringing it to my attention

heat/core/logical.py
```diff
@@ -516,10 +516,13 @@ def sanitize_input_type(
...
     y = sanitize_input_type(y, x)
 
-    # if one of the tensors is distributed, unsplit/gather it
-    if x.split is not None and y.split is None:
-        t1 = manipulations.resplit(x, axis=None)
-        return t1, y
+    # if one of the tensors is distributed and the other is not a scalar, unsplit/gather it
+    if x.split is not None and y.split is None and y.ndim > 0:
+        local_start = sum(x.lshape[: x.split])
+        local_end = local_start + x.lshape[x.split]
```
ClaudiaComito, line 522: The `lshape` attribute gives you information on the shape of the process-local data. `lshape` on a given process doesn't know about the size of slices of `x` distributed on other processes. 

Because you need to slice `y` along the distribution axis `x.split`, what you need here is information on how `x` is distributed across processes. I can think of 2 options:

**Option a):** use the information in `x.lshape_map`: 

`lshape_map` is 2-dimensional with shape (p, ndim) where p is the number of processes, ndim the number of dimension of a DNDarray. 
Example, say you have x.shape = (10, 5, 8) (this is the global shape) and x.split = 0, and say you're distributing across 3 MPI processes.  `x.lshape_map` might look like this
```
tensor([[4, 5, 8],
        [3, 5, 8],
        [3, 5, 8]])
```
The distribution along the first column (split axis = 0) might be different if `x` is not balanced.

**Option b):** use the `dndarray.counts_displs()` method which returns tuples of the counts (elements of `x` along the split axis for every process) and displacements (starting index of the local slice as referred to the global array). For the example above (shape (10, 5, 8) on 3 processes), `x.counts_displs()` will return 
```shell
>>> counts
(4, 3, 3)
>>> displs
(0,4,3)
```

samadpls:
@ClaudiaComito , Firstly, I want to Thank you for your ongoing support on this PR. I apologize for the delay. Struggling with parallel test execution due to my limited MPI experience. I've tried various solutions, but issues persist. Think I might be missing something. If you have any advice or suggestions, I'd really appreciate it.

ClaudiaComito (changes requested):
@samadpls thanks for bearing with us. I suggested a change which should make everything much easier

heat/core/logical.py
```diff
@@ -516,10 +517,11 @@ def sanitize_input_type(
     x = sanitize_input_type(x, y)
     y = sanitize_input_type(y, x)
 
-    # if one of the tensors is distributed, unsplit/gather it
-    if x.split is not None and y.split is None:
-        t1 = manipulations.resplit(x, axis=None)
-        return t1, y
+    # if one of the DNDarrays is distributed and the other is not
+    if x.split is not None and y.split is None and y.ndim > 0:
```
ClaudiaComito, line 521:
```suggestion
    if x.is_distributed() and not y.is_distributed() and y.ndim > 0:
```
if `split is not None` but users are running things on 1 MPI process, the array is not distributed and no need for the extra steps

heat/core/logical.py
```diff
@@ -516,10 +517,11 @@ def sanitize_input_type(
...
+        t2 = factories.array(y.larray, device=x.device, split=x.split)
+        x, t2 = sanitation.sanitize_distribution(x, t2, target=x)
+        return x, t2
```
ClaudiaComito, line 525: We're missing the same check for y distributed, x non distributed I think. Otherwise we're good to go.

ClaudiaComito (changes requested):
Thanks @samadpls , two more small changes and we're good to go.

ClaudiaComito (approved):
We made it, thanks a lot @samadpls !

samadpls:
Thank you so much @ClaudiaComito, I couldn't have done it without your assistance.

ClaudiaComito:
> Thank you so much @ClaudiaComito, I couldn't have done it without your assistance.

My pleasure.

The CI is failing, but this has nothing to do with your PR, we've been working on it in #1313 and #1314 . There's nothing you can do about it, #1314 needs to be merged before everything else.  

We're about to start the Christmas break so things will be grinding to a halt until early January. Happy holidays! 🎄

ClaudiaComito (approved):

PR #1218, opened 2023-09-18
Distributed Fast Fourier Transforms

### LINKED ISSUE #1097: Provide Fast Fourier Transform (FFT)
**Feature functionality**
Provide routines for Fast Fourier Transform (FFT) and its inverse for $n$-dimensional `DNDarray`'s. Fourier transform along a single axis, a prescribed subset of axes, or along all axes should be possible. This would correspond roughly to the functionality of the `fft`-module in PyTorch: 

* `fftn`/`ifftn`: FFT/iFFT of an $n$-dim array along up to $n$ axes 
* `fft`/`ifft`: FFT/iFFT of an $n$-dim array along a single axis only 
* `fft2`/`ifft2`: FFT/iFFT of an $n$-dim array along two axes only 
* the same with `r...`: modifications in the case of real-valued input arrays 
* the same with `h...`: modifications in the case of hermitian input arrays 
* if time permits: the "helper functions"

Can potentially be used for convolutions (#1248 )

**Material**
- [arXiv:1804.09536](https://arxiv.org/abs/1804.09536) may serve as an overview on $n$-dimensional FFT. Heat's data distribution is referred to as "slab decomposition", so actually we will do something much more easy than described in this paper. 
- There is a [fft-module in PyTorch](https://pytorch.org/docs/stable/fft.html) that we can use for process-local computations. 

**Suggestions and Hints** 
The discrete Fourier transform $\mathcal{F}$ of an $n$-dimensional array is given by $\mathcal{F}=\mathcal{F}_{0}\circ... \circ\mathcal{F}_{d-1}$ where $\mathcal{F}_i$ denotes the 1-dimensional (i.e. usual) DFT along the $i$-th axis. Consequently, a similar factorization holds true for the inverse DFT as well. 

The "only" tricky part when adding parallelization to this concept is the Fourier transformation along the split-axis. I would like to recommend something like the following approach: 

1. do locally on each process: FFT along all axes $i$ such that $i \in dim$ and $i\neq split$ (use the respective PyTorch function). If $split \notin dim$ we are done. 
2. If $split \in dim$ now resplit the array to a different split axis. 
3. do locally on each process: FFT (1-dimensional, PyTorch) along axis $split$ (the original split axis, of course, which is currently no more the split axis...) 
4. resplit back to obtain same split-axis as at the beginning 

Similarly, also the inverse FFT (iFFT) can be implemented. Of course, this is just a very first idea, how FFT and iFFT could be implemented in Heat; if you have a good idea for a more elaborate, more efficient implementation, feel free to tell me :) 
Since PyTorch-FFT already supports CPU and GPU, the corresponding implementation in Heat should work on CPU und GPU without particular additional effort.

ClaudiaComito:
Resolves #1097  and can be used for #1248 

## Due Diligence

- General:
- Implementation:

## Description

Introducing `fft` module with all relevant operations.

Main points:
- Transforms along the split dimension are handled by redistributing to a different axis first, transforming the original split dimension locally, and then redistributing back.
- Multi-axis transforms on real input (`rfftn`, `ihfft` etc.) will be performed as the corresponding generic FFT operation  if transform of the split axis is required. This is due to the intrinsic recursion built into these operations (transpose, resplit, transform, resplit, transpose, transform), where the complex output of the first real transform cannot act as input to the second real transform.
- Multi-axis transforms on real input  still return the correct shape, i.e. for the last transformed axis only up to the Nyquist frequency.

Still TODO:

- [x] expand documentation esp. of real ffts
- [x] implement `fftfreq`, `fftshift` etc.
- [x] fix test failure (on 7 processes) EDIT: solved by setting correct output dtype of empty local tensor)

Issue/s resolved: #1097 

## Changes proposed:

- implemented FFT functions
- implemented tests
-
-

## Type of change

no


heat/fft/fft.py
```diff
@@ -0,0 +1,787 @@
...
+
+
+def fftn(
+    x: DNDarray, s: Tuple[int, ...] = None, axes: Tuple[int, ...] = None, norm: str = None
+) -> DNDarray:
+    """
+    Compute the N-dimensional discrete Fourier Transform.
+
+    This function computes the N-dimensional discrete Fourier Transform over any number of axes in an M-dimensional
+    array by means of the Fast Fourier Transform (FFT). By default, all axes are transformed, with the real transform
```
mrfh92, line 353: The comment on real transform along one axis and complex transform on the others does not appear in PyTorch. Is there a difference of the conventions of PyTorch and NumPy?

heat/fft/fft.py
```diff
@@ -0,0 +1,787 @@
...
+    axis : int, optional
+        Axis over which to compute the FFT. If not given, the last axis is used, or the only axis if x has only one
+        dimension. Default: -1.
+    norm : str, optional
+        Normalization mode: 'forward', 'backward', or 'ortho' (see `numpy.fft` for details). Default is "backward".
+
+    Notes
+    -----
+    This function requires MPI communication if the input array is transformed along the distribution axis.
+    If the input array is 1-D and distributed, this function copies the entire array on each MPI process! i.e. if the array is very large, you might run out of memory.
```
mrfh92, line 310: From my point of view, this is OK. However, I would recommend to open a corresponding issue to work on an alternative specifically for this distributed 1d-case (or does this not make any sense?)

heat/fft/fft.py
```diff
@@ -0,0 +1,787 @@
...
+    x : DNDarray
+        Input array, can be complex
+    s : Tuple[int, ...], optional
+        Shape of the output along the transformed axes. (default is x.shape)
+    axes : Tuple[int, ...], optional
+        Axes over which to compute the inverse FFT. If not given, the last `len(s)` axes are used, or all axes if `s` is
+        also not specified. Repeated indices in `axes` means that the transform over that axis is performed multiple
+        times. (default is None)
+    norm : str, optional
+        Normalization mode: 'forward', 'backward', or 'ortho' (see `numpy.fft` for details). Default is "backward".
```
mrfh92, line 542: One could also add a reference to the PyTorch-docs here, because they use the same options for `norm`

heat/fft/tests/test_fft.py
```diff
@@ -0,0 +1,235 @@
...
+        # FFT along last axis
+        n = 5
+        y = ht.fft.fft(x, n=n)
+        np_y = np.fft.fft(x.numpy(), n=n)
+        self.assertIsInstance(y, ht.DNDarray)
+        self.assertEqual(y.shape, np_y.shape)
+        self.assertTrue(y.split == 0)
+        self.assert_array_equal(y, np_y)
+
+        # on GPU, test only on less than 4 processes
```
mrfh92, line 41: Is there a reason for not testing on 8 GPUs?

heat/fft/fft.py
```diff
@@ -0,0 +1,787 @@
...
+    "ihfft",
+    "ihfft2",
+    "ihfftn",
+    "irfft",
+    "irfft2",
+    "irfftn",
+    "rfft",
+    "rfft2",
+    "rfftn",
+    # "fftfreq",
```
mrfh92, line 32: If there is no special reason for omitting these functions, I would suggest to open a separate issue to add them lateron.

heat/fft/fft.py
```diff
@@ -0,0 +1,787 @@
+"""Provides a collection of Discrete Fast Fourier Transforms (DFFT) and their inverses."""
+
+import torch
```
mrfh92, line 3: I saw the hint "Supports torch.half and torch.chalf on CUDA with GPU Architecture SM53 or greater. However it only supports powers of 2 signal length in every transformed dimensions." in the PyTorch docs on fft/fftn ... does this have any implications for us? Or is this such an exotic case that we dont cover anyway...

heat/fft/fft.py
```diff
@@ -0,0 +1,787 @@
...
+        torch.fft.rfft: torch.fft.fft,
+        torch.fft.ihfft: torch.fft.ifft,
+    }
+    real_op = fft_op in real_to_generic_fft_ops
+    if real_op:
+        nyquist_freq = n // 2 + 1
+        output_shape[axis] = nyquist_freq
+    else:
+        output_shape[axis] = n
+    print("DEBUGGING: n, output_shape = ", n, output_shape)
```
mrfh92, line 75: here, and at some other places debuggin-info gets printed; we should remove that from the code before merging

heat/fft/tests/test_fft.py
```diff
@@ -0,0 +1,235 @@
...
+        # FFT along last 2 axes
+        y = ht.fft.rfftn(x, axes=(1, 2))
+        np_y = np.fft.rfftn(x.numpy(), axes=(1, 2))
+        self.assertIsInstance(y, ht.DNDarray)
+        self.assertEqual(y.shape, np_y.shape)
+        self.assertTrue(y.split == 0)
+        self.assert_array_equal(y, np_y)
+        backwards = ht.fft.irfftn(y, s=x.shape[-2:])
+        self.assertTrue(ht.allclose(backwards, x))
+        # FFT along all axes
```
mrfh92, line 211: I have added a comment to that PR in order that we do not forget this

mrfh92 (commented):
From my point of view this looks fine. Thanks for the work :+1: 
I only have added some minor questions/comment that we can address in the PR talk; after that I guess we can merge. 

In addition, we could already add a benchmark for FFT if this makes sense.

heat/fft/fft.py
ClaudiaComito: Interesting, this is GitHub copilot extrapolating out of context. The comment on the real transform belongs in `rfftn`. I've modified the docs.

heat/fft/tests/test_fft.py
ClaudiaComito: I found what caused the problem on more than 3 GPUs and removed this constraint.

heat/fft/fft.py
ClaudiaComito: these are implemented now

heat/fft/fft.py
ClaudiaComito: Let's keep this in mind when we tackle #94

heat/fft/fft.py
ClaudiaComito: You're right, removed

heat/fft/fft.py
ClaudiaComito: Created issue #1263

heat/fft/fft.py
ClaudiaComito: I've removed the pointer to [numpy.fft](https://numpy.org/doc/stable/reference/routines.fft.html) for now, however that page does have a lot more infos than the torch.fft page. I think it would be good to mention the numpy page in the docs because it's so exhaustive.

ClaudiaComito:
> From my point of view this looks fine. Thanks for the work 👍 I only have added some minor questions/comment that we can address in the PR talk; after that I guess we can merge.
> 
> In addition, we could already add a benchmark for FFT if this makes sense.

@mrfh92 thanks a lot for the review. I think I've addressed all of your comments. I will add the benchmarks next week.

mrfh92 (approved):
As far as I can see, everything is fine now. 
Thanks for your work! :+1:

mrfh92:
It looks like that we have the same problem with the old PyTorch versions as in #933

Issue #1219, opened 2023-09-18
[Bug]: Benchmarks issues


ClaudiaComito:
### What happened?

The new incarnation of the benchmarks #1214 seems to have introduced some problems:

1. the benchmarks run on every PR, not just on those labeled `benchmark PR`
2. benchmarks run very long / appear to be hanging (see #1204 #1206 )

Thanks for looking into this @JuanPedroGHM !

### Version

1.3.x

### Python version

None

### PyTorch version

None

PR #1220, opened 2023-09-18
Trigger Benchmark Corrections

JuanPedroGHM:
## Description

Misconceptions about some github variables from the pull_request and push events lead to triggering the benchmarks with wrong values. This aims to correct it, while also avoiding to create an endless check that waits for the benchmark to finish.

Issue/s resolved: #1219 

## Changes proposed:
- Small corrections to trigger benchmark action

## Type of change
- Changes to CB actions.

### DIFF
--- a/.github/workflows/bench_report.yml
+++ b/.github/workflows/bench_report.yml
@@ -29,7 +29,7 @@ jobs:
           --header "PRIVATE-TOKEN: ${{ secrets.GITLAB_CB_API_TOKEN }}" \
           --output report.txt \
           "https://codebase.helmholtz.cloud/api/v4/projects/7930/jobs/${{ inputs.job_id }}/artifacts/heat/bench_data/report.txt"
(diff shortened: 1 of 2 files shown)


mrfh92:
@JuanPedroGHM is there an update on this issue? It seems that the benchmark action is still running endless (?)

JuanPedroGHM:
> @JuanPedroGHM is there an update on this issue? It seems that the benchmark action is still running endless (?)

It should be ready for merging. The check that run's endlessly is the outdated action, it is not waiting for anything really. It will be removed after the merge to main. The actual checkmark in this case is ** cb/report**. When you click on the details, it will take you to the benchmark data. This one failed because the CPU utilization is too high, it seems that the runners in gitlab are shared by all the users.

mrfh92 (approved):
As far as I can judge, this looks fine :+1: 
Thanks for the work

PR #1222, opened 2023-09-28
replaced version numbers from docker docs with placeholders

JuanPedroGHM:
## Description

Removed specific version numbers and image tags from docker documentation, to avoid constant changes with each version.


ClaudiaComito (approved):
Thank you so much @JuanPedroGHM ! Looks good to me

Issue #1184, opened 2023-07-27
in-place operators for `arithmetics.py`


mtar:
**Related**
The memory/time may be saved by using in-place functions.

**Feature functionality**
Implement in-place functions for some local operations.

**Additional context**
https://pytorch.org/docs/stable/generated/torch.Tensor.add_.html?highlight=add_#torch.Tensor.add_
https://docs.python.org/3/reference/datamodel.html#object.__iadd__

ClaudiaComito:
What in-place operations are we missing? Most if not all ops in `arithmetics` have already an in-place version.

PR #1228, opened 2023-10-06
Introduce in-place-operators for `arithmetics.py`

LScheib:
## Due Diligence

- General:
- Implementation:

## Description

Issue/s resolved: #1184 

## Changes proposed:

New in-place operators (in `heat/core/arithmetics.py`):
`add_`, `sub_`/`subtract_`, `mul_`, `div_`/`divide_`, `bitwise_and_`, `bitwise_or_`, `bitwise_xor_`, `floordiv_`/`floor_divide_`, `remainder_`/`mod_`, `neg_`/`negative_`, `copysign_`, `cumprod_`/`cumproduct_`, `cumsum_`, `gcd_`, `hypot_`, `invert_`/`bitwise_not_`, `lcm_`, `nan_to_num`, `pow_`/`power_`, `fmod_`, `left_shift_`, `right_shift_`.

Corresponding test functions (in `heat/core/tests/test_arithmetics.py`).

The functions `diff`, `divmod`, `nanprod`, `nansum`, `prod`, and `sum` do not operate in a way which allows for in-place versions. `pos`/`positive` is an element-wise function which could work in-place, but there are no corresponding in-place functions in PyTorch or NumPy.

yes / no


LScheib:
Question for PR talk: How can we test whether the newly defined in-place operators (e.g. my function iadd in arithmetics.py) actually operate in-place? Can we establish such a test in test_arithmetics.py?

heat/core/tests/test_arithmetics.py
```diff
@@ -503,6 +503,101 @@ def test_hypot(self):
...
+        self.assertTrue(ht.equal(self.a_tensor, result))
+        self.assertTrue(torch.equal(self.a_tensor.larray, result.larray))
+        self.assertTrue(ht.equal(self.a_scalar, ht.float32(2.0)))
+        self.a_tensor = ht.array([[1.0, 2.0], [3.0, 4.0]])
+
+        self.assertTrue(ht.equal(ht.iadd(self.a_tensor, self.another_tensor), result))
+        self.assertTrue(ht.equal(self.a_tensor, result))
+        self.assertTrue(torch.equal(self.a_tensor.larray, result.larray))
+        self.assertTrue(ht.equal(self.another_tensor, ht.array([[2.0, 2.0], [2.0, 2.0]])))
+        self.a_tensor = ht.array([[1.0, 2.0], [3.0, 4.0]])
```
mrfh92, line 523: I guess this results in an error due to the definition 
```
cls.a_vector = ht.float32([2, 2])
cls.another_vector = ht.float32([2, 2, 2])
```

mrfh92:
I have added a comment to the place where I think the error comes from (see "Files changed")

heat/preprocessing/preprocessing.py
```diff
@@ -129,10 +129,10 @@ def transform(self, X: ht.DNDarray) -> ht.DNDarray:
         _has_n_features(self.mean_, X)
         if self.copy:
             return (X - self.mean_) * self.scale_
-        # else in place:
```
mrfh92, line 132: The original version was correct due to the return in the if-clause (same at further places here)

LScheib: Got it, thank you.

mrfh92:
@LScheib I have replaced `__idiv__` by `__itruediv__` since `__div__` etc. seem to have been deprecated in favour of `__truediv__` etc.

mrfh92:
Except for `RobustScaler`, all preprocessing routines allow for calls like `scaler.transform(X)` when in in-place mode (`copy=False`). Only for `RobustScaler` this was not possible (why???) and I had to keep `X=scaler.transform(X)` etc. there.

heat/naive_bayes/gaussianNB.py
```diff
@@ -390,8 +390,9 @@ def __joint_log_likelihood(self, x: DNDarray) -> DNDarray:
         joint_log_likelihood = ht.empty(jll_shape, dtype=x.dtype, split=x.split, device=x.device)
         for i in range(jll_size):
             jointi = ht.log(self.class_prior_[i])
-            n_ij = -0.5 * ht.sum(ht.log(2.0 * ht.pi * self.sigma_[i, :]))
-            n_ij -= 0.5 * ht.sum(((x - self.theta_[i, :]) ** 2) / (self.sigma_[i, :]), 1)
+            n_ij = -0.5 * ht.sum(ht.log(2.0 * ht.pi * self.sigma_[i, :])) - 0.5 * ht.sum(
```
mrfh92, line 393: Here, we should find out why `-=` produced an error (?)

heat/preprocessing/tests/test_preprocessing.py
```diff
@@ -172,75 +238,110 @@ def test_MaxAbsScaler(self):
...
+                                    ht.ones(X.shape[1])[:-2],
+                                    atol=atol_fit,
                                 )
-                            Z = ht.zeros(
-                                (MPI.COMM_WORLD.Get_size() * 10, MPI.COMM_WORLD.Get_size() * 2),
-                                split=split,
                             )
-                            with self.assertRaises(ValueError):
-                                scaler.transform(Z)
+                        Y = scaler.inverse_transform(X)
```
mrfh92, line 343: Here, we also should find out why the robust scaler could not be realized in the same way as the other scalers for the in-place variant.

mrfh92 (commented):
From my point of view, the changes look fine :+1: However, I suggest that somebody who was not directly involved (@mtar ?) also has a brief look on it... 

Nevertheless, there are two points that are somehow strange and should probably be investigated: 
* ~~in the tests for `GaussianNB` a `-=` had to be removed: why?~~ (has been clarified)
* (actually my fault:) the `RobustScaler` with option `with_scaling` does not work with the same API as the other scalers in the in-place variant. For whatever reason, an API like `scaler.transform(X)` could not be realized and instead we use `X=scaler.transform(X)` here...

Since the arithmetics tests run through, an option could be to merge and to open separate issues for the two points above. I suggest to discuss in the PR-talk.

heat/core/arithmetics.py
```diff
@@ -129,6 +169,65 @@ def _add(self, other):
...
+        )
+
+    def wrap_add_(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
+        return a.add_(b)
+
+    return _operations.__binary_op(wrap_add_, t1, t2)
+
+
+DNDarray.__iadd__ = lambda self, other: add_(self, other)
+DNDarray.__iadd__.__doc__ = add_.__doc__
```
mtar, line 228:
```suggestion
DNDarray.__iadd__ = add_
```
You don't really need the lambda function here and the reference to the docstring then.

mtar (changes requested):
I was primary looking at the changes in `arithmetics.py`. Thank you for cleaning the unit tests up a little bit.
I only commented on the first occurrence for now because it would get quite repetitive to write the same thing every single time.

heat/core/arithmetics.py
```diff
@@ -28,44 +28,83 @@
...
     "remainder",
+    "remainder_",
     "right_shift",
+    "right_shift_",
     "sub",
+    "sub_",
     "subtract",
+    "subtract_",
     "sum",
+    "sum_",
```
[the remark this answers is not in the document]
LScheib: Okay so, if I get this right, you want me to delete the names of the in-place functions from this `__all__ = [...]` statement and instead add methods in the class `DNDarray` in `dndarray.py` which call the functions which I implemented in `arithmetics.py`.

So e.g. for `add_` I would add a method
```
@property
def add_(self, t2: Union[DNDarray, float]) -> DNDarray:
"""
Some explanation
"""
return arithmetics.add_(self, t2)
```

Is this correct?

heat/core/arithmetics.py
```diff
@@ -575,6 +1027,33 @@ def diff(
...
+        Input array
+    n : int, optional
+        The number of times values are differenced. If zero, the input is returned as-is.
+        ``n=2`` is equivalent to ``diff_(diff_(a))``
+    axis : int, optional
+        The axis along which the difference is taken, default is the last axis.
+    """
+    raise NotImplementedError(
+        "Not implemented as this is not an element-wise operation and there is no in-place version of `diff` in PyTorch."
+    )
```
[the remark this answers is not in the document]
LScheib: As I added in-place versions to almost all arithmetic operations, I thought it might make sense to add NotImplementsErrors for the remaining operations which don't have an in-place version. Therefore, I scanned the existing code to find examples for not implemented functions which throw such errors (e.g. `fit` in `heat/core/base.py`) and did it the same way. Also I tried to explain why specified functions like `diff_` here don't work in order to spare future contributors some time wondering why we have in-place version for all arithmetic operations except for a hand full of them.
If you say this is unnecessary and it should be self-explanatory enough, why there is an `add_` but not a `diff_` function, I can delete all of that of course.

heat/core/tests/test_arithmetics.py
```diff
@@ -98,44 +108,154 @@ def test_add(self):
...
+        a = a[:-1]
+        a += b[1:]
+        self.assertTrue((a == 1).all())
+        self.assertEqual(a.lshape, b[1:].lshape)
+
+        a = ht.ones(10, split=None)  # reset
+        a = a[1:]
+        a += b[:-1]
+        self.assertTrue((a == 1).all())
+        self.assertEqual(a.lshape, b[:-1].lshape)
```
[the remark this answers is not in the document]
LScheib: `add_` has the same property as `add` here: When adding two DNDarrays of which one has split axis `None` and the other has split axis `0`, the result has split axis `0`.

heat/core/tests/test_arithmetics.py
```diff
@@ -98,44 +108,154 @@ def test_add(self):
...
+        # Check for every possible combination of inputs whether the right solution is computed and
+        # saved in the right place and whether the second input stays unchanged. After every tested
+        # computation, we reset changed variables.
+        self.assertTrue(ht.equal(ht.add_(self.a_tensor, self.a_scalar), result))  # test result
+        self.assertTrue(ht.equal(self.a_tensor, result))  # test in-place
+        self.assertTrue(torch.equal(self.a_tensor.larray, underlying_torch_tensor))  # test in-place
+        self.assertTrue(
+            ht.equal(self.a_scalar, ht.float32(2.0))
+        )  # test if other input is unchanged
+        self.a_tensor.larray = ht.array([[1.0, 2.0], [3.0, 4.0]]).larray  # reset
```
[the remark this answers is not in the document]
LScheib: Unfortunately, for some variables the behaviour of the reset DNDarrays was not as expected and let to errors in other tests. That it is why I used this style of resetting for all variables for simplicity and uniformity.

Here is an example:
Code
After removing the `.larray` parts in line 288, it first seems that the reset worked fine, but then in the beginning of the next function `self.an_int_tensor` contains wrong values:
Fail

If I reference the underlying tensors in line 288 instead (as done in my code), everything works fine.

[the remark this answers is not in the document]
LScheib: PR talk: This behaviour should not appear, and we need to dig into this.

heat/core/arithmetics.py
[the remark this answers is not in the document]
LScheib: @mtar and @ClaudiaComito: After discussing this in the PR talk today, I looked up how NumPy and PyTorch are doing it for `add` and `add_`: They allow for directly calling the "normal" arithmetic operation with `numpy.add(T1, T2)` and `torch.add(T1, T2)` for some `numpy.ndarrays` or `torch.Tensors` `T1` and `T2`, but they do not allow for calling the in-place version of it in the same way. Therefore, I will remove the in-place functions form `__all__ = [...]` here as @mtar suggested.
Furthermore, due to the lines `DNDarray.__add__ = _add` and `DNDarray.__iadd__ = add_`, we are already able to use the operators as methods for `DNDarrays` with `T1.__add__(T2)` and `T1.__iadd__(T2)`, respectively. This is analogous to the NumPy / PyTorch functionality. Thus, I do not think that I need to change anything for the existing "normal" operators.

heat/core/tests/test_arithmetics.py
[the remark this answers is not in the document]
mtar: I guess it is a side effect. At the beginning of the unit test _self.a_tensor_ references the same object as _cls.a_tensor_ which is then modified by the in-place operation.
The change I suggested above now updates the reference of _self.a_tensor_ to the new array object. _cls.a_tensor_ still references the old modified array and that is used in the later test methods.

My new suggestion would be to use a deep copy and only use that for the in-place operations. That way we can ensure that the unit test methods don't affect each other .

heat/core/tests/test_arithmetics.py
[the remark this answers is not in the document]
mtar: The result, however, has new tensors internally.

```python
import heat as ht

a = ht.array([1,2,3,4], split=None)
b = ht.array([1,2,3,4], split=0)
a_ten = a.larray
a += b
print(a is a_ten)
```

```console
[1,0]<stdout>:False
[1,3]<stdout>:False
[1,1]<stdout>:False
[1,2]<stdout>:False
```

mtar:
There is also another issue here. The DNDarray before and after the operation are not identical, only the underlying tensors are.
```python
import heat as ht

a_ori = a = ht.array([1,2,3,4], split=None)
a_ten = a.larray
a += 2

print(a_ori.larray is a_ten)
print(a.larray is a_ten)
print(a is a_ori)
```
```console
[1,0]<stdout>:True
[1,0]<stdout>:True
[1,0]<stdout>:False
[1,1]<stdout>:True
[1,1]<stdout>:True
[1,1]<stdout>:False
[1,2]<stdout>:True
[1,2]<stdout>:True
[1,2]<stdout>:False
[1,3]<stdout>:True
[1,3]<stdout>:True
[1,3]<stdout>:False
```

heat/core/tests/test_arithmetics.py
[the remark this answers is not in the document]
LScheib: Okay, got it. Thank you. Please confirm whether my following thought is correct: 
Whenever I resplit a DNDarray, I initialize new underlying tensors.
```
def test_split(self):
    a_tensor = ht.array([[1.0, 2.0], [3.0, 4.0]], split=None)
    underlying_torch_tensor = a_tensor.larray
    a_tensor.resplit_(0)
    self.assertIs(a_tensor.larray, underlying_torch_tensor)  # test in-place
```
```
======================================================================
FAIL: test_split (heat.core.tests.test_arithmetics.TestArithmetics)
  File "/home/sche_lk/heat/heat/core/tests/test_arithmetics.py", line 115, in test_split
    self.assertIs(a_tensor.larray, underlying_torch_tensor)  # test in-place
AssertionError: tensor([[1., 2.]]) is not tensor([[1., 2.],
        [3., 4.]])

Ran 55 tests in 0.558s

FAILED (failures=1)
ok

======================================================================
FAIL: test_split (heat.core.tests.test_arithmetics.TestArithmetics)
  File "/home/sche_lk/heat/heat/core/tests/test_arithmetics.py", line 115, in test_split
    self.assertIs(a_tensor.larray, underlying_torch_tensor)  # test in-place
AssertionError: tensor([[3., 4.]]) is not tensor([[1., 2.],
        [3., 4.]])
```
Therefore, in the tested case above the addition with `add_` is not actually in-place. To make it in-place the behaviour of `add_` has to differ from `add` here so that the split axis of _a_ doesn't change as you mentioned intially.

LScheib:
Thanks @mtar. I will have a look into this as well. Interestingly, this issue does not arise when writing `ht.add_(a,2)` or `a.__iadd__(2)` instead of `a += 2`.

mtar:
> Thanks @mtar. I will have a look into this as well. Interestingly, this issue does not arise when writing `ht.add_(a,2)` or `a.__iadd__(2)` instead of `a += 2`.

 `x += y` is equivalent to `x = x.__iadd__(y)`

heat/core/tests/test_arithmetics.py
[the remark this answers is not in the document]
mtar: Many functions that change the distribution create new tensors to fit the new local shapes.

We can discuss in a larger group next time what we want to do when the axes differ. Personally, I would prefer to be stricter on the arithmetic in-place operations.

LScheib:
Questions for next PR talk:
- My main problem at the moment is the following issue observed by @mtar: The DNDarrays of `x` before and after `x += y` are not the same object, only the underlying tensors are. I do not know how to solve this. With `x.__iadd__(y)` it works as expected, which leads me to believe that my function `add_` works as it should. But as `x += y` equals `x = x.__iadd__(y)` a new DNDarray is created. [PyTorch and NumPy do not have this problem: There `x = x.__iadd__(y)` and `x.__iadd__(y)` yield the same behaviour for `x`, `y` as NumPy arrays or torch tensors.] Could it be possible to redefine `+=` to be equivalent to `x.__iadd__(y)` in this context?
- There are some in-place operators which do not have a symbol like `+=` to use them. I implemented them anyway because they are available in PyTorch. Should I add new DNDarray methods (e.g. `DNDarray.copysign_ = copysign_`) to make them usable and/or do we want them to be called directly (as `ht.copysign_(x,y)`)? In PyTorch both is possible.
  - This applies to `invert_`/`bitwise_not_`, `copysign_`, `cumprod_`/`cumproduct_`, `cumsum_`, `fmod_`, `gcd_`, `hypot_`, `lcm_`, `nan_to_num_`, and `neg_`/`negative_`.
- As discussed before, I implemented functions like `pow_` as DNDarray methods. Can I discard the alias support `power_ = pow_` then? This seems to be unnecessary if `pow_` is not directly callable.
- `divmod` and `pos` are element-wise operators which could be implemented in-place but are not (yet) in PyTorch and NumPy. Therefore, do we want to throw `NotImplementedErrors` for them? And leave the functions to do so as place holders in `arithmetics.py` for later implementation? 
- Which functionality do we want when in-place operations are used for DNDarrays with different split axes?
  - For example, for `x + y` with `x.split == None` und `y.split != None` the result has the same split axis as `y`. Now for `x += y` it might make more sense to keep the split of `x` as otherwise new underlying tensors are created, and therefore the operation is not actually in-place.
  - Secondly, what about 2D arrays with different split axes? If `x.split == 0` and `y.split == 1`, `x + y` returns a DNDarray with split axis 0 while `y + x` returns it with split axis 1. @mtar's comment suggested that for `+=` the case should throw an exception, but alternatively I could implement this case in a way that `x += y` internally resplits `y` to add it in-place.
- Regarding the side effect of using class variables for the tests of in-place operations (discussed here with @mtar): Do I need to define copies of the class variables for the in-place operations in the beginning of every test function (see `test_add_`) or is there a better way? Doing it the way I did before, resetting the the underlying tensors (see `test_sub_`), seemed a bit more efficient to me in terms of repetition and code length.

heat/core/tests/test_arithmetics.py
```diff
@@ -98,44 +109,173 @@ def test_add(self):
...
+        # Check for every possible combination of inputs whether the right solution is computed and
+        # saved in the right place and whether the second input stays unchanged. After every tested
+        # computation, we reset changed variables.
+        self.assertTrue(ht.equal(a_tensor.__iadd__(a_scalar), result))  # test result
+        self.assertTrue(ht.equal(a_tensor, result))  # test result in-place
+        self.assertIs(a_tensor, a_tensor_double)  # test in-place
+        self.assertIs(a_tensor.larray, underlying_torch_tensor)  # test in-place
+        self.assertTrue(ht.equal(a_scalar, ht.float32(2.0)))  # test if other input is unchanged
+        a_tensor = a_tensor_double = ht.array([[1.0, 2.0], [3.0, 4.0]])  # reset
+        underlying_torch_tensor = a_tensor.larray  # reset
```
mtar, line 139:
```suggestion
        underlying_torch_tensor.copy_(self.a_tensor.larray)  # reset
```
Regarding the last point: Now, you can write the reset in a single line and save some space :grinning:

LScheib:
@mtar @mrfh92 Please have a look at my `add_` and `test_add_` functions. The failed `Receive PR / build (pull_request)` is due to the changed functionality in regard to differently split DNDarrays. I will have a closer look at these errors later.

heat/core/arithmetics.py
```diff
@@ -129,6 +161,79 @@ def _add(self, other):
...
+              [4., 4.]], dtype=ht.float32, device=cpu:0, split=None)
+    >>> T2
+    DNDarray([[4., 4.],
+              [4., 4.]], dtype=ht.float32, device=cpu:0, split=None)
+    >>> s
+    2.0
+    """
+    if not isinstance(t1, DNDarray):
+        raise TypeError(
+            "Input 1 must be a DNDarray and input 2 either a DNDarray or a scalar. "
```
mrfh92, line 207: Our convention for printing is to use fstrings, so sth like `TypeError(f"Input ... but is {type(t1)} and {type(t2)}")`

heat/core/arithmetics.py
```diff
@@ -129,6 +161,79 @@ def _add(self, other):
 DNDarray.__radd__.__doc__ = add.__doc__
 
 
+def add_(t1: DNDarray, t2: Union[DNDarray, float]) -> DNDarray:
+    """
+    WILL BE REVISED LATER
```
mrfh92, line 166: Here we can briefly recall that either `t1.split=t2.split` or `t2.split=None` is required

LScheib:
After modifying the support of differently split DNDarrays for `add_`, there are two problems with other tests that need to be addressed:
1.)
```
======================================================================
FAIL: test_hermitian (heat.utils.data.tests.test_matrixgallery.TestMatrixgallery)
  File "/home/sche_lk/heat/heat/utils/data/tests/test_matrixgallery.py", line 40, in test_hermitian
    self.assertTrue(A.dtype == ht.complex128)
AssertionError: False is not true
```
2.) The following error only occurs for execution on 2 processes:
```
======================================================================
ERROR: test_convolve (heat.core.tests.test_signal.TestSignal)
  File "/home/sche_lk/heat/heat/core/tests/test_signal.py", line 69, in test_convolve
    conv = ht.convolve(signal, dis_kernel_odd, mode=mode)
  File "/home/sche_lk/heat/heat/core/signal.py", line 182, in convolve
    signal_filtered += global_signal_filtered[start_idx : start_idx + gshape]
  File "/home/sche_lk/heat/heat/core/arithmetics.py", line 227, in add_
    return _operations.__binary_op(wrap_add_, t1, t2, out=t1)
  File "/home/sche_lk/heat/heat/core/_operations.py", line 171, in __binary_op
    sanitation.sanitize_out(out, output_shape, output_split, output_device, output_comm)
  File "/home/sche_lk/heat/heat/core/sanitation.py", line 299, in sanitize_out
    raise ValueError(f"Expecting output buffer of split {output_split}, got {out.split}")
ValueError: Expecting output buffer of split 0, got None

======================================================================
ERROR: test_convolve (heat.core.tests.test_signal.TestSignal)
  File "/home/sche_lk/heat/heat/core/arithmetics.py", line 208, in add_
    resulting_shape = torch.broadcast_shapes(t1.larray.shape, t2.larray.shape)
  File "/home/sche_lk/.virtualenvs/heatenv/lib/python3.8/site-packages/torch/functional.py", line 126, in broadcast_shapes
    raise RuntimeError("Shape mismatch: objects cannot be broadcast to a single shape")
RuntimeError: Shape mismatch: objects cannot be broadcast to a single shape

During handling of the above exception, another exception occurred:

  File "/home/sche_lk/heat/heat/core/tests/test_signal.py", line 69, in test_convolve
    conv = ht.convolve(signal, dis_kernel_odd, mode=mode)
  File "/home/sche_lk/heat/heat/core/signal.py", line 182, in convolve
    signal_filtered += global_signal_filtered[start_idx : start_idx + gshape]
  File "/home/sche_lk/heat/heat/core/arithmetics.py", line 210, in add_
    raise ValueError(
ValueError: The differently split inputs (splits None and 0 can not be processed in-place without resplitting, because the underlying tensors have shapes which are not broadcastable (shapes torch.Size([18]) and torch.Size([0])).
```

mrfh92:
Regarding 1.) there seems to be a problem with data-type-broadcasting for `add_`: the problem occures in lines 61/62 of `heat/utils/data/matrixgallery.py`:
```
        matrix = randn(n, n, dtype=real_dtype, split=split, device=device, comm=comm)
        matrix += 1j * randn(n, n, dtype=real_dtype, split=split, device=device, comm=comm)
```
If you replace these lines by the non-in place variant 
```
matrix = randn(n, n, dtype=real_dtype, split=split, device=device, comm=comm) + 1j * randn(n, n, dtype=real_dtype, split=split, device=device, comm=comm)
```
everything works as expected. 
Actually, doing this not in-place is the correct way, because adding a (w.r.t. to memory) much larger complex128-tensor in-place to a float64-tensor does not make any sense. 

However, it would be a good idea to throw a type-error in this case, i.e. if one tries to perform an in-place operation with a data-type combination that would increase the amount of memory needed.

mrfh92:
Regarding 2.) the problem is in the following line of `heat/core/signal.py`:
```
signal_filtered += global_signal_filtered[start_idx : start_idx + gshape]
```
The problem is that the lshapes of the arrays on the left and the right hand side of += do not coincide in every case. This is very likely to occure if one creates potentially unbalanced DNDarrays by slicing as done here. I introduced a simple try-except workaround to keep the in-place variant whenever possible and to proceed differently when needed.

LScheib:
New errors thrown after pushing that are probably due to wrong usage of `*=`:

```
======================================================================
ERROR: test_fftfreq_rfftfreq (heat.fft.tests.test_fft.TestFFT)
  File "/home/runner/work/heat/heat/heat/fft/tests/test_fft.py", line 158, in test_fftfreq_rfftfreq
    y = ht.fft.fftfreq(n, d=d, split=0)
  File "/home/runner/work/heat/heat/heat/fft/fft.py", line 474, in fftfreq
    return __fftfreq_op(
  File "/home/runner/work/heat/heat/heat/fft/fft.py", line 344, in __fftfreq_op
    freqs *= channel_width
  File "/home/runner/work/heat/heat/heat/core/arithmetics.py", line 2317, in mul_
    raise TypeError(
TypeError: Can not cast from <class 'heat.core.types.float32'> to <class 'heat.core.types.int32'> for in-place operations.

======================================================================
ERROR: test_fftshift_ifftshift (heat.fft.tests.test_fft.TestFFT)
  File "/home/runner/work/heat/heat/heat/fft/tests/test_fft.py", line 202, in test_fftshift_ifftshift
    x = ht.fft.fftfreq(5, d=1 / 5, split=0) + 0.1 * ht.fft.fftfreq(5, d=1 / 5, split=0).reshape(
  File "/home/runner/work/heat/heat/heat/fft/fft.py", line 474, in fftfreq
    return __fftfreq_op(
  File "/home/runner/work/heat/heat/heat/fft/fft.py", line 344, in __fftfreq_op
    freqs *= channel_width
  File "/home/runner/work/heat/heat/heat/core/arithmetics.py", line 2317, in mul_
    raise TypeError(
TypeError: Can not cast from <class 'heat.core.types.float32'> to <class 'heat.core.types.int32'> for in-place operations.

======================================================================
ERROR: test_RobustScaler (heat.preprocessing.tests.test_preprocessing.TestRobustScaler)
  File "/home/runner/work/heat/heat/heat/preprocessing/tests/test_preprocessing.py", line 267, in test_RobustScaler
    Y = scaler.transform(X)
  File "/home/runner/work/heat/heat/heat/preprocessing/preprocessing.py", line 566, in transform
    Y *= self.scale_
  File "/home/runner/work/heat/heat/heat/core/arithmetics.py", line 2317, in mul_
    raise TypeError(
TypeError: Can not cast from <class 'heat.core.types.float64'> to <class 'heat.core.types.float32'> for in-place operations.
```

mrfh92:
Two changes required to make use of in-place routines without allowing questionable dtype-casts for in-place ops: 
* in preprocessing I found some problems that originated from the fact that the dtype of the output of the percentile-function depended on the percentile dtype (i.e. ht.float64 whenever you use a float here) but not on the dtype of the input array. I decided to change this in the sense that now depends on the dtype of the input array. 
* frequencies in fft were computed using arange and an in-place division. I adapted the dtype chosen in arange in order to use in-place without illegal dtype-casts.

mrfh92:
> Is there anything I can do about the failing benchmark report?
> 
> ```
> Error: # :warning: **Performance Alert** :warning:
> 
> Possible performance regression was detected for benchmark.
> Benchmark result of this commit is worse than the previous benchmark result exceeding threshold `2`.
> ```

No, since this is not an actual error. The problem is that the performance monitoring currently runs on (shared) gitlab runners, so we actually do not measure how fast our code is but rather how busy the runners are overall. In the future we will transfer this to more appropriate runners and then the results will be meaningful.

LScheib:
> > Is there anything I can do about the failing benchmark report?
> > ```
> > Error: # :warning: **Performance Alert** :warning:
> > 
> > Possible performance regression was detected for benchmark.
> > Benchmark result of this commit is worse than the previous benchmark result exceeding threshold `2`.
> > ```
> 
> No, since this is not an actual error. The problem is that the performance monitoring currently runs on (shared) gitlab runners, so we actually do not measure how fast our code is but rather how busy the runners are overall. In the future we will transfer this to more appropriate runners and then the results will be meaningful.

Ah, right. Now I remember you talked about that in one of our PR meetings. Thanks.

mrfh92 (dismissed):
Looks fine to me :) 
Thanks for the work! 👍

ClaudiaComito (changes requested):
@LScheib thanks for the great work, this is a very important contribution.  My only change request is related to the split heuristics. 

Thanks again and all the best!

ClaudiaComito (changes requested):
@LScheib thanks for the changes. This looks fine to me. As far as I'm concerned, you can apply the same change to all other operations.

With this implementation, we are allowing in-place operations even if the second operand requires redistribution, i.e. the distribution of the two operands is not aligned. I think this is OK as the fundamental condition here is being able to write into the first operand's buffer.

heat/core/_operations.py
```diff
@@ -84,6 +84,32 @@ def __binary_op(
...
+        "wrap_mul_",
+        "wrap_pow_",
+        "wrap_remainder_",
+        "wrap_bitwise_right_shift_",
+        "wrap_sub_",
+    }
+    if operation.__name__ in in_place_arithmetics:
+        if not types.can_cast(types.heat_type_of(t2), types.heat_type_of(t1)):
+            raise TypeError(
+                f"Can not cast from {types.heat_type_of(t2)} to {types.heat_type_of(t1)} for "
```
mtar, line 109:
```suggestion
                f"Cannot cast from {types.heat_type_of(t2)} to {types.heat_type_of(t1)} for "
```

LScheib:
I added a `try... except` block in `add_` (and also in the other functions) to extend the error description in the case of differently split DNDarrays. Otherwise, we would only throw a `NotImplementedError`, which seems wrong to me, as we do not intend to implement arithmetic in-place operations for DNDarrays with different split axes that don't make sense.

heat/core/arithmetics.py
```diff
@@ -129,6 +132,72 @@ def _add(self, other):
...
+        return a.add_(b)
+
+    try:
+        return _operations.__binary_op(wrap_add_, t1, t2, out=t1)
+    except NotImplementedError:
+        raise ValueError(
+            f"The differently split inputs (splits {t1.split} and {t2.split} cannot be processed "
+            + "in-place without resplitting, because the underlying tensors have shapes which are "
+            + f"not broadcastable (shapes {t1.larray.shape} and {t2.larray.shape})."
+        )
```
ClaudiaComito, line 194:
```suggestion
            f"In-place operation not allowed: operands are distributed along different axes.
            Operand 1 with shape {t1.shape} is split along axis {t1.split}.
            Operand 2 with shape {t2.shape} is split along axis {t2.split}."
        )
```

How about this? The problem isn't that local shapes aren't broadcastable. They might be broadcastable, but still the operation would return the wrong result because it wouldn't be performed on the correct elements of the second operand.

ClaudiaComito (changes requested):
Thanks @LScheib , you're right about the NotImplementedError not applying here, I like your solution. I suggest a rephrase of the error message (see suggestion in `add_`), if you agree please also apply it to the other operations.

Otherwise, from my point of view this is a great PR and ready to be merged.

ClaudiaComito:
@mrfh92 as discussed earlier today, I ran a basic `tracemalloc` check and realized that the memory footprint is actually larger for the new in-place operations than it was before (when we didn't have in-place ops). Here an example with `torch.add_`:

```python
import torch
import tracemalloc

m = 190000
n = 8000

a = torch.arange(m*n, dtype = torch.int64).reshape(m, n)
b = torch.arange(m*n, dtype = torch.int64).reshape(m, n)
tracemalloc.start()
current, peak = tracemalloc.get_traced_memory()
print(f"START: Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")
a += b
current, peak = tracemalloc.get_traced_memory()
print(f"IN-PLACE TORCH: Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")
c = a + b
current, peak = tracemalloc.get_traced_memory()
print(f"OUT-OF-PLACE TORCH: Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")
```
Result:
```bash
START: Current memory usage is 0.0MB; Peak was 0.0MB
IN-PLACE TORCH: Current memory usage is 0.0MB; Peak was 0.000485MB
OUT-OF-PLACE TORCH: Current memory usage is 0.0001MB; Peak was 0.000527MB
```


And now the Heat version: 
```python
import torch
import tracemalloc
import heat as ht

m = 190000
n = 8000

a = torch.arange(m*n, dtype = torch.int64).reshape(m, n)
b = torch.arange(m*n, dtype = torch.int64).reshape(m, n)
a = ht.array(a)
b = ht.array(b)
tracemalloc.start()
current, peak = tracemalloc.get_traced_memory()
print(f"START: Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")
a += b
current, peak = tracemalloc.get_traced_memory()
print(f"IN-PLACE HEAT: Current memory usage is {current / 10**6}MB; Peak was {peak / 10**6}MB")
```

Outcome, 1 process. 

On `main`:
```bash
START: Current memory usage is 0.0MB; Peak was 0.0MB
IN-PLACE HEAT: Current memory usage is 0.00844MB; Peak was 0.008772MB
```

On `features/1184-in-place_operators`:
```bash
START: Current memory usage is 0.0MB; Peak was 0.0MB
IN-PLACE HEAT: Current memory usage is 0.009144MB; Peak was 0.010773MB
```

mrfh92:
@ClaudiaComito For the example you show above you have two arrays of size 190000x8000 with int64 entries, and in my opinion this means: 

190000 x 8000 x 8 byte = 12.16 * 10^9 bytes ~ 12 GB. 

In particular, I get a memory error if I make these arrays even larger... 
However, `tracemalloc` yields results below a single MB (on my machine too); even if I move the `.start()` and `.get_traced_memory()` directly around the lines where `a` and `b` are generated, I only get these small numbers. So I guess that something with tracemalloc is completely wrong (or I do not understand correctly what this function does/is supposed to do).

mrfh92:
I have done a similar experiment on a SLURM-cluster and tracked the overall RSS with SLURM jobinfo: 
**"normal":**
```
import heat as ht

x = ht.zeros((2000,2000,2000),dtype=ht.int64,split=0)
y = ht.ones((2000,2000,2000),dtype=ht.int64,split=0)
x+y
```
**in-place:**
```
import heat as ht

x = ht.zeros((2000,2000,2000),dtype=ht.int64,split=0)
y = ht.ones((2000,2000,2000),dtype=ht.int64,split=0)
x+=y
```
were executed both on `main` and on this branch, on 1 and 2 nodes (CPU, 8 procs per node, 10 CPUs per proc). 

**1 node**

| | main | this branch | 
|------| ------ | ------ |

mrfh92:
@ClaudiaComito In https://github.com/pytorch/pytorch/issues/91823 they seem to have the same problem with tracemalloc

mrfh92:
Follow up on the above, now with `time.sleep(5)`'s added in the hope to produce more stable measurements despite logging frequency of 1/s. 

| | | MaxRSS | AveRSS | 
| --- | --- | --- | --- |
| normal | (`main`) | | |
| in-place | (`main`) | | |
| normal | (this branch) | | |
| in-place | (this branch) | | |

heat/core/arithmetics.py
mrfh92: I have done the suggested modifications also at the other places where this error message appeared.

ClaudiaComito (dismissed):
Thanks a lot for your work @LScheib and @mrfh92 !

ClaudiaComito:
@mrfh92 we have a backward compatibility problem with PyTorch 1.8 and 1.9.

mrfh92:
@ClaudiaComito As PyTorch 1.10 has been release in Oct 21, 2021, shouldn't we drop support of 1.8 and 1.9 anyway due to NEP29-like support?

ClaudiaComito:
> @ClaudiaComito @mtar @JuanPedroGHM As PyTorch 1.10 has been release in Oct 21, 2021, shouldn't we drop support of 1.8 and 1.9 anyway due to NEP29-like support?
> 
> if yes, I would open an issue and PR for doing so.

@mrfh92 I'm all for it!

ClaudiaComito (approved):

Issue #1096, opened 2023-02-08
Provide a solver for triangular systems


mrfh92:
**Feature functionality**
Provide a solver for the linear equation $Ax = b$, where $b \in \mathbb{K}^{m \times n}$, $A \in \mathbb{K}^{m \times m}$ are two-dimensional `DNDarrays` such that $A$ is either upper- or lower-triangular. 

Intended scope: 
- different combinations of `split` for $A$ and $b$
- of course, scalability of the routine is desired

**Additional context**
* Often required in combination with Cholesky decompositon #886 
* also required in context of linear regression (implementing a "QR + triangular solve"-based linear regressor for massive data could be a cool second part of a corresponding student project or PR)

mrfh92:
I would suggest to put `triangular_solve` into `.../solver.py` and to create a new file `/core/linalg/cholesky.py` for `cholesky`

Gaurav17Joshi:
Hi, I have created a new function triangular_solver in .../solver.py. It takes in A and B as inputs and returns the solution x for Ax = B, also, this function raises error flag if the user does not input a triangular matrix or the matrix is not full rank.

<img width="458" alt="lower matrix (No split)" src="https://user-images.githubusercontent.com/111291834/218690475-a0503aac-e7f8-4f78-950d-f4ffb4739afe.png">

<img width="458" alt="lower matrix (Split)" src="https://user-images.githubusercontent.com/111291834/218690705-a6743558-08e0-41bd-81f2-6303b8cc8227.png">

<img width="524" alt="upper matrix (No split)" src="https://user-images.githubusercontent.com/111291834/218690764-89851118-932c-4ccf-97f9-db37f1a47968.png">

<img width="474" alt="upper matrix (split)" src="https://user-images.githubusercontent.com/111291834/218690791-5984285d-94c4-477f-84d7-24f840f1d57d.png">

Gaurav17Joshi:
Raising Errors:
<img width="703" alt="Not Triangular" src="https://user-images.githubusercontent.com/111291834/218691028-a192f206-9957-4dc7-abd2-fff1746b068e.png">
<img width="700" alt="Not Full rank" src="https://user-images.githubusercontent.com/111291834/218691045-0e9c0c47-88b5-4543-b4da-80a6ac7e3c4a.png">

Gaurav17Joshi:
@mrfh92 , can I create a PR so , that my code can be reviewed.

mrfh92:
[...]

Issue #1104, opened 2023-02-14
Implement consistent linear algebra for arrays with dimension > 2, in particular `matmul`


mrfh92:
In Pytorch, most linear algebra routines can be applied to batches of matrices. In Heat it would be nice to have a similar rule how to "define" linear algebra (`matmul`,`qr`,`cholesky`,`svd` etc.) for `DNDarray`'s with dimension exceeding 2. 

To do so, follow the PyTorch convention (see, e.g., https://pytorch.org/docs/stable/generated/torch.linalg.solve.html#torch.linalg.solve or also the code in #893): for arrays with `ndim>=3` we regard the last two dimensions (in the case of matrices) or the last dimension (in the case of vectors) of the array as "linear algebra-dimensions" and the indices of the first `ndim-2` (matrices) or `ndim-1` (vectors)  dimensions as indices of a "batch of linear algebra objects" 

**Particular example:** see `matmul` (#890 )

**Suggestion:**

* *"minimal solution" (data/batch-parallel higher-dim matmul only)*: 
    introduce batched matmul for `ndim>=3` for the case that the `split`-dimension is a batch-dimension, i.e. no parallel matrix-matrix multiplication algorithm is required
* *"full solution"*: go through the existing matmul-code and adapt it in such a way that you add batch-dimensions in order to handle the case of `split` equal to one of the two last dimensions that has been omitted in the "mimal solution" 

**Note: qr, svd, cholesky have been split off to #1320**

ClaudiaComito:
Related: #890

mrfh92:
see #890 for the particular case of `matmul` on higher dim arrays 

Reviewed within #1109 : still open and relevant

PR #1236, opened 2023-10-10
Add solver for distributed triangular systems 

mrfh92:
This PR is intended to add a triangular solver (see, e.g., `torch.linalg.solve_triangular`) to Heat; see #1096. 

## Due Diligence

- General:
- Implementation:


mrfh92:
Things to be addressed: 
* all combinations of splits for A and b; b either vector or matrix 
* performance (scaling) 
* unit tests to be added

mrfh92:
Things to be done: 
* which cases have been covered so far? (`A.split` $\in \{$ `0`, `1,`, `None`$\}$,

mrfh92:
With a view on #1104 #1320 it might also be usefull to include the case of batched inputs (however, for the moment it should suffice to cover the simple case split=batch-dimension)

mrfh92:
There seems to be a small bug somewhere regarding the device, so probably you habe created a tensor somewhere without specifying the device and thus the default (cpu) has been used, which results in a bug when it has to interact with other tensors on GPU.

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,187 @@ def lanczos(
...
+        A - an upper triangular (possibly batched) invertible square (n x n) matrix
+        b - (possibly batched) n x k matrix
+
+    Output:
+        The unique solution x of A * x = b.
+
+    Vectors b have to be given as n x 1 matrices.
+    """
+    if not isinstance(A, DNDarray) or not isinstance(b, DNDarray):
+        raise RuntimeError("Arguments need to be a DNDarrays.")
```
ClaudiaComito, line 289:
```suggestion
        raise RuntimeError(f"Arguments need to be of type DNDarray, got {type(A)}, {type(b)}.")
```

ClaudiaComito: This should be a TypeError

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,187 @@ def lanczos(
...
+    if not A.ndim >= 2:
+        raise RuntimeError("A needs to be a 2D matrix.")
+    if not b.ndim == A.ndim:
+        raise RuntimeError("b needs to be a 2D matrix.")
```
ClaudiaComito, line 293: These would be ValueErrors I think. The message is a bit confusing, because we do allow n-D arrays

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,187 @@ def lanczos(
...
+    if not A.shape[-2] == A.shape[-1]:
+        raise RuntimeError("A needs to be a square matrix.")
```
ClaudiaComito, line 295: ValueError, see https://docs.python.org/3/library/exceptions.html#

"... a square matrix or a batch of square matrices"

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,187 @@ def lanczos(
...
+
+    batch_dim = A.ndim - 2
+    batch_shape = A.shape[:batch_dim]
+
+    if not A.shape[:batch_dim] == b.shape[:batch_dim]:
+        raise RuntimeError("Batch dimensions of A and b must be of the same shape.")
+    if b.split == batch_dim + 1:
+        raise RuntimeError("split=1 is not allowed for the right hand side.")
+    if not b.shape[batch_dim] == A.shape[-1]:
+        raise RuntimeError("Dimension mismath of A and b.")
```
ClaudiaComito, line 305: All ValueErrors I think.

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,187 @@ def lanczos(
...
+
+    if (
+        A.split is not None and A.split < batch_dim or b.split is not None and b.split < batch_dim
+    ):  # batch split
+        if A.split != b.split:
+            raise RuntimeError(
+                "If a split dimension is a batch dimension, A and b must have the same split dimension."
```
ClaudiaComito, line 312: Value Error

Let's help the users out here, suggest what they're suppose to change in practice.

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,187 @@ def lanczos(
...
+            )
+    else:
+        if (
+            A.split is not None and b.split is not None
+        ):  # both la dimensions split --> b.split = batch_dim
+            if not all(A.lshape_map[:, A.split] == b.lshape_map[:, batch_dim]):
+                raise RuntimeError("Local arrays of A and b have different sizes.")
```
ClaudiaComito, line 319: what's the user supposed to do about this?

heat/core/linalg/tests/test_solver.py
```diff
@@ -135,3 +135,80 @@ def test_lanczos(self):
...
+
+        for s0, s1 in (None, None), (-2, -2), (-1, -2), (-2, None), (-1, None), (None, -2):
+            a.resplit_(s0)
+            b.resplit_(s1)
+
+            res = ht.linalg.solve_triangular(a, b)
+            self.assertTrue(ht.equal(res, c))
+
+        # batched tests
+        batch_shape = (10,)  # batch dimensions shape
```
ClaudiaComito, line 181: Shall we test higher-dimensional batches as well, i.e. 6-dimensional with split=3?

ClaudiaComito (changes requested):
Great job @FOsterfeld @mrfh92 ! Thanks a lot for this. 

I have a few comments, mostly on docs and error messages. Otherwise it looks great to me.

heat/core/linalg/solver.py
FOsterfeld: I not sure whether this error can even occur because at this point both dndarrays should have the same size in that axis. (Otherwise lines 294 or 304 should have thrown an exception.)
What do you think @mrfh92 ?

heat/core/linalg/tests/test_solver.py
FOsterfeld: That's a good idea, I have done so with 22507ae77498b616da01b65b85cb51f3c52de77e.

heat/core/linalg/solver.py
mrfh92: Actually, this can happen if at least one of the arrays does not follow the "standard" splitting scheme, e.g., if it is unbalanced for whatever reason.

mrfh92: I have updated the error message accordingly.

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,187 @@ def lanczos(
...
+    Solve upper triangular systems of linear equations.
+
+    Input:
+        A - an upper triangular (possibly batched) invertible square (n x n) matrix
+        b - (possibly batched) n x k matrix
+
+    Output:
+        The unique solution x of A * x = b.
+
+    Vectors b have to be given as n x 1 matrices.
```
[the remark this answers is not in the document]
mrfh92: I have updated the docstring according to our conventions.

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,196 @@ def lanczos(
         V.resplit_(axis=None)
 
     return V, T
+
+
+def solve_triangular(A: DNDarray, b: DNDarray) -> DNDarray:
+    """
+    This function provides a solver for (possibly batched) upper triangular systems of linear equations: it returns x in Ax = b, where A is a (possibly batched) upper triangular matrix and
```
ClaudiaComito, line 277:
```suggestion
    This function provides a solver for (possibly batched) upper triangular systems of linear equations: it returns `x` in `Ax = b`, where `A` is a (possibly batched) upper triangular matrix and
```

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,196 @@ def lanczos(
...
+    b a (possibly batched) vector or matrix of suitable shape, both provided as input to the function.
```
ClaudiaComito, line 278:
```suggestion
    `b` a (possibly batched) vector or matrix of suitable shape, both provided as input to the function.
```

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,196 @@ def lanczos(
...
+    The implementation builts on the corresponding solver in PyTorch and implements a block-wise version thereof.
```
ClaudiaComito, line 279:
```suggestion
    The implementation builts on the corresponding solver in PyTorch and implements a (MPI-parallel? memory-distributed?) block-wise version thereof.
```

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,196 @@ def lanczos(
...
+
+    Parameters
+    ----------
+    A : DNDarray
+        An upper triangular (possibly batched) invertible square (n x n) matrix, i.e. an DNDarray of shape (..., n, n).
```
ClaudiaComito, line 284:
```suggestion
        An upper triangular invertible square (n x n) matrix or a batch thereof,  i.e. a ``DNDarray`` of shape `(..., n, n)`.
```

heat/core/linalg/solver.py
```diff
@@ -269,3 +270,196 @@ def lanczos(
...
+    b : DNDarray
+        a (possibly batched) n x k matrix, i.e. an DNDarray of shape (..., n, k), where the batch-dimensions denoted by ... need to coincide with those of A.
+        Vectors have to be provided as n x 1 matrices and the split dimension of b must the second last dimension if not None.
```
ClaudiaComito, line 287: Not sure I understand this sentence (line 287), esp. in relation to the previous one.

ClaudiaComito (changes requested):
Thanks a lot for the changes @mrfh92 @FOsterfeld , looks great, I only have a few more editing suggestions.

heat/core/linalg/solver.py
mrfh92: Has been addressed

heat/core/linalg/solver.py
mrfh92: Has been (hopefully) clarified.

mrfh92:
@ClaudiaComito I have taken your review into account now.

ClaudiaComito (approved):
Thanks a lot @FOsterfeld , @mrfh92 !

PR #1247, opened 2023-10-16
Continuous Benchmarking for the Preprocessing Module

### LINKED ISSUE #1246: Continuous Benchmarking for Preprocessing Module
The preprocessing module currently lacks tests in the continuous benchmarking 
W.r.t. #1184 it could be interesting to track changes in required memory for the in-place-versions of the transformations.

mrfh92:
This PR adds continuous benchmarking for the preprocessing module and therefore resolves #1246 

Monitoring memory consumption of the respective routines could be particularly interesting for #1184.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 Heat is a distributed tensor framework for high performance data analytics.
 
-## Project Status
+# Project Status
 
 [![CPU/CUDA/ROCm tests](https://codebase.helmholtz.cloud/helmholtz-analytics/ci/badges/heat/base/pipeline.svg)](https://codebase.helmholtz.cloud/helmholtz-analytics/ci/-/commits/heat/base)
 [![Documentation Status](https://readthedocs.org/projects/heat/badge/?version=latest)](https://heat.readthedocs.io/en/latest/?badge=latest)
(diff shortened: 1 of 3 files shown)


JuanPedroGHM (approved):
The benchmarks look great code wise. One problem I'm seeing from the [report summary](https://github.com/helmholtz-analytics/heat/actions/runs/6571797733#summary-17851646080) is that the new benchmarks, except **apply_inplace_robust_scaler_and_inverse**, have a runtime of less than 0.01 seconds, some even are in the millisecond range. It would probably be beneficial to increase the size of X for those benchmarks to get more measurable results. We should probably adjust all the sizes again for the GPU tests on the GitLab runners, as the current ones were tailored for the GitHub runners, which were considerably weaker. Other than that, approved.

mrfh92:
@JuanPedroGHM thx for reviewing. I had a look in the report: as far as I understand it, executing `apply_inplace_robust_scaler_and_inverse` on GPU currently needs more than one minute, and since this is done 10x, benchmarking this function requires already more than 10 minutes. Therefore, the only possibility to increase the problem size would be to keep the small size for this routine and to increase it only for the other preprocessing routines. But I am not sure whether this is a good idea, because at a first look one might be tempted to leave the problem size the same for all preprocessing benchmarks. Maybe we should discuss this in the PR meeting on Monday.

Issue #1243, opened 2023-10-13
Refactoring of `communication`: separate `mpi4py`-wrappers from `DNDarrays`


mrfh92:
Currently, the code wrapping `mpi4py` for `torch.Tensor`'s is contained in `heat/core/communication.py` and not fully separated from constructions related to the `DNDarray`-class (e.g., computation of chunk sizes is also introduced there). 

I would suggest to separate this clearly, which may also make #482 and similar work easier: 

* code wrapping `mpi4py` for `torch.Tensor`'s becomes an own module of heat 
* in `heat/core/communication.py` this module can be used 

**Diclaimer:** feel free to discuss other ideas for refactoring and/or that status-quo does not need changes

mrfh92:
As mentioned in #1772, we discussed that refactoring communication.py is not a priority at the moment.

PR #1265, opened 2023-11-20
Features/1243 refactoring of communication separate mpi4py  wrappers from dn darrays

mrfh92:
This PR addresses #1243 

Description: TBD ... see comments below


mrfh92:
Done so far: 

* a new module `communication_backends` has been introduced 
* `communication.py` from `heat/core` has been divided into `communications.py` (containing the basis `Communication`-class) and `mpi4py4torch.py` (containing the `MPICommunication`-class and the wrappers for `mpi4py`-functionality for `torch`) 
* calling any MPI-communication with DNDarrays as buffers is deprecated; everywhere in the code such calls have been replaced by the equivalent way using the local arrays of a DNDarray as buffers instead (actually, such a check and reference of the local arrays has been done in all communication-functions previously) 

To do: 

* the tests for the no-more existing `heat/core/communication.py` are still in `heat/core/tests/test_communication.py` (although being adapted according to the last bullet point above); however, it would be favourable to rewrite them as tests to `heat/communication_backends/mpi4py4torch.py`, i.e. to remove any reference to `DNDarrays` there in order to achieve a clear separation.

mrfh92:
As discussed this week, no refactoring of communication.py is planned at the moment.

PR #1269, opened 2023-11-22
Bugs/1258  bug lasso does not work on gpu

mrfh92:
Bug fix has been merged in release/1.3.x only so far.


ClaudiaComito (approved):
Thanks @mrfh92 !

Issue #1287, opened 2023-11-30
Batch-parallel K-means and K-medians


mrfh92:
Implement variants of K-means and K-medians that can take advantage of Heats data structure

PR #1288, opened 2023-11-30
Batch-parallel K-means and K-medians 

mrfh92:
### Due Diligence

- General:
- Implementation:
    - [X] unit tests: all split configurations tested (here: only split=0 allowed)

### Description

This PR implements a variant of K-Means that can take advantage of Heat data distribution structure. The algorithm is described in the paper https://doi.org/10.1016/j.cie.2020.107023. In essence, the idea is to perform K-means on each chunk of data separately in parallel as a first step. In the second step, the obtained cluster centers are "merged": the set of all process-local cluster centers is collected, and then another application of K-means is used to determine the final "global" cluster centers als cluster centers of the set of "local" cluster centers. 
Of course, this idea generalizes to K-Medians in a straightforward way. To improve scalability of this approach even to very high number of processes, the "merging" can also be done in hierarchical manor.

**Caveat:** This does not necessarily yield the same results as the classical K-Means. 

**Issue/s resolved:** #1287 

### Changes proposed:

- New classes `heat.cluster.BatchParallelKMeans` and  `heat.cluster.BatchParallelKMedians`

### Type of change
new feature

### Memory requirements
TBD

### Performance
Example of ~130GB on up to 8 GPU-nodes of the Terrabyte-Cluster. Left plot shows standard K-means (4 clusters) and right plot shows BatchParallelKmeans (4 and 40 clusters) 

no


ClaudiaComito (changes requested):
@mrfh92 this is great. 

Would it be possible or advisable to make batch-parallel clustering available from within a general `ht.cluster.KMeans` call? I'm thinking something like:

```python
kmeans = ht.cluster.KMeans(n_clusters=k, parallel="batch")
```
where `parallel` can be "batch" or "standard".

mrfh92:
@ClaudiaComito I think K-Means/Medians and Batch-parallel K-Means/Medians should be handled as separate functions since from an algorithmic point of view they are quite different. However, it's possible (and I have done so) to add batch-parallel K-Means/Medias as additional Initialization for classical K-Means/Medians

heat/cluster/batchparallelclustering.py
```diff
@@ -0,0 +1,446 @@
...
+
+"""
+Actual classes for batch parallel K-means and K-medians
+"""
+
+
+class BatchParallelKMeans(_BatchParallelKCluster):
+    r"""
+    Batch-parallel K-Means clustering algorithm from Ref. [1].
+    This requires data to be given as DNDarray of shape (n_samples, n_features) with split=0 (i.e. split along the sample axis).
```
ClaudiaComito, line 326:
```suggestion
    The input must be a ``DNDarray`` of shape `(n_samples, n_features)`, with split=0 (i.e. split along the sample axis).
```

heat/cluster/batchparallelclustering.py
```diff
@@ -0,0 +1,446 @@
...
+    The idea of the method is to perform the classical K-Means on each batch of data (i.e. on each process-local chunk of data) individually and in parallel.
```
ClaudiaComito, line 327:
```suggestion
    This method performs K-Means clustering on each batch (i.e. on each process-local chunk) of data individually and in parallel.
```

ClaudiaComito (changes requested):
Looks great to me @mrfh92 , I only have a few edits to the documentation. Otherwise we can merge.

mrfh92:
@ClaudiaComito I have committed all suggested changes.

ClaudiaComito (approved):
:clap: @mrfh92 !

PR #1331, opened 2024-01-22
Bugs/1330  bug continuous benchmarks fail for qr due to deprecated value for tiles per proc

mrfh92:
* Since we deprecated the option tiles_per_proc=1 in QR, this should also to be removed from the continuous benchmarks 
* the warning for tiles_per_proc=1 for split=0 is now implemented crash-free


mrfh92:
for some reason cb/report hangs although it looks like that they have been executed on codebase

mtar (approved):

PR #1315, opened 2024-01-02
Fix some NumPy deprecations in the core tests.

FOsterfeld:
## Due Diligence

- General:
    - [x]  **base branch** must be `main` for new features, latest release branch (e.g. `release/1.3.x`) for bug fixes
    - [x]  **title** of the PR is suitable to appear in the [Release Notes](https://github.com/helmholtz-analytics/heat/releases/latest)
- Implementation:
    - [x] unit tests: all split configurations tested
    - [x] unit tests: multiple dtypes tested
    - [x] documentation updated where needed

## Description

Issue/s resolved: #

## Changes proposed:

-
-
-
-

#### Does this change modify the behaviour of other functions? If so, which?
yes / no

PR #1337, opened 2024-02-01
Bugs/1336  NumPy-backward compatibility of statistics tests

### LINKED ISSUE #1336: [Bug]: Bug in tests of statistics
Failure of the CI (also on main): 

```
heat/core/tests/test_statistics.py::TestStatistics::test_percentile - TypeError: _percentile_dispatcher() got an unexpected keyword argument 'method'
```

Problem is that after #1315 our tests only run on new numpy versions but it looks like that we currently use an old one...

mrfh92:
resolves #1336 

Background: the changes in #1315 let our tests fail on old numpy versions


mtar (approved):
Thank you very much!

PR #1348, opened 2024-02-06
Features/1347 bundle several bump actions

mrfh92:
This PR collects the PRs #1342-#1346 created by dependabot

### DIFF
--- a/.github/workflows/CommentPR.yml
+++ b/.github/workflows/CommentPR.yml
@@ -21,7 +21,7 @@ jobs:
           egress-policy: audit
 
       - name: 'Download artifact'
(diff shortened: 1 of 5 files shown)


mrfh92:
btw: dependabot doesnt like that we combine its PRs into another one

ClaudiaComito:
> btw: dependabot doesnt like that we combine its PRs into another one :D

Yeah, let's not do this too often. I think the time saving is minimal, and there will be instances when one of the updates fails and we only find out after approving, see the setup-mpi update #1351 .

ClaudiaComito:
Closing this

Issue #1321, opened 2024-01-11
[Bug]: factories.array illegal parameter combinations copy & split


mtar:
### What happened?

ht.array creates a new DNDarray on the CPU when the arguments copy = False and split >= 0 are passed. This should not happen.
In addition, copy = None returns the same DNDarray which doesn't infer the device.

### Code snippet triggering the error

```shell
import torch
import heat as ht

m,n = 40,20

a = torch.randn(m, n, dtype=torch.double, device='cuda')
b = ht.array(a, copy=False, split=0, dtype=ht.double)

print('a.device =', a.device)
print('b.device =', b.device)
print('b.split =', b.split)
```

### Error message or erroneous outcome

```shell
a.device = cuda:0
b.device = cpu:0
b.split = 0

a.device = cuda:0
b.device = cpu:0
b.split = 0
```

### Version

1.3.x

### Python version

None

### PyTorch version

None

PR #1363, opened 2024-02-10
Fix: `ht.array` constructor respects implicit torch device when copy is set to false

JuanPedroGHM:
## Due Diligence
- General:
- Implementation:

## Description

`ht.array` constructor deleted original object torch device when the copy flag is set to False. 

Issue/s resolved: #1321 

## Changes proposed:

- Earlier check in `ht.array` factory for the original object's device. 
- Added a test for `ht.array` using torch tensor, on different devices (when changing the env variable), and with copy set to false.

## Type of change

yes?


JuanPedroGHM:
So I did a bit of a brute force grid search and found all this instances that to me are incorrect behavior from heat. Let me know what you think:

Current inconsistent behavior (Summary):

- When creating from a {list, numpy, torch}, if **is_split** is set, and **split** is not, the **split** property of the new array is set to that split, but the array is not actually split.
    - This also happens for heat, when the original array is not split.
    - Expected result: the **is_split** argument should be ignored, as they are not split originally. New **DNDArray** should have the same split indicated as the **split arg**
- When both **split** and **is_split** are set, it always throws an exception. Could be changed to ignore in some cases, for list, numpy, torch arrays for example, as they are not split by default.
- When the original obj is a heat DNDArray that has already been split, the behavior is very inconsistent.
    - The **split** property is incorrect a lot of times, even if the split is done right.
    - When the original array is split, and the argument **split** of **ht.array** is set, the new array is split a second time.  Probably not what we want.
- When the og object is a heat dndarray, the split/is_split exclusivity can be skipped as long as all the properties from the old and new array match.
- No proper test with multiple devices yet, Horeka and Haicore are unavailable.

Attached is also a full table with the data. 

[bug_results.md](https://github.com/helmholtz-analytics/heat/files/14375318/bug_results.md)

JuanPedroGHM:
Changes:

- Raised exceptions when using the `ht.array` method would resplit the existing and already split DNDArray. If the original DNDArray is not split (`obj.split is None`), using `ht.array` will return a split object. This is to ensure compatibility with the rest of the heat code. 
- The check if the arguments `split` and `is_split` was moved higher to ensure it cannot be circumvented with the right choice of arguments. 
- Added additional tests that check those combinations of arguments.

heat/core/tests/test_factories.py
```diff
@@ -200,6 +200,64 @@ def test_array(self):
...
+        self.assertEqual(dndarray_2d.dtype, ht.float64)
+        self.assertEqual(dndarray_2d.gshape, (3, 3))
+        self.assertEqual(len(dndarray_2d.lshape), 2)
+        self.assertLessEqual(dndarray_2d.lshape[0], 3)
+        self.assertEqual(dndarray_2d.lshape[1], 3)
+        self.assertEqual(dndarray_2d.split, 0)
+        self.assertTrue(
+            (
+                dndarray_2d.larray == torch.tensor([1.0, 2.0, 3.0], device=self.device.torch_device)
+            ).all()
```
mtar, line 220: This only checks if the values are the same. We are more interested whether `larray` uses the same memory as `array_2d` when `copy` is set to `False`.

mrfh92 (dismissed):
From my point of view this looks fine; in particular, I guess that the additional error messages will be very helpful (in particular since I have already made a mistake regarding split and is_split myself). 
Thanks 👍

ClaudiaComito (changes requested):
Thanks a lot @JuanPedroGHM .  

The CI matrix must be updated here before we can merge.

Also, just to be safe, can you test that the tutorials still behave as expected after these changes? [Internals](https://github.com/helmholtz-analytics/heat/blob/main/tutorials/local/3_internals.ipynb) would be a good one to check (also feel free to add content there, I don't think we've addressed the `copy` attribute at all).

Thanks again!

JuanPedroGHM:
> Thanks a lot @JuanPedroGHM .
> 
> The CI matrix must be updated here before we can merge.
> 
> Also, just to be safe, can you test that the tutorials still behave as expected after these changes? [Internals](https://github.com/helmholtz-analytics/heat/blob/main/tutorials/local/3_internals.ipynb) would be a good one to check (also feel free to add content there, I don't think we've addressed the `copy` attribute at all).
> 
> Thanks again!

Thanks for the suggestion @ClaudiaComito ! I ran the tutorial (on my laptop, so single gpu and without the datasets) and they ran perfectly. I did find one error in the internals tutorial with an undefined variable "target_map", which I corrected with the last commit.

ClaudiaComito (approved):
Thanks again, next try...

Issue #1385, opened 2024-02-27
[Bug]: versioning on main branch


ClaudiaComito:
### What happened?

Version on `main` is 1.4.1-dev, should be 1.4.0-dev. This a relic of the latest import of fixes from `release/1.3.x`, hopefully the new backporting workflow will prevent it from now on.

### Version

main (development branch)

### Python version

None

### PyTorch version

None

PR #1386, opened 2024-02-27
Set correct version on `main`, update Bug report template

ClaudiaComito:
## Due Diligence

- General:
- Implementation:

## Description

Issue/s resolved: #1385 

## Changes proposed:

- set version back to 1.4.0-dev on `main`
- updated PyTorch, Python versions, removed outdated ones in Bug report template
-
-

## Type of change

- Documentation update
- 
## Memory requirements

NA
## Performance

NA
no


mrfh92 (approved):
Looks fine from my point of view. CI failure seems to be related to sth different (happens occasionally also for other PRs, restart usually solves the problem)

Issue #900, opened 2022-01-17
Open multiple files into a single DNDarray


coquelin77:
**Feature functionality**
load multiple files into a single DNDarray.
- each process takes a number of files (sequential) and loads them as a single torch tensor
    - shape : `[file, (dimes)]`
- balance the DNDarray at the

mrfh92:
@krajsek (tagged you because I closed old #740 where you were assigned)

mrfh92:
reviewed and updated within #1109

mrfh92:
(assigned to me as reservation for @Reisii)

PR #1388, opened 2024-02-28
Load functionality for multiple .npy files

Reisii:
## Due Diligence

- General:
- Implementation:

## Description
Many smaller .npy-files will be loaded at the same time 
Added to one big DNDarray 

Issue/s resolved: #900 

## Changes proposed:

-
-
-
-

## Type of change
New feature

 no


heat/core/manipulations.py
```diff
@@ -1,6 +1,7 @@
 """
 Manipulation operations for (potentially distributed) `DNDarray`s.
 """
+
```
mrfh92, line 4: I guess this has been added unintentionally

heat/core/tests/test_io.py
```diff
@@ -739,3 +740,20 @@ def test_save_netcdf_exception(self):
...
+        # Abc
+        crea_array = np.random.randint(1024, size=(random.randint(1, 100), 5, 11))
+        np.save("data0", crea_array)
+
+        for i in range(1, 100):
+            x = np.random.randint(1024, size=(random.randint(1, 100), 5, 11))
+            crea_array = np.concatenate(x, 0)
+            np.save("data" + str(i), x)
+
+        array1 = ht.load_npy_from_path("/home/nguy_t4/home/heat/heat/core/tests")
```
mrfh92, line 754: Der Pfad wird natürlich nicht gefunden, da die Tests auf den Runnern nicht auf deinem Rechner laufen... du musst hier einen relativen Pfad angeben, vermutlich `/heat/core/tests` o.ä.

mrfh92:
As the error 
```
ERROR: test_load_npy (heat.core.tests.test_io.TestIO)
  File "/home/runner/work/heat/heat/heat/core/tests/test_io.py", line 760, in test_load_npy
    self.assertEqual(load_array.numpy(), int_array)
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/unittest/case.py", line 902, in _baseAssertEqual
    if not first == second:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
tells, comparing two numpy arrays `a` and `b` results in an ambiguous result as an array with the same shape as `a` and `b` but with entries `True` or `False` is returned and, in general, it is difficult to assign a single `True`/`False` to such an outcome. 

In our case, however, `True` should be the outcome if and only if the two arrays coincide in every element. Thus, the `.all()` function needs to be used as follows: 
```
self.assertTrue((a==b).all())
```
I have made these changes.

mrfh92 (approved):
From my point of view this looks fine. 
Thanks for the work 👍

heat/core/io.py
```diff
@@ -1132,3 +1141,52 @@ def save(
...
+
+
+def load_npy_from_path(
+    path: str,
+    dtype: datatype = types.int32,
+    split: Optional[int] = None,
+    device: Optional[str] = None,
+    comm: Optional[Communication] = None,
+):
+    """
```
mrfh92, line 1153: Docstring missing

mrfh92 (changes requested):
A docstring is missing so far. 
Except for that it looks fine from my point of view.

heat/core/tests/test_io.py
```diff
@@ -739,3 +742,62 @@ def test_save_netcdf_exception(self):
...
+        # testing for int arrays
+        if ht.MPI_WORLD.rank == 0:
+            crea_array = []
+            for i in range(0, 20):
+                x = np.random.randint(1000, size=(random.randint(0, 30), 6, 11))
+                np.save(os.path.join(os.getcwd(), "heat/datasets", "int_data") + str(i), x)
+                crea_array.append(x)
+            int_array = np.concatenate(crea_array)
+        else:
+            time.sleep(2)
```
mrfh92, line 756: ht.MPI_WORLD.Barrier() can be used too

heat/core/tests/test_io.py
```diff
@@ -739,3 +742,62 @@ def test_save_netcdf_exception(self):
...
+        # testing for float arrays and split dimension other than 0
+        if ht.MPI_WORLD.rank == 0:
+            crea_array = []
+            for i in range(0, 20):
+                x = np.random.rand(2, random.randint(1, 10), 11)
+                np.save(os.path.join(os.getcwd(), "heat/datasets", "float_data") + str(i), x)
+                crea_array.append(x)
+            float_array = np.concatenate(crea_array, 1)
+        else:
+            time.sleep(2)
```
mrfh92, line 780: also with Barrier() (see above)

heat/core/io.py
```diff
@@ -1132,3 +1141,52 @@ def save(
...
 DNDarray.save.__doc__ = save.__doc__
+
+
+def load_npy_from_path(
+    path: str,
+    dtype: datatype = types.int32,
+    split: Optional[int] = None,
+    device: Optional[str] = None,
+    comm: Optional[Communication] = None,
+):
```
mrfh92, line 1152: return type is missing

heat/core/io.py
```diff
@@ -1132,3 +1141,52 @@ def save(
...
+
+    process_number = MPI_WORLD.size
+    file_list = []
+    for file in os.listdir(path):
+        if fnmatch.fnmatch(file, "*.npy"):
+            file_list.append(file)
+    n_files = len(file_list)
+
+    if n_files == 0:
+        raise ValueError("No NPY Files were found")
```
mrfh92, line 1169: .npy

heat/core/io.py
```diff
@@ -1132,3 +1141,89 @@ def save(
...
+        Data type of the resulting array.
+    split : int or None : optional
+        Along which axis the resulting array should be split.
+        Default is ``None`` which means each node will have the full array.
+    device : str, optional
+        The device id on which to place the data, defaults to globally set default device.
+    comm : Communication, optional
+        The communication to use for the data distribution, defaults to global default
+
+    Raises
```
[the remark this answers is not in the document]
Reisii: There actually is, for example for the load_csv() function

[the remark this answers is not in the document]
mrfh92: yes, but this does load a single csv-file without using any optimized library for loading; therefore, its not clear what the performance implications are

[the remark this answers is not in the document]
mrfh92: sry, I misunderstood this comment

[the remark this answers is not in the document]
mrfh92: then its fine

heat/core/io.py
```diff
@@ -1131,3 +1140,65 @@ def save(
 
 DNDarray.save = lambda self, path, *args, **kwargs: save(self, path, *args, **kwargs)
 DNDarray.save.__doc__ = save.__doc__
+
+
+def load_npy_from_path(
+    path: str,
+    dtype: datatype = types.int32,
+    split: int = 0,
```
mrfh92, line 1148: I would remove the default 0 here and make split a mandatory argument

heat/core/io.py
```diff
@@ -1131,3 +1140,65 @@ def save(
...
+        Along which axis the loaded arrays should be concatenated.
+    device : str, optional
+        The device id on which to place the data, defaults to globally set default device.
+    comm : Communication, optional
+        The communication to use for the data distribution, default is 'heat.MPI_WORLD'
+    """
+    if not isinstance(path, str):
+        raise TypeError(f"path must be str, not {type(path)}")
+    elif split is not None and not isinstance(split, int):
+        raise TypeError(f"split must be None or int, not {type(split)}")
```
mrfh92, line 1171: consequently, this needs to be adapted

heat/core/tests/test_io.py
```diff
@@ -739,3 +742,67 @@ def test_save_netcdf_exception(self):
...
+            ht.load_npy_from_path("heat/datasets", split="ABC")
+        with self.assertRaises(ValueError):
+            ht.load_npy_from_path(path="heat", dtype=ht.int64, split=0)
+        if ht.MPI_WORLD.size > 1:
+            if ht.MPI_WORLD.rank == 0:
+                x = np.random.rand(2, random.randint(1, 10), 11)
+                np.save(os.path.join(os.getcwd(), "heat/datasets", "float_data"), x)
+            ht.MPI_WORLD.Barrier()
+            with self.assertRaises(RuntimeError):
+                ht.load_npy_from_path("heat/datasets/npy_dummy", dtype=ht.int64, split=0)
```
Reisii, line 806: You are loading the dummy file from the wrong directory or you saved it in the wrong directory

mrfh92: thx for the hint 👍

mrfh92: now it should work again

mrfh92 (dismissed):
approve to let full CI-matrix run (test combinations of different Python and PyTorch versions)

mrfh92:
@mtar if you have time can you look into this PR and review? (to follow the convention not to review own students' work)

mrfh92 (approved):
approve again to let CI-matrix run

heat/core/io.py
```diff
@@ -1131,3 +1140,65 @@ def save(
...
+
+    local_list = [
+        file_list[i]
+        for i in range(
+            rank * (n_files // process_number), rank * (n_files // process_number) + n_for_procs
+        )
+    ]
+    array_list = []
+    for element in local_list:
+        array_list.append(np.load(path + "/" + element))
```
mtar, line 1199:
```suggestion
    n_for_procs = n_files // process_number
    idx = rank * n_for_procs
    if rank + 1 == process_number:
        n_for_procs += (n_files % process_number)
    array_list = [np.load(path + "/" + element) for element in file_list[idx : idx+n_for_procs]]
```

mrfh92: @Reisii du kannst die Empfehlung übernehmen, indem du auf "Commit suggestion" gehst. Dann geht das autlmatisch.

mrfh92 (approved):
approve to let CI run; also @mtar has reviewed yesterday already

Issue #1400, opened 2024-03-11
Implement unfold-operation similar to `torch.Tensor.unfold`


mrfh92:
**Feature functionality**
The idea is to introduce a function `ht.unfold` in the manipulations-module that has the same functionality as [PyTorchs unfold](https://pytorch.org/docs/stable/generated/torch.Tensor.unfold.html). 
In fact, this only requires additional efforts for the case `split=dimension` (`dimension` is the axis along which unfold takes place); otherwise you can directly apply the torch-function to the local arrays. 

*Hint:* some auxiliary stuff required in the "difficult" case may already be implemented under the name "`halo`". 

**Additional context**
With the help of such an operation it will be possible to generate sliding patches of an data set, e.g., to generate from an image (2/3d-array) the 4/5d-array of all size x size sub-images etc.

mrfh92:
@github-actions bot: it is planned, because someone is actively working on it!

PR #1419, opened 2024-04-02
Implement distributed `unfold` operation

FOsterfeld:
## Due Diligence

- General:
- Implementation:

## Description

Add the function unfold to the available manipulations. unfold(a, dimension, size, step) for a DNDarray a behaves like [torch.Tensor.unfold](https://pytorch.org/docs/stable/generated/torch.Tensor.unfold.html).

Example:
```
>>> x = ht.arange(1., 8)
>>> x
DNDarray([1., 2., 3., 4., 5., 6., 7.], dtype=ht.float32, device=cpu:0, split=e)
>>> ht.unfold(x, 0, 2, 1)
DNDarray([[1., 2.],
          [2., 3.],
          [3., 4.],
          [4., 5.],
          [5., 6.],
          [6., 7.]], dtype=ht.float32, device=cpu:0, split=None)
>>> ht.unfold(x, 0, 2, 2)
DNDarray([[1., 2.],
          [3., 4.],
          [5., 6.]], dtype=ht.float32, device=cpu:0, split=None)
```

Issue/s resolved: #1400 

## Changes proposed:

-
-
-
-

## Type of change

no


mrfh92:
The tests on the CUDA-runner seem to hang at `test_manipulations.py` for 5 MPI-processes

mrfh92:
On the Terrabyte cluster, using 8 processes on 2 nodes with 4 GPUs each I get the following error: 
```
ERROR: test_unfold (heat.core.tests.test_manipulations.TestManipulations)
  File "/dss/dsshome1/03/di93zek/heat/heat/core/tests/test_manipulations.py", line 3775, in test_unfold
    ht.unfold(x, 0, min_chunk_size, min_chunk_size + 1)  # no fully local unfolds on some nodes
  File "/dss/dsshome1/03/di93zek/heat/heat/core/manipulations.py", line 4272, in unfold
    ret_larray = torch.cat((unfold_loc, unfold_halo), dimension)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument tensors in method wrapper_CUDA_cat)

Ran 32 tests in 26.574s
```

ClaudiaComito (changes requested):
@FOsterfeld thanks a lot for this. Can you write a description of the PR? What is the function supposed to do, maybe a small example? Thanks!

FOsterfeld:
@ClaudiaComito I have now added a PR description and tests with different datatypes.

ClaudiaComito (changes requested):
Thanks a lot @FOsterfeld for this PR, this is going to be really useful. I have a few change requests, mostly about simplifying the code by using existing methods. Thanks a lot!

heat/core/manipulations.py
```diff
@@ -4178,3 +4179,97 @@ def mpi_topk(a, b, mpi_type):
...
+                            dimension * (slice(None, None, None),)
+                            + (slice(max_index, None, None), Ellipsis)
+                        ],
+                        a.halo_next[
+                            dimension * (slice(None, None, None),)
+                            + (slice(None, rem, None), Ellipsis)
+                        ],
+                    ),
+                    dimension,
+                ).unfold(dimension, size, step)
```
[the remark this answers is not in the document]
FOsterfeld: I think this might not really make things easier because as far as I understand it I would then have to make a case distinction to remove the ´halo.prev´ for the nodes in the middle. Is there any way to only append the ´halo.next´?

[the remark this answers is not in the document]
ClaudiaComito: does #1509 help?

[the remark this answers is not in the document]
FOsterfeld: That should be exactly what I needed, I'll merge it into this PR to test it out.

mrfh92:
@FOsterfeld there seems to be an error now on the CUDA runner. As it fails in unfold, its maybe not a random-CI-error due to overloaded runners but really sth in unfold

FOsterfeld:
There seems to be something wrong with the communication in DNDarray.get_halo(). Sometimes the halo that is sent from the last rank to the rank before is faulty. This happened irregularly in my tests without any randomization in the data, so maybe it occurs depending on the order in which the non-blocking halo-sends are fulfilled there.

In 825979ce6f924fc29ada6634cdfc372f2c873b9e I tested get_halo(prev=False) with blocking sends instead, this eliminated all errors but is obviously no final solution to the problem.

ClaudiaComito:
> There seems to be something wrong with the communication in DNDarray.get_halo(). Sometimes the halo that is sent from the last rank to the rank before is faulty. This happened irregularly in my tests without any randomization in the data, so maybe it occurs depending on the order in which the non-blocking halo-sends are fulfilled there.
> 
> In [825979c](https://github.com/helmholtz-analytics/heat/commit/825979ce6f924fc29ada6634cdfc372f2c873b9e) I tested get_halo(prev=False) with blocking sends instead, this eliminated all errors but is obviously no final solution to the problem.

Hi @FOsterfeld , thanks for all the debugging. I too ran a few tests and indeed with the current test implementation, blocking-send is necessary.

The test that fails is a loop of loops in which halos are constantly collected and rewritten. I can't prove it, but I suspect that many Heat operations that rely on non-blocking communication will fail such a test. 

The test fails at random positions in the loop even with constant number of processes, for example in one case it fails at:
```
split: 0, size: 8, step: 3, dimension: 0, equal: False
```
next run it will fail at a different spot.

But testing those specific failing combinations, isolated or in a smaller loop, on the same number of processes, works fine.

Do we anticipate the need of collecting halos in a big loop? If yes, by all means let's use blocking-sends (for both `prev` and `next` cases). What do you think @FOsterfeld ?

ClaudiaComito (changes requested):
See [yesterday's comment ](https://github.com/helmholtz-analytics/heat/pull/1419#issuecomment-2217164012)- if we expect halos to be collected in big loops, we're justified to use blocking send (please change both in `prev` and `next` case).

Otherwise, we stick to non-blocking send but have to write more realistic tests.

Thanks for all the work @FOsterfeld !

FOsterfeld:
@ClaudiaComito Do we know any reason why the non-blocking send will sometimes fail for these kinds of loops where we often send the halos?

As this is a subtle bug that just silently corrupts random data when it occurs, I don't think that we should just ignore it. If we cannot fix the non-blocking send, I could implement the halo send with non-blocking sends in two phases. In the first phase all even nodes send their halo and the odd ones receive it and in the second phase these roles reverse. This should use constant time, whereas my last quickfix would need linear time in the number of nodes.

mrfh92:
@ClaudiaComito @mtar @JuanPedroGHM @FOsterfeld 
I tried to replace the `Send` in the "`next`"-block of `get_halo` by an `Isend` in order to avoid deadlocks etc. I have tested on a varying number of processes and several times each on our workstation. It worked. Hopefully, this resolves the problem for the moment. 
I dont think that synchronization can be completely avoided when using halos, as relying on halos intrinsically requires adjacent processes to be synchronized; as all processes take part, this results in at least some kind of global synchronization.

mrfh92:
@FOsterfeld the issue regarding the problems with the communication of halos should be resolves now as we arrived at an error-free constellation; whether performance can be optimized should be addressed within the corresponding issue #1570 at a later point in time. 

However, did you already include an error thrown for size=1? If yes, the PR is ready for merging from my point of view.

FOsterfeld:
@mrfh92 I have now added the error for the case that size=1. I could also verify that the synchronization errors that caused data corruption do not occur anymore, so this PR should be ready for merging.

mrfh92:
@FOsterfeld from my point of view this now looks fine 
@ClaudiaComito do you agree?

mrfh92 (approved):
Looks fine from my point of view. @FOsterfeld Thanks :+1:

heat/core/tests/test_manipulations.py
```diff
@@ -3752,3 +3752,77 @@ def test_vstack(self):
...
+                                diff_indices = ht.nonzero((u - v).resplit_(None)).resplit_(None)
+                                if y.comm.rank == 0:
+                                    print(
+                                        f"\ndtype: {dtype}\nsplit: {split}\ndimension: {dimension}\nstep: {step}\nsize: {size}"
+                                    )
+                                    # print(f"u.shape: {u_shape}")
+                                    # print(f"v.shape: {v_shape}")
+                                    print(f"diff: {diff}")
+                                    print(f"displs: {y.counts_displs()[1]}")
+                                print(f"diff_indices: {diff_indices}")
```
ClaudiaComito, line 3826: Do we still need to include this?

heat/core/dndarray.py
```diff
@@ -433,25 +437,30 @@ def get_halo(self, halo_size: int) -> torch.Tensor:
...
 
-            if rank != first_rank:
-                self.comm.Isend(a_prev, prev_rank)
-                res_next = torch.zeros(
-                    a_next.size(), dtype=a_next.dtype, device=self.device.torch_device
-                )
-                req_list.append(self.comm.Irecv(res_next, source=prev_rank))
+            if next:
+                if rank != first_rank:
+                    # self.comm.Send(a_prev, prev_rank)
```
ClaudiaComito, line 451: dead code?

ClaudiaComito (changes requested):
@FOsterfeld @mrfh92 this looks great, I only found some (presumably) dead code that can be removed, otherwise I think it can be merged. Thanks a lot!

heat/core/tests/test_manipulations.py
FOsterfeld: That was code I used for debugging so it can be removed.

heat/core/dndarray.py
FOsterfeld: Yes, that was the old quick fix for the bug in get_halo().

ClaudiaComito (approved):
Great job @FOsterfeld !

Issue #1414, opened 2024-03-27
[Bug]: Scorecard analysis failure


ClaudiaComito:
### What happened?

"Merger of the Week" @mrfh92 reports that the Scorecard analysis  workflow [fails on `main`](https://github.com/helmholtz-analytics/heat/actions/runs/8418158944/job/23048080154) after merging one of the `dependabot` updates of the week.

### Version

main (development branch)

### Python version

None

### PyTorch version

None

mrfh92:
This bug still exists as running the job again still results in failure

mrfh92:
```
2024/03/27 06:45:31 error signing scorecard json results: error signing payload: getting key from Fulcio: verifying SCT: updating local metadata and targets: error updating to TUF remote mirror: invalid key
remote status:{
	"mirror": "https://sigstore-tuf-root.storage.googleapis.com/",
	"metadata": {
		"root.json": {
			"version": 9,
			"len": 6766,
			"expiration": "12 Sep 24 06:53 UTC",
			"error": ""
		},
		"snapshot.json": {
			"version": 133,
			"len": 2302,
			"expiration": "16 Apr 24 16:08 UTC",
			"error": ""
		},
		"targets.json": {
			"version": 9,
			"len": 5478,
			"expiration": "12 Sep 24 06:13 UTC",
			"error": ""
		},
		"timestamp.json": {
			"version": 171,
			"len": 719,
			"expiration": "02 Apr 24 16:08 UTC",
			"error": ""
		}
	}
}
```

This might be the corresponding error message but I am not sure whether this is really the point of failure

matthyx:
hi, we have the same issue since last Friday on https://github.com/kubescape/kubescape

mrfh92:
This happened after #1407 had been merged

PR #1421, opened 2024-04-04
Update actions in scorecard.yml

mtar:
## Due Diligence

- General:
- Implementation:

## Description

Upgrades the actions used in scorecard.yml.
**The actual changes can't be tested beforehand because this action only works on main.**

Issue/s resolved: #1414

## Changes proposed:

- update scorecard.yml

## Type of change
- github action

## Memory requirements

n/a

## Performance

n/a

no


mrfh92 (approved):
Looks fine from my point of view. 
Thanks 👍

Issue #1289, opened 2023-12-05
[Bug]: CI matrix fails with new `setup-mpi` 1.2.1


ClaudiaComito:
### What happened?

The Python version vs. PyTorch version cross-checks we run after a PR is approved are failing. Something has changed in the jobs set up that we couldn't pinpoint yet.

### Code snippet triggering the error

```shell
ci.yaml action
```

### Error message or erroneous outcome

```shell
Tests fail at test_random.py
```

### Version

1.3.x

### Python version

3.10

### PyTorch version

1.13

### MPI version

```shell
Don't know.
```

mtar:
A few configurations of the test matrix complete with AssertionErrors. After removing `test_rand`, `test_randint`, and `test_randn`, all jobs complete.

```shell
=================================== FAILURES ===================================
_____________________________ TestRandom.test_rand _____________________________

self = <heat.core.tests.test_random.TestRandom testMethod=test_rand>

    def test_rand(self):
        # int64 tests
    
        # Resetting seed works
        seed = 12345
        ht.random.seed(seed)
        a = ht.random.rand(2, 5, 7, 3, split=0)
        self.assertEqual(a.dtype, ht.float32)
        self.assertEqual(a.larray.dtype, torch.float32)
        b = ht.random.rand(2, 5, 7, 3, split=0)
        self.assertFalse(ht.equal(a, b))
        ht.random.seed(seed)
        c = ht.random.rand(2, 5, 7, 3, dtype=ht.float32, split=0)
        self.assertTrue(ht.equal(a, c))
    
        # Random numbers with overflow
        ht.random.set_state(("Threefry", seed, 0xFFFFFFFFFFFFFFF0))
        a = ht.random.rand(2, 3, 4, 5, split=0)
        ht.random.set_state(("Threefry", seed, 0x10000000000000000))
        b = ht.random.rand(2, 44, split=0)
        a = a.numpy().flatten()
        b = b.numpy().flatten()
        self.assertEqual(a.dtype, np.float32)
        self.assertTrue(np.array_equal(a[32:], b))
    
        # Check that random numbers don't repeat after first overflow
        seed = 12345
        ht.random.set_state(("Threefry", seed, 0x100000000))
        a = ht.random.rand(2, 44)
        ht.random.seed(seed)
        b = ht.random.rand(2, 44)
        self.assertFalse(ht.equal(a, b))
    
        # Check that we start from beginning after 128 bit overflow
        ht.random.seed(seed)
[...]

PR #1443, opened 2024-04-18
Bugfix: `_like` factories now default to same device as input DNDarray 

mrfh92:
## Due Diligence

- General:
- Implementation:

## Description

Issue/s resolved: #1464 

## Changes proposed:
**Edited by @ClaudiaComito on Apr 30**

- changed default device in `factories.__factory_like(a)`, now device defaults to `a.device` unless explicitly set 
- addressed pre-commit complaints 
- removed unnecessary/legacy `numpy()` calls from `test_random` to facilitate testing on AMD runner
-

## Type of change

- Documentation update

## Memory requirements

NA
## Performance

`test_random` now takes a bit longer as it's running statistical functions on distributed DNDarrays instead of on the gathered numpy arrays
no


mrfh92:
First bug fix for v1.4  (unfortunately we need to release this as it makes problems in one of our benchmarks) 

background: this can't be found in our tests as `HEAT_TEST_USE_DEVICE=gpu` seems to set the default for devices to "gpu", whereas in the standard mode we have default = "cpu" at many places ... therefore, errors of this kind are only found when working with device="gpu" manually in a code

ClaudiaComito:
Error in `random` module in AMD runner, similar but not same as #1289

ClaudiaComito:
@mrfh92 it should work now. I don't know why pre-commit fails.

mtar (approved):

PR #1504, opened 2024-06-01
Add solver for triangular systems

FOsterfeld:
Copy of the previous PR for solve_triangular #1236 .

to add the context: we had already merged this PR, but then some error with the CI occurred and undid the changes (which were, however, not responsible for this error as we found out later)


mrfh92:
to add the context: we had already merged this PR, but then some error with the CI occurred and undid the changes (which were, however, not responsible for this error as we found out later)

ClaudiaComito (approved):
Thanks for resubmitting!

ClaudiaComito (changes requested):
@FOsterfeld apologies, I just noticed the lines that aren't covered by tests, can we increase the test coverage?

ClaudiaComito (approved):
Great @FOsterfeld thanks for the changes!

PR #1508, opened 2024-06-03
Features/1448 Refactor random module

### LINKED ISSUE #1448: Allow random arrays of arbitrary size / speed up random arrays
The current very elaborate implementation of our random generator and random arrays suffers from two drawbacks: 
* the maximum number of entries is quite low (~maxint32 / maxint64 for the respective data types) 
* it is comparatively slow. 

I suggest to consider something like the following: 
```
global_seed = <...> 
local_seed = global_seed + ht.MPI_WORLD.rank 
local_rand = torch.rand(local_shape, ...) # with local_seed as seed 
global_array = ht.array(local_rand, ...) 
```
To discuss: 
* does this destroy the "quality" of the resulting pseudo-random array? (I know, as a trained mathematician I should be able to answer that question, but at the moment I can't) 
* even if the quality might be lower in the end, I suggest to add this version of random arrays, e.g., with a keyword argument `quick_and_dirty=True` or something like this

mrfh92:
## Due Diligence

- General:
- Implementation:

## Description
Issue/s resolved: #1448 

## Changes proposed:
- default RNG is now "Batchparallel", i.e. choose a global seed, get process-local seeds by "global seed + rank" on each MPI-rank, use torch's RNG with these local seeds 
- existing Threefry RNG can be used as user-specified option using set_state(...). 

## Type of change
should not break existing applications as far as they do not rely on strong reproducibility ensured by Threefry but not batchparallel RNG

## Performance
likely to improve performance of random number generation compared to existing solution

yes, everything related to or using randomness in Heat


mrfh92:
(The error in the other random functions is due to the fact that the way of setting the seed in my new implementation interferes with setting the seed in the other functionality. This would have to be removed if we want to keep the new idea.)

mrfh92:
still to be done: figure out influence of proposed changes on other functions

heat/core/random.py
```diff
@@ -109,8 +129,8 @@ def __counter_sequence(
         c_0 = (__counter & (max_count << 64)) >> 64
     c_1 = __counter & max_count
     total_elements = torch.prod(torch.tensor(shape))
-    if total_elements.item() > 2 * max_count:
-        raise ValueError(f"Shape is to big with {total_elements} elements")
+    # if total_elements.item() > 2 * max_count:
+    #    raise ValueError(f"Shape is to big with {total_elements} elements")
```
mtar, line 133: Commented-out code. Not needed anymore?

heat/core/random.py
```diff
@@ -451,30 +484,48 @@ def rand(
...
 
-        # combine the values into one tensor and convert them to floats
-        values = __int32_to_float32(torch.stack([x_0, x_1], dim=1).flatten()[lslice]).reshape(
-            lshape
-        )
-    elif dtype == types.float64:
-        x_0, x_1, lshape, lslice = __counter_sequence(shape, torch.int64, split, device, comm)
-        x_0, x_1 = __threefry64(x_0, x_1, seed=__seed)
+    if __rng == "Threefry":
+        # use Threefry RNG (note: batchparallel RNG comes at the very end in only line only)
```
mtar, line 489: What does the note say? I don't get it.

heat/core/random.py
```diff
@@ -451,30 +484,48 @@ def rand(
...
-        )
+        return DNDarray(values, shape, dtype, split, device, comm, balanced)
     else:
-        # Unsupported type
-        raise ValueError(f"dtype is none of ht.float32 or ht.float64 but was {dtype}")
-
-    return DNDarray(values, shape, dtype, split, device, comm, balanced)
+        # use batchparallel RNG
+        x = factories.__factory(
+            shape if shape != () else (1,),
```
mtar, line 516:
```suggestion
            shape if len(shape) > 0 else (1,),
```
easier to understand

heat/core/random.py
```diff
@@ -451,30 +484,48 @@ def rand(
...
+            split if split is not None else 0,
+            torch.rand,
+            device,
+            comm,
+            "C",
+        )
+        if split is None:
+            x = x.resplit_(None)
+        if shape == ():
+            x = x.reshape(shape)
```
mtar, line 527: Can the support for the special cases be integrated into __factory such that the arrays doesn't have to be changed back and forth?

heat/core/random.py
mrfh92: actually, this is still needed and I forgot to comment in again
Thx :+1:

heat/core/random.py
mrfh92: it was a hint that the second case (batchparallel RNG) was one line only (before black) and could easily be overlooked therefore 
I removed the comment

heat/core/random.py
mrfh92: actually, I would like to avoid doing this as the reason for this workaround is not rooted in `__factories` but in the specific nature of batchparallel RNG: 

`__factories` creates a DNDarray with the local arrays being created by the input function, i.e., in our case `torch.randn`. Since torch's random seed is different (and must be different) on each MPI process for batchparallel RNG, this would result in different local arrays across processes in the case `split=None`; hence, there are three options, none of them really cool: 

*  save the states of each of the process-local torch RNGs, synchronize their states, call `__factories`, give the process-local RNGs back their unsynchronized state (I dont even know whether that works...) 

* don't use `__factories`: use torch's RNG only on one process (say process 0) and broadcast the resulting arrays to all other processes; manually generate a non-split DNDarray from these local arrays 

* tell `__factories` to create the desired array, but with artificial split=0. Then resplit. 

In my opinion, the last variant (as implemented) is the most easy one; in particular, since the second-last variant also involves communication between all processes.

mrfh92: @mtar do you agree?

mtar:
few lines missing coverage

mrfh92:
@mtar now we have 100% coverage of patch

mtar (approved):
Looks fine to me :ok_hand:

mtar (changes requested):
Checks fail on older PyTorch versions < 2.0. The minimum version needs to be raised first. See also #1572

mrfh92:
@mtar I have created #1573 to solve this problem.

mtar:
Bug affecting all regular PRs addressed #1576

mrfh92:
@mtar now it workes?

mtar (approved):

mrfh92:
@mtar thanks for reviewing :+1:

Issue #1514, opened 2024-06-07
Support batch 1D convolution


ClaudiaComito:
**Related**
#1500 

**Feature functionality**
1-D convolution on the last dimension of an n-D batch of signals. 

**Additional context**
Radioastronomy support, RFI mitigation project

mrfh92:
#1516 could be of interest

Issue #1396, opened 2024-03-08
Support 2-D convolution on n-D arrays


ClaudiaComito:
**Feature functionality**
We want to be able to exploit torch's batch convolution functionality. The current implementation of `ht.signal.convolve2d` #1007 only supports 2D signal and filter, just like `scipy.signal.convolve2d`.

lolacaro:
Happy to take this over, and expand the issue to allow 2D convolutions in general.

Addendum: I found issue 920 which is convolve2D, but I am happy to include this issue there.

PR #1515, opened 2024-06-07
Support batch 1-d convolution in `ht.signal.convolve`

ClaudiaComito:
## Due Diligence

- General:
- Implementation:

## Description

This PR expands `ht.signal.convolve` to perform 1-D convolutions on a batch of 1-D arrays, i.e. 

- if `signal.shape = (..., n)` and `kernel.shape = (m,)`:  `signal`'s last dimension will be convolved with `kernel`. 

- if e.g. `signal.shape = (i, j,  n)` and `kernel.shape = (i, j,  m,)`: each element `signal[i, j]` of size `n`  will be convolved with the corresponding element `kernel[i,j]` of size `m`. 

- Signal and kernel can be any number of dimensions, as long as the batch dimensions match.

- Signal and kernel can be distributed along any of the batch dimensions.

Issue/s resolved: #1514 
Related: #1396 

## Changes proposed:

- enable batch 1-D convolution
- expand documentation, examples
- expand tests
-

## Type of change

yes, instead of returning `ValueError` when `signal` is n-D, `ht.convolve` now raises a ValueError only if batch convolution isn't possible (shapes mismatch, unsuitable distribution)

### DIFF
--- a/heat/core/signal.py
+++ b/heat/core/signal.py
@@ -174,7 +174,7 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
         # all operations are local torch operations, only the last dimension is convolved
         local_a = a.larray
         local_v = v.larray
-        # flip filter for convolution as Pytorch conv1d computes correlations
+        # flip filter for convolution, as Pytorch conv1d computes correlations
         local_v = torch.flip(local_v, [-1])
         local_batch_dims = tuple(local_a.shape[:-1])
 
--- a/heat/core/signal.py
+++ b/heat/core/signal.py
@@ -196,7 +196,7 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
 
         # cast to single-precision float if on GPU
         if local_a.is_cuda:
-            float_type = promote_types(local_a.dtype, torch.float32)
+            float_type = torch.promote_types(local_a.dtype, torch.float32)
             local_a = local_a.to(float_type)
             local_v = local_v.to(float_type)
 
--- a/heat/core/signal.py
+++ b/heat/core/signal.py
@@ -123,6 +123,7 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
     if v.shape[-1] > a.shape[-1]:
         a, v = v, a
 
+    # assess whether to perform batch processing, default is False (no batch processing)
     batch_processing = False
     if a.ndim > 1:
         # batch processing requires 1D filter OR matching batch dimensions for signal and filter
@@ -145,6 +146,7 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
                 else:
                     v.resplit_(axis=a.split)
         batch_processing = True
+
     if not batch_processing and v.ndim > 1:
         raise ValueError(
             f"1-D convolution only supported for 1-dimensional signal and kernel. Signal: {a.shape}, Filter: {v.shape}"
--- a/heat/core/signal.py
+++ b/heat/core/signal.py
@@ -146,6 +146,7 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
         local_batch_dims = tuple(local_a.shape[:-1])
 
         # reshape signal and filter to 3D for Pytorch conv1d function
+        # see https://pytorch.org/docs/stable/generated/torch.nn.functional.conv1d.html
         local_a = local_a.reshape(
             torch.prod(torch.tensor(local_batch_dims, device=local_a.device), dim=0).item(),
             local_a.shape[-1],
--- a/heat/core/signal.py
+++ b/heat/core/signal.py
@@ -88,8 +88,22 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
     a = a.astype(promoted_type)
     v = v.astype(promoted_type)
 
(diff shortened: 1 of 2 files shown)


heat/core/signal.py
```diff
@@ -88,34 +119,104 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
...
+        # verify that the filter shape is consistent with the signal
+        if v.ndim > 1:
+            if v.shape[:-1] != batch_dims:
+                raise ValueError(
+                    f"Batch dimensions of signal and filter must match. Signal: {a.shape}, Filter: {v.shape}"
+                )
+        if a.is_distributed():
+            if a.split == a.ndim - 1:
+                raise ValueError(
+                    "Please distribute the signal along the batch dimension, not the signal dimension. For in-place redistribution use `a.resplit_(axis=0)`"
```
JuanPedroGHM, line 139: Signal and filter could have been swapped at this point, leading to the user to resplit the wrong thing.

heat/core/tests/test_signal.py
```diff
@@ -119,3 +120,29 @@ def test_convolve(self):
...
         conv = ht.convolve(1, 5)
         self.assertTrue(ht.equal(ht.array([5]), conv))
+
+        # test batched convolutions, distributed along the first axis
+        signal = ht.random.randn(1000, dtype=ht.float64)
+        batch_signal = ht.empty((10, 1000), dtype=ht.float64, split=0)
+        batch_signal.larray[:] = signal.larray
+        kernel = ht.random.randn(19, dtype=ht.float64)
+        batch_convolved = ht.convolve(batch_signal, kernel, mode="same")
+        self.assertTrue(ht.equal(ht.convolve(signal, kernel, mode="same"), batch_convolved[0]))
```
JuanPedroGHM, line 130: Just to be sure, I would test a couple more of the batches against the original signal, at least one per rank.

heat/core/tests/test_signal.py
```diff
@@ -119,3 +120,29 @@ def test_convolve(self):
...
+
+        # distributed kernel
+        dis_kernel = ht.array(kernel, split=0)
+        batch_convolved = ht.convolve(batch_signal, dis_kernel)
+        self.assertTrue(ht.equal(ht.convolve(signal, kernel), batch_convolved[0]))
+        batch_kernel = ht.empty((10, 19), dtype=ht.float64, split=1)
+        batch_kernel.larray[:] = dis_kernel.larray
+        batch_convolved = ht.convolve(batch_signal, batch_kernel, mode="full")
+        self.assertTrue(ht.equal(ht.convolve(signal, kernel, mode="full"), batch_convolved[0]))
+
```
JuanPedroGHM, line 140: An extra test where the signal or kernel has more than two dimensions is missing.

JuanPedroGHM (dismissed):

heat/core/tests/test_signal.py
ClaudiaComito: Added

heat/core/tests/test_signal.py
ClaudiaComito: added a check on the last rank. The signal is repeated across the -1 dimension, so the result really should be the same everywhere.

heat/core/signal.py
ClaudiaComito: Fixed error message.

heat/core/tests/test_signal.py
mrfh92: I would like to suggest another test with a.ndim > 3 in order to test the workaround that reduces to ndim = 3 as required by the underlying pytorch function

mrfh92: (Im surprised why these lines are covered so far although no arrays with ndim > 3 are in the tests (?) )

mrfh92 (changes requested):
except for the missing test of ndim > 3 (see comment) looks very fine :)
Thx :+1:

heat/core/signal.py
```diff
@@ -88,34 +119,108 @@ def convolve(a: DNDarray, v: DNDarray, mode: str = "full") -> DNDarray:
...
+                raise ValueError(
+                    "Please distribute the signal along the batch dimension, not the signal dimension. For in-place redistribution use the `DNDarray.resplit_()` method with `axis=0`"
+                )
+            if v.is_distributed():
+                if v.ndim == 1:
+                    # gather filter to all ranks
+                    v.resplit_(axis=None)
+                else:
+                    v.resplit_(axis=a.split)
+        batch_processing = True
```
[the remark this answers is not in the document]
ClaudiaComito: > Why is batch_processing = True hard coded?

Hi @krajsek , do you mean, it would be better to let the user set it as a keyword argument? If so, I agree.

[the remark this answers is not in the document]
mrfh92: As far as I understand things, it is set `batch_processing=True` and, depending on some conditions, it is set `batch_processing=False` lateron; in my opinion this is fine and I would not suggest to introduce a kwarg since actually the correct value of `batch_processing` is already uniquely determined by the other inputs.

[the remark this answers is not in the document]
krajsek: o.k.,  I agree, I also do not advocate for kwarg. Setting variables to some values in code always irritates me, but after reading the code again it makes sense. Maybe we spend a comment why this is hard coded here.

ClaudiaComito:
@krajsek @mrfh92 thanks for reviewing, I think I addressed all points, let me know if something's missing.

mrfh92 (approved):
All change requests have been addressed. 
Thanks for the work! 👍

PR #1529, opened 2024-06-16
Make `dataset.ishuffle` optional 

krajsek:
## Due Diligence

- General:
    - [x]  title of the PR is suitable to appear in the release notes
- Implementation:
    - [x] unit tests: no unit test added, so far datatools.py has no unit tests
    - [x] unit tests: no, same as above

## Description

So far, the Heat dataset requires the implementation of an ishuffle method that defines the shuffling of data between epochs. This is not compatible with the PyTorch dataset class. To enable the use of PyTorch dataset classes, ishuffle is now optional.

## Type of change
- New feature

 no


ClaudiaComito (approved):
Great, thanks @krajsek !

ClaudiaComito (approved):

PR #1616, opened 2024-08-12
downgrade sphinx version

mtar:
## Description

This fix downgrades the sphinx version back because sphinx_rtd_theme is not compatible with sphinx >= 8

## Changes proposed:

- set sphinx back to 7.4.7

## Type of change

Documentation / Fix


mrfh92 (approved):
approve to let CI run

Issue #1598, opened 2024-08-05
Support mpi4py 4.x


mrfh92:
**Related**
Is this feature request related to a problem or bug? Please describe.

**Feature functionality**
A clear and concise description of what the feature should do exactly and any alternative solutions.

**Additional context**

PR #1618, opened 2024-08-15
Support mpi4py 4.x.x

JuanPedroGHM:
## Due Diligence

- General:
- Implementation:

## Description

Not sure why this just appeared now, but MPI was complaining that our count and displacement tuples had some floats, when it should have been only int32. I don't think the changelog makes a specific point about this, maybe they were doing the checks before, but now it is up to us. 

This caused a lot of our tests to fail, as a lot of the functions using `Allgatherv` did not properly cast the count and displacements tuples.

Issue/s resolved: #1598 

## Changes proposed:

- On the method `as_buffer` in **comunications.py**, added extra lines that explicitly cast the elements of tuples count and displs to ints to avoid any unexpected casting errors further down the call stack. 
- Removed the limit to allow the usage of mpi4py==4.0.0 on the dependencies list.
- Corrected some of the function signatures to properly signal that some types are optional. 
- Properly skip tests if multiple nodes are required but only one is available ( looking at you, DASO )
- Tested it using OpenMPI 4.1, python 3.11, on 1 and 2 nodes in horeka, both with and without GPU. 

## Type of change
- Bug fix (could have some other breaking changes down the line)

yes

All the communication functions that make use of `as_buffer` to send/recv objects through MPI.


JuanPedroGHM:
### Benchmarks results - Sponsored by [perun](https://github.com/Helmholtz-AI-Energy/perun)

| function                                |   mpi_ranks | device   | metric   |      value |   ref_value |         std |   % change | type            | alert   |   lower_quantile |   upper_quantile |
|-----------------------------------------|-------------|----------|----------|------------|-------------|-------------|------------|-----------------|---------|------------------|------------------|
| matmul_split_1                          |           4 | GPU      | RUNTIME  | 0.0327519  |  0.0200216  | 0.00791738  | 63.5827    | jump-detection  | True    |     nan          |     nan          |
| qr_split_1                              |           4 | GPU      | RUNTIME  | 0.068254   |  0.0617905  | 0.00525142  | 10.4603    | jump-detection  | True    |     nan          |     nan          |
| heat_benchmarks                         |           4 | GPU      | RUNTIME  | 9.88148    |  9.88539    | 0.0185374   | -0.0395443 | trend-deviation | True    |       9.88446    |       9.88632    |
| matmul_split_0                          |           4 | GPU      | RUNTIME  | 0.0310716  |  0.0295967  | 0.00815078  |  4.98342   | trend-deviation | True    |       0.0284257  |       0.0307677  |
| matmul_split_1                          |           4 | GPU      | RUNTIME  | 0.0327519  |  0.0203692  | 0.00791738  | 60.7911    | trend-deviation | True    |       0.0200564  |       0.0206821  |
| qr_split_1                              |           4 | GPU      | RUNTIME  | 0.068254   |  0.0646602  | 0.00525142  |  5.55794   | trend-deviation | True    |       0.0620775  |       0.0672429  |
| lanczos                                 |           4 | GPU      | RUNTIME  | 0.581009   |  0.581877   | 0.00268783  | -0.149069  | trend-deviation | True    |       0.581694   |       0.582059   |
| hierachical_svd_rank                    |           4 | GPU      | RUNTIME  | 0.0938962  |  0.0946442  | 7.32051e-05 | -0.79029   | trend-deviation | True    |       0.0939748  |       0.0953135  |
| hierachical_svd_tol                     |           4 | GPU      | RUNTIME  | 0.117647   |  0.11929    | 0.0026648   | -1.37669   | trend-deviation | True    |       0.118977   |       0.119603   |
| kmeans                                  |           4 | GPU      | RUNTIME  | 0.630026   |  0.625066   | 0.00431094  |  0.79344   | trend-deviation | True    |       0.623466   |       0.626666   |
| kmedians                                |           4 | GPU      | RUNTIME  | 1.01198    |  1.00229    | 0.00508973  |  0.966815  | trend-deviation | True    |       0.999379   |       1.00519    |
| kmedoids                                |           4 | GPU      | RUNTIME  | 1.13906    |  1.13015    | 0.00587684  |  0.788974  | trend-deviation | True    |       1.1241     |       1.13619    |
| reshape                                 |           4 | GPU      | RUNTIME  | 0.250914   |  0.244356   | 0.0164063   |  2.68387   | trend-deviation | True    |       0.241201   |       0.24751    |
| apply_inplace_robust_scaler_and_inverse |           4 | GPU      | RUNTIME  | 5.50584    |  5.4586     | 0.0125942   |  0.865433  | trend-deviation | True    |       5.43485    |       5.48235    |
| apply_inplace_normalizer                |           4 | GPU      | RUNTIME  | 0.00166509 |  0.00168571 | 8.57187e-05 | -1.22341   | trend-deviation | True    |       0.00166924 |       0.00170219 |

[Grafana Dashboard](http://57bc8d92-72f2-4869-accd-435ec06365cb.ka.bw-cloud-instance.org:3000/d/adjpqduq9r7k0a/heat-cb?orgId=1)
Last updated: 2024-08-19T07:23:50Z

heat/core/tests/test_io.py
```diff
@@ -147,6 +148,10 @@ def test_load_csv(self):
         with self.assertRaises(TypeError):
             ht.load_csv(self.CSV_PATH, header_lines="3", sep=";", split=0)
 
+    @unittest.skipIf(
+        len(TestCase.get_hostnames()) > 1,
+        "Test only works on single node, file creation is not synchronized across nodes",
```
mtar, line 153:
```suggestion
        len(TestCase.get_hostnames()) > 1 or not os.environment.get('TMPDIR'),
        "Test only works on single node, file creation is not synchronized across nodes. Set the environment variable 'TMPDIR' to a path globally accessible by all nodes otherwise",
```

setup.py
```diff
@@ -33,7 +33,7 @@
         "Topic :: Scientific/Engineering",
     ],
     install_requires=[
-        "mpi4py>=3.0.0, <4.0.0",
+        "mpi4py>=3.0.0",
```
mtar, line 36: Did you run it on 3.x.x?

heat/core/tests/test_io.py
JuanPedroGHM: Could this cause problems when TMPDIR is set only on some of the nodes? Can that happen? I'm also not sure if this guarantees that TMPDIR is globally accessible by all the nodes.

setup.py
JuanPedroGHM: Not sure what you mean by that. I installed mpi4py 4.0.0 manually.

mtar: Have you checked wether the changes are backwards compatible with the older versions? If not we have to raise the minimum dependency or add some more lines reflecting that.

JuanPedroGHM: You are right, I did not check for backwards compatibility. I think the only problem with that one might be the rename of buffer/memory, but I will run some more tests.

heat/core/tests/test_io.py
JuanPedroGHM: I guess if it is meant for us, it should be good enough. I don't think there are a lot of people out there running the tests.

mtar (approved):
Looks good :+1:

PR #1677, opened 2024-10-15
Refactor `test_random` to minimize collective calls 

ClaudiaComito:
## Due Diligence

- General:
- Implementation:
    - [x] benchmarks: created for new functionality

## Description

`test_random` has been giving us problems in connection to `.numpy()` calls (aka Allgather/Allgatherv and copying to CPU) before. 

As far as I can tell, it isn't any particular instance of "allgathering" that doesn't work. On the AMD runner (2-process GPU tests), since this Monday, `test_random` has been failing consistently around the 10th numpy() call in the module. 

I have refactored `test_random` to gather and copy only when absolutely necessary. It now gathers/copies to CPU only 8 times, as opposed to 47 in the legacy implementation.

Issue/s resolved: #1682 

## Changes proposed:

- remove unnecessary numpy() calls
-
-
-

## Type of change

## Memory requirements

NA
## Performance

NA
no

### DIFF
--- a/heat/core/statistics.py
+++ b/heat/core/statistics.py
@@ -982,11 +982,18 @@ def mean(x: DNDarray, axis: Optional[Union[int, Tuple[int, ...]]] = None) -> DND
         return mu_tot[0][0] if mu_tot[0].size == 1 else mu_tot[0]
 
     # ----------------------------------------------------------------------------------------------
+    # sanitize dtype
+    if types.heat_type_is_exact(x.dtype):
+        if x.dtype is types.int64:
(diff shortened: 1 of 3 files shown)


mtar (commented):
Thank you. You skipped the median tests. Is this intentional?

ClaudiaComito:
> Thank you. You skipped the median tests. Is this intentional?

Yes, I skipped the `ht.median` tests because they are very communication-intensive. The np.median tests are all still in.

mtar (approved):

PR #1681, opened 2024-10-16
Release drafter update and autolabeler

JuanPedroGHM:
## Due Diligence

- General:
- Implementation:
    - [x] benchmarks: created for new functionality

## Description

Update to the configuration. This will not work properly on the next release, but as long as we stay on the same release branch, this should be fine. We could stick with our current branching strategy, but every time we create a new release branch, it will cause some trouble. 

This also includes some auto-labeling functionality. 

Issue/s resolved: #

## Changes proposed:

- Workflow changes to releases drafter

no


.github/rd-release-config.yml
```diff
@@ -0,0 +1,167 @@
+filter-by-commitish: False
+include-pre-releases: false
+categories:
+  - title: '🔥 Features'
+    labels:
+      - 'features'
+  - title: '🚀 Enhancemets'
+    labels:
+      - 'performance'
+      - 'enhancement'
```
ClaudiaComito, line 10: So far we've used the label `enhancement` for new features

.github/rd-release-config.yml
```diff
@@ -0,0 +1,167 @@
...
+    title:
+    - '/feat/i'
+    - '/feature/i'
+    branch:
+      - '/feature/i'
+      - '/feat/i'
+  - label: 'enhancement'
+    title:
+      - '/Refactor.+/'
+      - '/Expand.+/'
```
ClaudiaComito, line 84: I understand the mix-up now. Indeed so far we always labeled new features as `enhancement`. Branches where new features are developed start with `features/`

PRs containing new features typically start with "Implement ....", "Add...  ", or "features/" if the branch name is being used as a title.

Maybe we *should* always keep the branch name as a PR title initially to facilitate autolabeling.

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,85 @@
...
+        env:
+          PR_BRANCH: pre-release/${{ inputs.release_version }}
+          VERSION: ${{ inputs.release_version }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Create branch
+          git checkout -b pre-release/${VERSION}
+
+          # Split version number into major, minor, and micro
+          MAYOR=$(echo $VERSION | cut -d. -f1)
```
ClaudiaComito, line 42:
```suggestion
          MAJOR=$(echo $VERSION | cut -d. -f1)
```

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,85 @@
...
+          MINOR=$(echo $VERSION | cut -d. -f2)
+          MICRO=$(echo $VERSION | cut -d. -f3)
+
+          # Write on to the version.py file
+          sed -i "s/major: int = \([0-9]\+\)/major: int = $MAYOR/g" heat/core/version.py
```
ClaudiaComito, line 47:
```suggestion
          sed -i "s/major: int = \([0-9]\+\)/major: int = $MAJOR/g" heat/core/version.py
```

.github/rd-release-config.yml
```diff
@@ -0,0 +1,167 @@
...
+    files:
+      - 'docker/**/*'
+  - label: 'bug'
+    title:
+      - '/bug/i'
+      - '/resolve/i'
+      - '/fix/i'
+    branch:
+      - '/bug/i'
+      - '/fix/i'
```
ClaudiaComito, line 102: bug branches start with `bugs/` .

For bugs it would be great to add the `backport` label by default.

.github/rd-release-config.yml
```diff
@@ -0,0 +1,167 @@
...
+    label: 'spatial'
+  - title: 'Utils'
+    label: 'utils'
+exclude-labels:
+  - 'github-actions'
+  - 'dependencies'
+  - 'chore'
+  - 'workflows'
+
+autolabeler:
```
ClaudiaComito, line 65: When is this action triggered? Ideally PRs should be labeled when they are first opened.

ClaudiaComito (changes requested):
Thanks a lot @JuanPedroGHM , I have a few comments / questions

.github/rd-release-config.yml
JuanPedroGHM: True, but that makes it difficult to distinguish between just general improvements to older stuff from and actual new features. We can maybe drop the enhancements tag/category.

.github/rd-release-config.yml
```diff
@@ -0,0 +1,167 @@
+filter-by-commitish: False
+include-pre-releases: false
+categories:
+  - title: '🔥 Features'
+    labels:
+      - 'features'
+  - title: '🚀 Enhancemets'
```
mtar, line 7:
```suggestion
  - title: '🚀 Enhancements'
```

.github/workflows/release-prep.yml
```diff
@@ -0,0 +1,85 @@
...
+
+          [1]: https://github.com/peter-evans/create-pull-request" \
+          --label invalid
+
+          # Create PR for main
+          gh pr create --base main \
+          --head ${{ env.PR_BRANCH }} \
+          --title "Heat ${{ env.VERSION }} - Main" \
+          --body "Copy of latest pre-release PR targeting release." \
+          --label invalid
```
ClaudiaComito, line 85: Should this PR be created before release or within the post-release workflow?

JuanPedroGHM: I thought it could be useful to have it from the start, as it might highlight merge conflicts and CI errors at the same time in both target branches at the same time, so we would find the earlier. 

We could ignore until after the release PR is merged, or move this to a post-release workflow

ClaudiaComito: Perfect, let's try it out

ClaudiaComito (approved):
Thanks for all the work @JuanPedroGHM !

Issue #1627, opened 2024-08-19
[Bug]: `test_vmap` fails on multi-node runs on hardware accelerators


JuanPedroGHM:
### What happened?

When running on more than one node and using GPUs at the same time, `test_vmap` fails. Needs further investigation. 

### Error message or erroneous outcome

```shell
The result of the test does not match the expected outcome.
```

### Version

main (development branch)

### Python version

3.11

### PyTorch version

2.2

### MPI version

```shell
OpenMPI 4.1, 5.0
```

JuanPedroGHM:
No, this is with OpenMPI 4.1 and mpi4py 3.1.6. I updated the description of the issue with the specific dependencies and configuration where it fails.

PR #1738, opened 2024-12-04
Decrease accuracy on `ht.vmap` tests on multi-node GPU runs 

mrfh92:
## Due Diligence

- General:
- Implementation:
    - [ ] ~~benchmarks: created for new functionality~~
    - [ ] ~~benchmarks: performance improved or maintained~~
    - [ ] ~~documentation updated where needed~~

## Description
Issue/s resolved: #1627 

## Changes proposed:
default accuracy in allclose is to high for one test on GPUs and some CPUs 

## Type of change
decrease accuracy (to tolerance 1e-4, which is still small enough to exclude passing of the test by lucky chance)


mrfh92:
seems to be even more strange: the GitHub runner also only achieves tol=1e-4 on CPU for the respective test

JuanPedroGHM (changes requested):
Did some tests, tried with extra configurations on CPU and GPU. I could not find why the error is so large when running on GPU.  Test pass with lower tolerance when using float64. 

Here is how I tested the multiple configurations, I would recommend rewriting the test like this. 

```python
def test_vmap_with_chunks(self):
            x1_splits = [None, 1]
            chunk_sizes = list(range(1,5))
            dtypes = [ht.float32, ht.float64]
            for x1_split in x1_splits:
                for cs in chunk_sizes:
                    for dtype in dtypes: 
                        with self.subTest(x1_split=x1_split, chunk_size=cs, dtype=dtype):
                            # same as before but now with prescribed chunk sizes for the vmap
                            x0 = ht.random.randn(5 * ht.MPI_WORLD.size, 10, 10, split=0, dtype=dtype)
                            x1 = ht.random.randn(10, 5 * ht.MPI_WORLD.size, split=x1_split, dtype=dtype)
                            out_dims = (0, 0)

                            def func(x0, x1, k=2, scale=1e-2):
                                return torch.topk(torch.linalg.svdvals(x0), k)[0] ** 2, scale * x0 @ x1

                            vfunc = ht.vmap(func, out_dims, chunk_size=cs)
                            y0, y1 = vfunc(x0, x1, k=2, scale=-2.2)

                            # compare with torch
                            x0_torch = x0.resplit(None).larray
                            x1_torch = x1.resplit(None).larray
                            vfunc_torch = torch.vmap(func, (0, x1_split), out_dims)
                            y0_torch, y1_torch = vfunc_torch(x0_torch, x1_torch, k=2, scale=-2.2)

                            self.assertTrue(torch.allclose(y0.resplit(None).larray, y0_torch))
                            tol = 1e-12 if dtype == ht.float64 else 1e-4
                            self.assertTrue(torch.allclose(y1.resplit(None).larray, y1_torch, atol=tol))
```

ClaudiaComito:
After the PR meeting on March 24th, I think we concluded that  we adopt the test implementation as per @JuanPedroGHM 's comment above and otherwise merge, is that correct @mrfh92 ?

mrfh92:
@ClaudiaComito yes this is correct. I have added the suggested code.

ClaudiaComito (approved):

PR #1753, opened 2024-12-16
Load slices of hdf5 dataset

JuanPedroGHM:
Some users have expressed interest in only loading partial slices from an HDF5 dataset. This PR adds the ```slices``` argument to ```load_hdf5```, to allow dataset slicing like in the with the ```h5py``` API.

## Due Diligence

- General:
- Implementation:

## Description

Issue/s resolved: #

## Changes proposed:

- Slice argument for ```load_hdf5```
- Auxiliary function to calculate the new size and offset of a sliced axis
- Hypothesis and pytest as a dependency for tests
- Hypothesis tests for the new functionality. 

yes / no

### DIFF
--- a/heat/core/io.py
+++ b/heat/core/io.py
@@ -38,6 +38,31 @@ __all__ = [
     "load_npy_from_path",
 ]
 
+
+def size_from_slice(size: int, s: slice) -> Tuple[int, int]:
+    """
+    Determines the size of a slice object.
+
+    Parameters
+    ----------
+    size: int
+        The size of the array the slice object is applied to.
+    s : slice
+        The slice object to determine the size of.
+
+    Returns
+    -------
+    int
+        The size of the sliced object.
+    int
+        The start index of the slice object.
+    """
+    from hypothesis import note
+
+    new_range = range(size)[s]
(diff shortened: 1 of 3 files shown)


Berkant03 (approved):

heat/core/io.py
```diff
@@ -505,10 +528,8 @@ def load_hdf5(
             Name of the dataset to be read.
         dtype : datatype, optional
             Data type of the resulting array.
-        load_fraction : float between 0. (excluded) and 1. (included), default is 1.
-            if 1. (default), the whole dataset is loaded from the file specified in path
-            else, the dataset is loaded partially, with the fraction of the dataset (along the split axis) specified by load_fraction
-            If split is None, load_fraction is automatically set to 1., i.e. the whole dataset is loaded.
+        slices : tuple of slice objects, optional
+            Load only the specified slices of the dataset.
```
ClaudiaComito, line 532: Shall we add an example below? Otherwise I think this is good to go, thanks a lot!

ClaudiaComito (approved):
:+1: thanks @JuanPedroGHM !

Issue #1674, opened 2024-10-14
Optimization of k-means initialization


Hakdag97:
- Implementation of the "k-means ||" initialization.
- Further imprvement: additionally implement for k-medians

mrfh92:
Completed in #1754

PR #1754, opened 2024-12-17
Parallel initialization for k-means

Hakdag97:
## Description
The bottleneck of k-means clustering (concerning runtime) is the initialization of centroids, which was previously built on a cost intensive serial algorithm. The aim of this pull request is to replace this algorithm by the more sophisticated k-means || initialization of centroids. 

Issue/s resolved: 
* #1674 

## Changes proposed:
* Complete new implementation of the initialization of centroids used for k-means, k-medians, and k-medoids
* Adjustment of classes (like *KMeans*) to match with the new implementation

## Type of change
* Bug fix
* New feature

## Performance
* Reducing the runtime of initialization of clustering algorithm in distributed and non-distributed mode with `split=None` and `split not None` by (at least) an order of magnitude (depending on the setting concerning, e.g., size of data and chosen parameters) 

* yes: the classes *KMeans*, *KMedoids*, *KMedians* and the function *where* are affected


heat/cluster/batchparallelclustering.py
```diff
@@ -169,7 +174,7 @@ def functional_value_(self) -> float:
         """
         return self._functional_value
 
-    def fit(self, x: DNDarray):
+    def fit(self, x: DNDarray, weights: torch.tensor = 1):
```
mrfh92, line 177: This is the fit of batchparallel clustering. If we allow for specification of weights here, it must be a DNDarray, and below the corresponding local arrays would have to be used for the local clusterings.

heat/cluster/batchparallelclustering.py
```diff
@@ -233,6 +241,7 @@ def fit(self, x: DNDarray):
                         self.max_iter,
                         self.tol,
                         local_random_state,
+                        weights,
```
mrfh92, line 244: see above

heat/cluster/kmeans.py
```diff
@@ -102,7 +102,7 @@ def _update_centroids(self, x: DNDarray, matching_centroids: DNDarray):
 
         return new_cluster_centers
 
-    def fit(self, x: DNDarray) -> self:
+    def fit(self, x: DNDarray, oversampling: float = 100, iter_multiplier: float = 20) -> self:
```
mrfh92, line 105: Are there references for these default values? I had a look into dask and they use oversampling=2 as default.

heat/cluster/_kcluster.py
```diff
@@ -123,53 +133,106 @@ def _initialize_cluster_centers(self, x: DNDarray):
...
+                # output format: tensor
+                init_min_distance = init_distance.min(axis=1)
+                # --> Pick the minimal distance of the data points to each centroid
+                # output format: vector
+                init_cost = init_min_distance.sum()
+                # --> Now calculate the cost
+                # output format: scalar
+                #
+                # Iteratively fill the tensor storing the centroids
+                for _ in ht.arange(0, iter_multiplier * ht.log(init_cost)):
```
mrfh92, line 162: for loop counters, the standard python range should be more effective

heat/cluster/_kcluster.py
```diff
@@ -123,53 +133,106 @@ def _initialize_cluster_centers(self, x: DNDarray):
...
+                    # Calculate the distance between data points and the current set of centroids
+                    distance = ht.spatial.distance.cdist(x, centroids, quadratic_expansion=True)
+                    min_distance = distance.min(axis=1)
+                    # Sample each point in the data to a new set of centroids
+                    prob = oversampling * min_distance / min_distance.sum()
+                    # -->   probability distribution with oversampling factor
+                    #       output format: vector
+                    idx = ht.where(sample <= prob)
```
mrfh92, line 170: one may think about moving the creation of sample here to have some new sampling every step of the loop

heat/cluster/_kcluster.py
```diff
@@ -123,53 +133,106 @@ def _initialize_cluster_centers(self, x: DNDarray):
...
+                    # -->   pick the data points that are identified as possible centroids and make sure
+                    #       that data points and centroids are split in the same way
+                    #       output format: vector
+                    centroids = ht.row_stack((centroids, local_data))
+                    # -->   stack the data points with these indices to the DNDarray of centroids
+                    #       output format: tensor
+                # Evaluate distance between final centroids and data points
+                if centroids.shape[0] <= self.n_clusters:
+                    raise ValueError(
+                        "The oversampling factor and/or the number of iterations are chosen"
```
mrfh92, line 183: why two strings? 
It may also be helpful to use sth like f"The oversampling factor (={oversampling}) ... " to give the user the values in the error message.

heat/cluster/_kcluster.py
```diff
@@ -123,53 +133,106 @@ def _initialize_cluster_centers(self, x: DNDarray):
...
+                    #     Note: as we only recluster the centroids for initialization with standard k-means ++,
+                    #     this list of centroids can also be used to initialize k-medians and k-medoids
+                    reclustered_centroids = batch_kmeans[0]
+                    # --> access the reclustered centroids
+                else:
+                    # ensure that all processes have the same data
+                    reclustered_centroids = torch.zeros(
+                        (self.n_clusters, centroids.shape[1]),
+                        dtype=x.dtype.torch_type(),
+                        device=centroids.device,
```
mrfh92, line 223: centroids.device.torch_device()?

heat/cluster/_kcluster.py
```diff
@@ -123,53 +133,106 @@ def _initialize_cluster_centers(self, x: DNDarray):
...
+                    )
+                    # -->  tensor with zeros that has the same size as reclustered centroids, in order to to
+                    #      allocate memory with the correct type in all processes(necessary for broadcast)
+                ht.MPI_WORLD.Bcast(
+                    reclustered_centroids, root=0
+                )  # by default it is broadcasted from process 0
+                reclustered_centroids = ht.array(reclustered_centroids, split=x.split)
```
mrfh92, line 230: I dont know whether we want to split the centroids as they are probably only a small number. So this might produce overhead compared to having split=None here.

heat/cluster/_kcluster.py
Hakdag97: For the sake of readability, I would like to avoid single lines of code that are too long and thus decided to split the string in two lines. However, in Python you need a separate string for each line of code. The values of the parameters have been added tp the error message.

heat/cluster/_kcluster.py
Hakdag97: Done

heat/cluster/_kcluster.py
Hakdag97: Could you please specify your comment?

heat/cluster/batchparallelclustering.py
Hakdag97: Yes, we do not need to use weights here at all. All weights in this fit method are removed.

heat/cluster/kmeans.py
Hakdag97: Larger values are more comfortable for the user, because the initialization might fail for too small values. A corresponding note (also regarding speed) is added to the docstring.

heat/cluster/_kcluster.py
Hakdag97: Adjusted the device accordingly.

mrfh92 (approved):
approve to let the CI-matrix run (because it fails somewhere else and I want to check whether this is due to changes there or a general problem)

mrfh92:
CI-matrix failed as expected; but this has nothing to do with this particular PR :)

mrfh92:
The continuous benchmarks seem to fail. You can reproduce this by going to the `benchmarks/cb` folder an running
```
HEAT_DEVICE=cpu mpirun -n 4 perun monitor main.py
```
For debugging I recommend to use 
```
HEAT_DEVICE=cpu mpirun -n 4 python cluster.py
```
as perun tends to hide the actual error messages. 

I get (at least sometimes) the following error that seems to be stochastic: 
```
  File "/home/hopp_fa/heat/benchmarks/cb/main.py", line 20, in <module>
    run_cluster_benchmarks()
  File "/home/hopp_fa/heat/benchmarks/cb/cluster.py", line 30, in run_cluster_benchmarks
    kmeans(data)
  File "/home/hopp_fa/.virtualenvs/heatenv_cuda12_torch251/lib/python3.10/site-packages/perun/api/decorator.py", line 30, in func_wrapper
    func_result = func(*args, **kwargs)
  File "/home/hopp_fa/heat/benchmarks/cb/cluster.py", line 8, in kmeans
    kmeans.fit(data)
  File "/home/hopp_fa/heat/heat/cluster/kmeans.py", line 127, in fit
    self._initialize_cluster_centers(x, oversampling, iter_multiplier)
  File "/home/hopp_fa/heat/heat/cluster/_kcluster.py", line 218, in _initialize_cluster_centers
    batch_kmeans = _kmex(
  File "/home/hopp_fa/heat/heat/cluster/batchparallelclustering.py", line 63, in _kmex
    centers = _initialize_plus_plus(X, n_clusters, p, random_state, weights)
  File "/home/hopp_fa/heat/heat/cluster/batchparallelclustering.py", line 45, in _initialize_plus_plus
    idxs[i] = torch.multinomial(weights * dist, 1)
RuntimeError: probability tensor contains either `inf`, `nan` or element < 0
```

Hakdag97:
> The continuous benchmarks seem to fail. You can reproduce this by going to the `benchmarks/cb` folder an running
> 
> ```
> HEAT_DEVICE=cpu mpirun -n 4 perun monitor main.py
> ```
> 
> For debugging I recommend to use
> 
> ```
> HEAT_DEVICE=cpu mpirun -n 4 python cluster.py
> ```
> 
> as perun tends to hide the actual error messages.
> 
> I get (at least sometimes) the following error that seems to be stochastic:

Interesting, I never came across that error. I will have a look into it

Hakdag97:
> > The continuous benchmarks seem to fail. You can reproduce this by going to the `benchmarks/cb` folder an running
> > ```
> > HEAT_DEVICE=cpu mpirun -n 4 perun monitor main.py
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For debugging I recommend to use
> > ```
> > HEAT_DEVICE=cpu mpirun -n 4 python cluster.py
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > as perun tends to hide the actual error messages.
> > I get (at least sometimes) the following error that seems to be stochastic:
> 
> Interesting, I never came across that error. I will have a look into it

Problem resolved.

mrfh92:
@Hakdag97 can you reproduce the error in the benchmarks locally? If not this might be a tricky one and we might debug tomorrow together

mrfh92:
```
Rank 0: probability tensor contains either `inf`, `nan` or element < 0
```
Could also be sth seed-dependent, so it could help to fix a seed.

mrfh92:
@Hakdag97 also changing the runners (from Nvidia A100 to AMD MI250x) did not remove the problem, so it looks like that there is really sth wrong in the implementation that causes this error in the benchmarks

JuanPedroGHM:
# Benchmarks results - Sponsored by [perun](https://github.com/Helmholtz-AI-Energy/perun)

## FTPx86-AMD-Milan-Mi250 - CPU - N4 - RUNTIME

| Function                             |   Value [s] |   Change % |   30 day avg. [s] |   Last entrie [s] |
|--------------------------------------|-------------|------------|-------------------|-------------------|
| kmeans                               |   0.0258541 |   -92.2669 |          0.334331 |          0.329861 |
| kmedians                             |   0.167711  |   -59.3692 |          0.412768 |          0.409243 |
| kmedoids                             |   0.169523  |   -76.9673 |          0.736011 |          0.72836  |
| convolution_array_distributed_stride |   0.622897  |   -10.742  |          0.697861 |          0.686329 |
---

## FTPx86-AMD-Milan-Mi250 - GPU - N4 - RUNTIME

| Function                                  |   Value [s] |   Change % |   30 day avg. [s] |   Last entrie [s] |
|-------------------------------------------|-------------|------------|-------------------|-------------------|
| heat_benchmarks                           | 52.1203     |   -10.6804 |       58.3526     |       58.2536     |
| kmeans                                    |  0.326896   |   -85.2736 |        2.21979    |        2.22169    |
| kmedians                                  |  1.48841    |   -58.6932 |        3.6033     |        3.60163    |
| kmedoids                                  |  1.554      |   -60.6734 |        3.95151    |        3.96252    |
| concatenate                               |  0.0943668  |   -11.0318 |        0.106068   |        0.0931919  |
| apply_inplace_standard_scaler_and_inverse |  0.047194   |   -10.7713 |        0.0528911  |        0.0672661  |
| apply_inplace_normalizer                  |  0.0130085  |   -22.3963 |        0.0167627  |        0.0136837  |
| convolution_array_distributed_stride      |  0.038895   |    19.9555 |        0.0324245  |        0.0314236  |
| convolution_kernel_distributed            |  0.109099   |   -11.1089 |        0.122733   |        0.111113   |
| convolution_distributed                   |  0.086565   |   -11.1304 |        0.0974068  |        0.0878557  |
| convolution_distributed_stride            |  0.129167   |    14.1722 |        0.113134   |        0.113142   |
| convolution_batch_processing              |  0.00336819 |   -33.2678 |        0.00504732 |        0.00423431 |
---

[Grafana Dashboard](https://930000e0-e69a-4939-912e-89a92316b420.ka.bw-cloud-instance.org)
Last updated: 2025-10-06T12:51:14Z

mrfh92:
@Hakdag97 do you know whether the fix increased the runtime? The +5406.5% in the benchmarks could be a random fluctuation (we also have not so many reference values since we switched to MI250x recently) but the jump is quite high.

mrfh92:
Although there is no error shown, there seems to be a problem with the benchmark main workflow: 
```
0/4: Working on gpu:0
salloc: Job 5454 has exceeded its time limit and its allocation has been revoked.
srun: Job step aborted: Waiting up to 32 seconds for job step to finish.
slurmstepd: error: *** STEP 5454.1 ON ftp-x86n8 CANCELLED AT 2025-07-29T16:06:18 DUE TO TIME LIMIT ***
srun: error: ftp-x86n8: tasks 0-3: Terminated
  File "/hkfs/home/project/hk-project-test-aihero2/io3047/builds/t2_y6EuQK/0/helmholtz-analytics/cb/heat/.venv/bin/perun", line 10, in <module>
[2025-07-29 16:06:21,908][perun.io.text_report][textReport][ERROR] - R0/1:Non existent run id
    sys.exit(cli())
             ^^^^^
```
@JuanPedroGHM do you know whether thats an ongoing problem with the setup or could this rather be sth related to the PR?

JuanPedroGHM:
I think it's pretty clear from the benchmark report and from the error message that kmeans and similar functions are running incredibly slow. 5000% slower compared to previous runs. The error is because the job runtime limits are reached. I can increase the allocation time if necessary. Did you change the size of the benchmarks significantly?

mrfh92:
Sry, I was not sure whether the two weeks old report was still valid. But you are right, thats likely the cause.

JuanPedroGHM:
So, the last update broke the benchmarks again, so don't worry too much about them. Sorry about that 😅 I would still check why there is such a big slowdown on AMD, or if it is also on nvidia.

mrfh92:
@JuanPedroGHM I created an own issue for the benchmarks to keep track of there

Hakdag97:
Benchmarking issue solved: there were just two keyword arguments set to "too large" values for the comparably small benchmarks.

mrfh92:
@JuanPedroGHM Due to 

[Merge branch 'fix/cuda_aware_ompi_check' of github.com:helmholtz-analytics/heat into features/1674-Optimization_of_k-means_initialization](https://github.com/helmholtz-analytics/heat/pull/1754/commits/e0282fabd8fa5d929e288be18d9a03197df84d7f)

this PR got mixed up with your Sturdier MPI Check PR. Do you have any idea how to proceed?

JuanPedroGHM:
> @JuanPedroGHM Due to
> 
> [Merge branch 'fix/cuda_aware_ompi_check' of github.com:helmholtz-analytics/heat into features/1674-Optimization_of_k-means_initialization](https://github.com/helmholtz-analytics/heat/pull/1754/commits/e0282fabd8fa5d929e288be18d9a03197df84d7f)
> 
> this PR got mixed up with your Sturdier MPI Check PR. Do you have any idea how to proceed?

Well, now we need to merge the sturdier MPI one first; otherwise, it will be obsolete, and might cause some conflicts later. 

At least the benchmarks run now, and with not a single alert :+1:

mrfh92:
ok :+1:

Issue #1632, opened 2024-08-26
support I/O with zarr


mrfh92:
This Issue is to investigate options for I/O in Heat with zarr-files.

mrfh92:
Found sth that could serve as template how to combine zarr and mpi4py: 
https://github.com/rabernat/zarr_hdf_benchmarks/blob/master/parallel_read_write.py

ClaudiaComito:
Confirming we want to implement this after discussing this on Dec 2

ClaudiaComito:
Sample datasets:
 
- https://zarr.dev/datasets/  (>= 3 TB)
- AtmoRep ERA5 data (12 grib files of ~1.5 GBytes each, with script to convert them to zarr format): https://datapub.fz-juelich.de/atmorep/era5-data.html

PR #1766, opened 2025-01-24
Implement support for `zarr` data format

Berkant03:
## Due Diligence

- General:
- Implementation:
  

## Description

Added functions to load zarr into memory and save a dndarray into zarr file.

Issue/s resolved: #1632 

## Type of change
- New feature

## Memory requirements

Memory requirements are as high as the dataset to be loaded is big for loading the data.
Saving an array can up to double the required memory. 

## Performance

Save Performance depends on the chunk sizes of the zarr array. This depends on various factors like number of
processes and the shape of the array.

no


mrfh92:
It looks like that the zarr-specific tests have not been run on codebase. The tests are clearly there, but the lines are not covered according to codecov; so I guess, thats related to the fact that zarr is optional dependency and not installed for the tests on codebase, so the tests are all skipped.

Berkant03:
Yes someone with access to the runner needs to add zarr to the pip installation.

mrfh92:
I'm currently trying to do so in #1775

mrfh92:
@Berkant03 the CI now took zarr into account. there seems to be a small error somewhere (you can see the error message by clicking onto "Details" next to the ci/codebase entry above. Then you can access the full logs of the run by clicking onto test-amd; test-cuda has been cancelled as the error seems to introduce a deadlock.

heat/core/io.py
```diff
@@ -1283,3 +1285,169 @@ def load_csv_from_folder(
...
+
+        # Wait for the file creation to finish
+        MPI_WORLD.handle.Barrier()
+        zarr_array = zarr.open(store=path, mode="r+", **kwargs)
+
+        if dndarray.split is not None:
+            _, _, slices = MPI_WORLD.chunk(dndarray.gshape, dndarray.split)
+
+            zarr_array[slices] = (
+                dndarray.larray.numpy()  # Numpy array needed as zarr can only understand numpy dtypes and infers it.
```
mrfh92, line 1447: this seems to cause a problem in the tests if the larray is actually on a GPU. 
See https://pytorch.org/docs/stable/generated/torch.Tensor.numpy.html for the options.

heat/core/io.py
```diff
@@ -1283,3 +1285,169 @@ def load_csv_from_folder(
...
+            if fnmatch.fnmatch(path, f"*{extension}"):
+                break
+        else:
+            raise ValueError("path does not end on an Zarr extension.")
+
+        if MPI_WORLD.rank == 0:
+            if os.path.exists(path) and not overwrite:
+                raise RuntimeError("Given Path already exists.")
+
+            # Zarr functions by chunking the data, were a chunk is a file inside the store.
```
mrfh92, line 1390: It might be a good idea to add this as 
```
Notes 
---------- 
...
```
to the docstring of the function.

heat/core/io.py
```diff
@@ -1283,3 +1285,171 @@ def load_csv_from_folder(
...
+            if dndarray.split is None or MPI_WORLD.size == 1:
+                chunks = None
+            else:
+                chunks = np.array(dndarray.gshape)
+                axis = dndarray.split
+
+                if chunks[axis] % MPI_WORLD.size != 0:
+                    chunks[axis] = 1
+                else:
+                    chunks[dndarray.split] //= MPI_WORLD.size
```
mrfh92, line 1429: Do chunk sizes in zarr have to be the same for each chunk? (If not, wouldn't it be an option to choose chunk size in zarr identical to the size of the larray's?)

heat/core/io.py
```diff
@@ -1283,3 +1285,171 @@ def load_csv_from_folder(
...
+                chunks=chunks,
+                **kwargs,
+            )
+
+        # Wait for the file creation to finish
+        MPI_WORLD.Barrier()
+        zarr_array = zarr.open(store=path, mode="r+", **kwargs)
+
+        if dndarray.split is not None:
+            _, _, slices = MPI_WORLD.chunk(dndarray.gshape, dndarray.split)
```
mrfh92, line 1446: As far as I understand this, this works best if the chunking of the DNDarray into larray's coincides with the chunking of the zarr file. Is this true?

mrfh92: My concern is the following: The chunk sizes for zarr are prescribed by shape[split] // MPI_WORLD.size, but the chunks in Heat (size of larray's) is shape[split] // MPI_WORLD.size + 1 for the first shape[split] % MPI_WORLD.size processes and only shape[split] // MPI_WORLD.size for the remaining processes.

heat/core/io.py
mrfh92: ok, I see that it looks like they have to be uniform unlike in Heat

mrfh92 (approved):
I have a minor comment on possible performance problems for writing, if chunk sizes in Heat intereact with chunk sizes in zarr, but at the moment I dont see a way out of this problem. 
Therefore, I recommend merging. If performance problems should be observed in edge cases, we can address this later. 

Thanks for the work :+1:

mrfh92:
@Berkant03 It looks like that the tests for Python 3.9 and 3.10 are hanging at `test_io` independent of the PyTorch version used 
In https://pypi.org/project/zarr/ they say you need Python>=3.11 for zarr.

heat/core/io.py
Berkant03: I would not necessarily say that it works the best if the chunks coincide, as that means that the array is in theory loaded two times, as a chunk is always loaded for fast access by Zarr. Which does not matter for a small amount of data. Its more of a trade of for easy distribution and possible faster writes.

Regarding the second point it should  be no problem as the +1 is only added in the case that the array cannot be distributed evenly. In which case it goes back to chunk size 1 along the split axis.

Berkant03:
If zarr can only be used with python 3.11 then those who want to use must use 3.11. I quickly looked at the wiki for the zarr2 versions but these were probably deleted or something else is happening as it says these sites do not exist. At least its an optional dependency for.

mrfh92:
During experimenting with the scalability I got the following error (while writing zarr): 
```
Codec does not support buffers of > 2147483647 bytes...
```
This seems to be a problem when the chunks for writing zarr get too large. Can we somehow catch this, either by formulating our own error message, or by choosing a different chunk size?

mrfh92:
Code zum Benchmarken: 
python```
import heat as ht 
from perun import monitor 
import time 
import os
import shutil

@monitor()
def save_as_zarr(array, path):
    ht.save_zarr(path,array)

@monitor()
def load_from_zarr(path,split):
    array = ht.load_zarr(path,split=split,device="cpu")
    return array

if __name__ == "__main__":
    PATH = f"your_path/test_{ht.MPI_WORLD.size}.zarr"

    array = ht.random.randn(1000, 1000, split=0, dtype=ht.float32, device="cpu") 

    if ht.MPI_WORLD.rank == 0: 
        shutil.rmtree(PATH) if os.path.exists(PATH) else None
    ht.MPI_WORLD.Barrier()
    save_as_zarr(array, PATH)
    ht.MPI_WORLD.Barrier() 
    del array 

    ht.MPI_WORLD.Barrier()
    array = load_from_zarr(PATH,0)
    ht.MPI_WORLD.Barrier()
    del array

    if ht.MPI_WORLD.rank == 0: 
        shutil.rmtree(PATH) if os.path.exists(PATH) else None
```

File: heat/core/io.py (hunk 3 above)
Comment (mrfh92) [inline heat/core/io.py:1429]: I guess this can cause a problem as the chunks can become quite large here. So it might be reasonable to restrict the chunk size to the allowed number of bytes in zarr.

**Comment by mrfh92** [comment]
memory
time

Memory and runtime for the 2000000 x 12500-array (~100GB) with the above script.

**Comment by mrfh92** [comment]
@mtar can you give us a rough description of the zarr-files you are working with and the corresponding setup? (like dimensions of the array, split axis, chunk sizes, number of processes etc.) 
Then we could benchmark the targeted use case.

File: heat/core/io.py (hunk 3 above)
Comment (Berkant03) [reply heat/core/io.py:1429]: The easiest would be to just restrict the Split Axis to one element (But this may even slow things down because of frequent I/O). Another could be to use the Synchronized Array which @mtar has mentioned to me once after finishing the feature. Will look into it.

**Comment by Berkant03** [comment]
@mrfh92 I added a reduction of the chunk size if the limit is exceeded. Can you check if it now works with 64 processes? I also tried an approach where i use locks to allow possibly better writing performance but synchronization was thrown out the window with zarr v2. I tried to write my own locking mechanism but it turned out to have too many issues with locking regions if i try to be generic. I will look into this approach another time but for the time being this should probably help with atleast writing any data, albeit slow.

**Comment by mrfh92** [comment]
@Berkant03 Im at a conference this week, so I can try to check next week

@Berkant03 @ClaudiaComito do you think it makes sense (and is possible) to add loading of slices (as for hdf5) as well?

**Review by ClaudiaComito — CHANGES_REQUESTED** [review]
@Berkant03 @JuanPedroGHM would it make sense to mention the argument `slices` in the `load()` docstring?

File: heat/core/io.py [hunk 5]
```diff
@@ -770,7 +774,7 @@ def load(
     """
     Attempts to load data from a file stored on disk. Attempts to auto-detect the file format by determining the
     extension. Supports at least CSV files, HDF5 and netCDF4 are additionally possible if the corresponding libraries
-    are installed.
+    are installed.is
```
ClaudiaComito, line 777:
```suggestion
    are installed.
```

heat/core/io.py
```diff
@@ -794,6 +798,16 @@ def load(
...
     DNDarray([ 1.0000,  2.7183,  7.3891, 20.0855, 54.5981], dtype=ht.float32, device=cpu:0, split=None)
+
+    See Also
+    --------
+    :func:`load_csv` : Loads data from a CSV file.
+    :func:`load_netcdf` : Loads data from a NetCDF4 file.
+    :func:`load_hdf5` : Loads data from an HDF5 file.
+    :func:`load_zarr` : Loads zarr-Format into DNDarray which will be returned.
+    :func:`load_npy_from_path` : Loads multiple .npy files into one DNDarray which will be returned.
+    :func:`load_csv_from_folder` : Loads multiple .csv files into one DNDarray which will be returned.
```
ClaudiaComito, line 809:
```suggestion
    :func:`load_csv` : Loads data from a CSV file.
    :func:`load_csv_from_folder` : Loads multiple .csv files into one DNDarray which will be returned.
    :func:`load_hdf5` : Loads data from an HDF5 file.
    :func:`load_netcdf` : Loads data from a NetCDF4 file.
    :func:`load_npy_from_path` : Loads multiple .npy files into one DNDarray which will be returned.
    :func:`load_zarr` : Loads zarr-Format into DNDarray which will be returned.
```

heat/core/io.py
```diff
@@ -1221,6 +1241,11 @@ def save(
             raise RuntimeError(f"netcdf is required for file extension {extension}")
     elif extension in __CSV_EXTENSION:
         save_csv(data, path, *args, **kwargs)
+    elif extension in __ZARR_EXTENSIONS:
+        if supports_zarr():
+            return save_zarr(data, path, *args, **kwargs)
+        else:
+            raise RuntimeError(f"zarr is required for file extension {extension}")
```
ClaudiaComito, line 1248:
```suggestion
            raise RuntimeError(f"Package zarr is required for file extension {extension}")
```

heat/core/io.py
```diff
@@ -811,6 +825,12 @@ def load(
             return load_netcdf(path, *args, **kwargs)
         else:
             raise RuntimeError(f"netcdf is required for file extension {extension}")
+    elif extension in __ZARR_EXTENSIONS:
+        if supports_zarr():
+            return load_zarr(path, *args, **kwargs)
+        else:
+            raise RuntimeError(f"zarr is required for file extension {extension}")
```
ClaudiaComito, line 832:
```suggestion
            raise RuntimeError(f"Package zarr is required for file extension {extension}")
```

ClaudiaComito (changes requested):
Thanks a lot @Berkant03 !

@mrfh92 I only have a few edits, if you agree I'll commit and we can merge this

mrfh92:
@ClaudiaComito @Berkant03 
I agree that the PR is fine now. Thanks for the work @Berkant03! :) 

The only thing I came across are the few lines missing coverage so far, but they are either for the case that zarr support is missing (which we can't test unfortunately if we also want to test zarr support) and those lines covering the case `byte_size > CODEC_LIMIT_BYTES`; here, I guess, it does not make sense to test because that refers to cases with large data that would likely (as the large count MPI tests) result in instable behaviour of the runners.

ClaudiaComito (approved):

Issue #1564, opened 2024-07-11
Add DMDc


mrfh92:
DMDc (Dynamic Mode Decomposition with control) shall be added to the decomposition module

PR #1794, opened 2025-02-14
Add Dynamic Mode Decomposition with Control (DMDc)

mrfh92:
## Due Diligence

- General:
- Implementation:
    - [ ] ~~benchmarks: created for new functionality~~ (not a new basic function, just applied linalg)
    - [ ] ~~benchmarks: performance improved or maintained~~ (does not apply because no changes in basic functions)

## Description

Issue/s resolved: #1564 

## Type of change
new feature 

no


heat/decomposition/tests/test_dmd.py
JuanPedroGHM, line 1: Consider expanding the test (splits, dtypes, shapes) and reducing code duplication using subtests and loops.

https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

heat/core/dndarray.py
```diff
@@ -368,6 +368,13 @@ def array_with_halos(self) -> torch.Tensor:
         """
         return self.__cat_halo()
 
+    @property
+    def info(self) -> str:
```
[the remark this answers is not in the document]
mrfh92: done :+1:

heat/decomposition/tests/test_dmd.py
mrfh92: I guess the problem is that its not that easy to parametrize these tests as you cant choose the admissible parameter combinations from a cartesian grid of choices

mrfh92: Moreover, due to the overal number of options I guess this would get quite expensive in terms of runtime

mrfh92: @JuanPedroGHM nevertheless, I have restructured the test and put them into smaller functions to ensure traceability of errors in the future; moreover, I have added a test for DMDc with split=1

heat/core/printing.py
```diff
@@ -205,6 +205,14 @@ def __str__(dndarray) -> str:
     )
 
 
+def __repr__(dndarray) -> str:
+    """
+    Returns a printable representation of the passed DNDarray.
+    Unlike the __str__ method, which prints a representation targeted at users, this method targets developers by showing key internal parameters of the DNDarray.
+    """
+    return f"MPI-rank: {dndarray.comm.rank}, Shape: {dndarray.shape}, Split: {dndarray.split}, Local Shape: {dndarray.lshape}, Device: {dndarray.device}, Dtype: {dndarray.dtype.__name__}"
```
JuanPedroGHM, line 213: I think it would be better if we surround everything inside ```f"DNDArray(...)"```. It will be easier to recognize what we are printing on long debugging outputs, and will be more consistent with the python practices.

heat/decomposition/dmd.py
```diff
@@ -5,6 +5,7 @@
 import heat as ht
 from typing import Optional, Tuple, Union, List
 import torch
+import numpy as np
```
JuanPedroGHM, line 8: Do we need numpy?

JuanPedroGHM (changes requested):

heat/core/printing.py
mrfh92: done. Changed to `<DNDarray(...)>`

heat/decomposition/dmd.py
mrfh92: removed

JuanPedroGHM (approved):

Hakdag97 (approved):

Issue #1789, opened 2025-02-11
Implementation of a Heat Aware DistributedSampler for Interoperability with Pytorch Ecosystem


Berkant03:
**Related**
This is related to the issue #1772 , where it was discussed to add a torch.distributed backend but was deemed not feasible. 

**Feature functionality**
It was proposed and agreed on to implement our own Heat aware `DistributedSampler` to be used by the pytorch functions for easier interoperability.

PR #1807, opened 2025-02-24
Add a Heat aware DistributedSampler for torch usage.

Berkant03:
… on a process local basis.

## Due Diligence

- General:
- Implementation:

## Description

Issue/s resolved: #1789 

## Changes proposed:

Add a Heat aware DistributedSampler for usage for PyTorch use cases  

## Type of change

no

### DIFF
--- a/heat/utils/data/datatools.py
+++ b/heat/utils/data/datatools.py
@@ -2,9 +2,13 @@
 Function and classes useful for loading data into neural networks
 """
 
+import random
 import torch
+import torch.distributed
 from torch.utils import data as torch_data
 from typing import Callable, List, Iterator, Union, Optional, Sized
+
+import torch.utils
 
 from ...core.dndarray import DNDarray
 from ...core.communication import MPI_WORLD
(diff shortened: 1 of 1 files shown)


Berkant03:
When using the normal comm.Bcast the Bcast only works the first time and the seconds time not anymore.
```shell
❯ mpirun -np 2 python test.py
0 tensor([2, 4, 3, 0, 1], dtype=torch.int32)
1 tensor([2, 4, 3, 0, 1], dtype=torch.int32)
DNDarray([[10, 11, 12, 13, 14],
          [20, 21, 22, 23, 24],
          [15, 16, 17, 18, 19],
          [ 0,  1,  2,  3,  4],
          [ 5,  6,  7,  8,  9]], dtype=ht.int32, device=cpu:0, split=0)

1 tensor([455,   0,   0,   0,  32], dtype=torch.int32)
0 tensor([1, 4, 3, 2, 0], dtype=torch.int32)
```

Berkant03:
The issue of the batches test is because of the way the iterator of the dataloader is created. Pytorch uses a wrapper Iterator for the Sampler and sets a random seed before the iterator of the sampler is invoked, therefore not using the provided seed. Because of this the samples do not match the shuffled reference array. Therefore i would just say to skip it.

JuanPedroGHM:
**TODO**
- ```__repr__``` might have been a mistake, should return to previous behaviour :cry: 
- Investigate difference between existing ```ht.random.permute``` and ```_alltoall_shuffle```
- Move ```SeedEnvironemnt``` to base test class. (Creation and teardown methods, possibly)

JuanPedroGHM (approved):

heat/core/tests/test_suites/basic_test.py
```diff
@@ -63,6 +63,7 @@ def setUpClass(cls) -> None:
             )
 
         cls.device, cls.other_device, cls.envar, cls.is_mps = ht_device, other_device, envar, is_mps
+        seed(42)
```
mrfh92: In my understanding, this line can be removed as with #1947 a seed has already been set in lines 21/22.

mrfh92 (dismissed):
Looks fine to me. I only have the one comment below.

Issue #1552, opened 2024-07-03
Make unit tests compatible with NumPy 2.x 


mrfh92:
Many of our unit tests use NumPy for computation of a reference solution; some of them fail (in their purely NumPy-part) due to small changes introduced in NumPy 2.x 

at the end: do not forget to remove restriction to numpy<2 from the depenencies

PR #1826, opened 2025-03-14
Make unit tests compatible with NumPy 2.x

Marc-Jindra:
## Due Diligence

- General:
- Implementation:

## Description

Fixed DeprecationWarnings from numpy 2.x
- [x] row_stack to vstack
- [x] passing axes to np.fft.fftn when s is not None
- [ ] `__array__` implementation doesn't accept a copy keyword (probably related to `Tensor.__array__`) --> torch issue
- [ ] `__array_wrap__` must accept context and return_scalar arguments --> torch issue
- [x] Arrays of 2-dimensional vectors are deprecated (`np.cross()`)

Issue/s resolved: #1552 

## Changes proposed:

- unittests use vstack instead of row_stack if the numpy version is >= 2.0.0
- unittests now pass axes to np.fft.fftn when s is not None
- changed setup.py to allow numpy versions >= 2.0.0
- changed requirements (python 3.10, numpy 1.23.5, scipy 1.14.0)
- changed workflows to use python 3.10

yes / no

### DIFF
--- a/heat/core/linalg/tests/test_basics.py
+++ b/heat/core/linalg/tests/test_basics.py
@@ -62,14 +62,20 @@ class TestLinalgBasics(TestCase):
         b_2d = ht.array(np_b[:-1, :, :], split=1)
         cross_3d_2d = ht.cross(a, b_2d, axisa=1, axisb=0)
         np_cross_3d_2d = np.cross(
-            np_a, np.concatenate([np_b[:-1, :, :], np.zeros((1, 40, 50))], axis=0), axisa=1, axisb=0
+            np_a,
+            np.concatenate([np_b[:-1, :, :], np.zeros((1, 40, 50))], axis=0, dtype=np.float32),
+            axisa=1,
+            axisb=0,
         )
         self.assert_array_equal(cross_3d_2d, np_cross_3d_2d)
 
         a_2d = ht.array(np_a[:, :-1, :], split=0)
         cross_2d_3d = ht.cross(a_2d, b, axisa=1, axisb=0)
         np_cross_2d_3d = np.cross(
-            np.concatenate([np_a[:, :-1, :], np.zeros((40, 1, 50))], axis=1), np_b, axisa=1, axisb=0
+            np.concatenate([np_a[:, :-1, :], np.zeros((40, 1, 50))], axis=1, dtype=np.float32),
+            np_b,
+            axisa=1,
+            axisb=0,
         )
         self.assert_array_equal(cross_2d_3d, np_cross_2d_3d)
 
--- a/heat/core/linalg/tests/test_basics.py
+++ b/heat/core/linalg/tests/test_basics.py
@@ -61,20 +61,23 @@ class TestLinalgBasics(TestCase):
         self.assert_array_equal(cross_axisc, np_cross_axisc)
 
         # test vector axes with 2 elements
-        if np.lib.NumpyVersion(np.__version__) < "2.0.0b1":
-            b_2d = ht.array(np_b[:-1, :, :], split=1)
-            cross_3d_2d = ht.cross(a, b_2d, axisa=1, axisb=0)
-            np_cross_3d_2d = np.cross(np_a, np_b[:-1, :, :], axisa=1, axisb=0)
-            self.assert_array_equal(cross_3d_2d, np_cross_3d_2d)
+        b_2d = ht.array(np_b[:-1, :, :], split=1)
+        cross_3d_2d = ht.cross(a, b_2d, axisa=1, axisb=0)
+        np_cross_3d_2d = np.cross(
+            np_a, np.concatenate([np_b[:-1, :, :], np.zeros((1, 40, 50))], axis=0), axisa=1, axisb=0
+        )
+        self.assert_array_equal(cross_3d_2d, np_cross_3d_2d)
 
-            a_2d = ht.array(np_a[:, :-1, :], split=0)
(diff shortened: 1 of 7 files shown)


Marc-Jindra:
# Numpy 2 unittest Warnings

## `row_stack` (solved)

```
DeprecationWarning: `row_stack` alias is deprecated. Use `np.vstack` directly.
```

self explanatory.

## `__array__` (I think it’s related to torch)

```
DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments. To learn more, see the migration guide https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword
```
[Here is the link](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword)

Triggered in: `test_save (heat.core.tests.test_io.TestIO.test_save)`<br>
Or: `/heat/heat/core/io.py:441/439` by the following code: `var[merged_slices] = data.larray.cpu()`

Apparently, it's related to PyTorch, as the `Tensor.__array__` method does not accept a copy keyword: `def __array__(self, dtype=None)`.

## `__array_wrap__`

```
DeprecationWarning: __array_wrap__ must accept context and return_scalar arguments (positionally) in the future. (Deprecated in NumPy 2.0)
```
Triggered in:
- test_fit_iris_unsplit (heat.cluster.tests.test_kmedians.TestKMedians.test_fit_iris_unsplit)
- test_exp2 (heat.core.tests.test_exponential.TestExponential.test_exp2)
- test_sort (heat.core.tests.test_manipulations.TestManipulations.test_sort)
- test_randn (heat.core.tests.test_random.TestRandom_Batchparallel.test_randn)
- test_RobustScaler (heat.preprocessing.tests.test_preprocessing.TestRobustScaler.test_RobustScaler)

Or:
- numpy/_core/fromnumeric.py:48
- heat/heat/core/tests/test_exponential.py:99

test_exponential Line 99: `tmp = np.exp2(torch.arange(elements, dtype=torch.float64))`

I couldn't find anything in the NumPy documentation, but there should be a `__array_wrap__` method in an ndarray subclass to which `context=None` and `return_scalar=None` must be added. I couldn’t find such a method in heat.

## `np.cross()`

```
DeprecationWarning: Arrays of 2-dimensional vectors are deprecated. Use arrays of 3-dimensional vectors instead. (deprecated in NumPy 2.0)
```

Calling `np.cross` with two-dimensional arrays is no longer supported. As a solution, [NumPy](https://numpy.org/doc/stable/reference/generated/numpy.cross.html#) suggests writing your own function:
```
def cross2d(x, y):
    return x[..., 0] * y[..., 1] - x[..., 1] * y[..., 0]
```
However, this is not the desired solution.

## `axes` (solved)

```
DeprecationWarning: `axes` should not be `None` if `s` is not `None` (Deprecated in NumPy 2.0). In a future version of NumPy, this will raise an error and `s[i]` will correspond to the size along the transformed axis specified by `axes[i]`. To retain current behaviour, pass a sequence [0, ..., k-1] to `axes` for an array of dimension k.
```

Now, when `s` is passed to `np.fft.fftn`, `axes` must also be provided. To maintain the previous default behavior, `axes` must be passed for `k` elements in `s`, which correspond to the **last** k dimensions of the array (starting from 0).

Example for a 3-dimensional array: `np.fft.fftn(x.numpy(), s=(6, 6), axes=(1, 2))`

Marc-Jindra:
ci/codebase fails because it tries to build with python 3.9 instead of 3.10

mtar:
The runner uses Python 3.12.3. Why do you want a higher version?

mrfh92:
I refered to the comment above that the tests fail because of Python 3.9, but actually it looks more like thats the usual problem in the large count MPI test on the AMD runnner that I am already working on.

Marc-Jindra:
No it is about `ERROR: No matching distribution found for scipy>=1.14.0` in test_mps. I had that same Error before and it got fixed when i changed the python version from 3.9 to 3.10.

mtar:
The wrong python version has been fixed yesterday. However, as we don't support apple mps yet, it will still fail regardless.

heat/core/linalg/tests/test_basics.py
```diff
@@ -61,19 +61,20 @@ def test_cross(self):
...
+            self.assert_array_equal(cross_3d_2d, np_cross_3d_2d)
+
+            a_2d = ht.array(np_a[:, :-1, :], split=0)
+            cross_2d_3d = ht.cross(a_2d, b, axisa=1, axisb=0)
+            np_cross_2d_3d = np.cross(np_a[:, :-1, :], np_b, axisa=1, axisb=0)
+            self.assert_array_equal(cross_2d_3d, np_cross_2d_3d)
+
+            cross_z_comp = ht.cross(a_2d, b_2d, axisa=1, axisb=0)
+            np_cross_z_comp = np.cross(np_a[:, :-1, :], np_b[:-1, :, :], axisa=1, axisb=0)
+            self.assert_array_equal(cross_z_comp, np_cross_z_comp)
```
ClaudiaComito, line 77: @Marc-Jindra During the PR meeting this morning we decided to keep the "cross_2d" functionality in Heat for now and compare it to the pytorch result in the tests when numpy==2. I.e. compare the distributed `ht.cross` (on split arrays) with the non-distributed pytorch output.

ClaudiaComito (approved):

PR #1834, opened 2025-03-20
Skip large-count communication tests on AMD runner

mrfh92:
So far, the only test of large count MPI actually run on the AMD runner was the test for Isend-Recv on CPU on 4 processes. For whatever reason this often failed. I have run the tests for the communications module (incl. the large scale tests) both on CPU and GPU (2x Nvidia) on our workstation for 2 to 8 processes. No error occurred; thus it is very likely a problem related to the AMD-runner which motivated skipping these tests on the AMD runners.

New version is to run large count Isend-Recv and Allreduce only on the CUDA-runner but on 2 and 3 (both GPU) processes.

### DIFF
--- a/heat/core/tests/test_communication.py
+++ b/heat/core/tests/test_communication.py
@@ -2522,8 +2522,8 @@ class TestCommunication(TestCase):
 
     # the following test is only for up to three processes to save memory
     @unittest.skipIf(
-        ht.MPI_WORLD.size > 3 or "rocm" in torch.__version__,
-        "Only for up to three processes and not on the AMD runner",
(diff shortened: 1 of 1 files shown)


mrfh92:
I have run the tests for the communications module (incl. the large scale tests) both on CPU and GPU (2x Nvidia) on our workstation for 2 to 8 processes. No error occurred; thus it is very likely a problem related to the AMD-runner which motivated skipping these tests on the AMD runners.

JuanPedroGHM (approved):

ClaudiaComito (approved):

Issue #1820, opened 2025-03-10
Support device awareness of `result_type`, `promote_types`


ClaudiaComito:
**Related**
#1129 #1053 

**Feature functionality**
Apple MPS don't support double precision floats or complex numbers, so we have had to introduce many dtype checks in the codebase (see #1129, example https://github.com/helmholtz-analytics/heat/pull/1129#discussion_r1888789987 ).  Most of these checks could be bypassed if `type.result_type()` and `type.promote_types()` were aware of what device they are setting the dtype for (for example with a kwarg `device=None` with options `cpu`, `gpu`, `mps`, potentially `xpu` later, see #1757 ) 
 

**Additional context**
PR meeting March 10 2025

PR #1836, opened 2025-03-21
Support device awareness of result_type and promote_types

Marc-Jindra:
## Due Diligence

- General:
- Implementation:

## Description

Apple MPS don't support double precision floats or complex numbers, so this PR introduces some checks for that in result_type and promote_types. 

Issue/s resolved: #1820 

## Changes proposed:

- result_type now checks whether an arrays device is mps and changes the dtype accordingly (float64 -> float32, complex128 -> complex64)
- promote_types now takes an optional device argument and checks if that device is mps
-
-

## Type of change

no

### DIFF
--- a/heat/core/types.py
+++ b/heat/core/types.py
@@ -855,7 +855,9 @@ def issubdtype(
 
 
 def promote_types(
-    type1: Union[str, Type[datatype], Any], type2: Union[str, Type[datatype], Any]
+    type1: Union[str, Type[datatype], Any],
+    type2: Union[str, Type[datatype], Any],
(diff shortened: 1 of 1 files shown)


JuanPedroGHM (changes requested):
There are still a couple of instances where references to MPS need to be removed from the codebase. Some quick examples that I found:
- ```_operations.py: _binary_op and _cum_op```
- ```arithmetics.py: hypot and hypot_```

There might be more, but those are some I could find quickly that already use ```promote_type``` or ```result_type```. The references to MPS should be removable. 

Otherwise, great code. Thanks!

PR #1859, opened 2025-04-14
Bugs/1858  bug load zarr tests fail on main

mrfh92:
* subTest'ified the zarr tests 
* at two cases I needed to exclude a "0" from the range of a loop to avoid tests failing 
* depending on the zarr version, one has to use either `create_array` (new) or `create` (old). It looks like that the new API requires zarr-Versions compatible only with Python 3.11 which I do not have available on my workstation for development.


mrfh92:
However, it might be reasonable to discuss this in the PR whether it makes sense to support the old API further

Berkant03:
Thanks for subtestifying the tests and pointing out this issue. The testcases with the removal of the 0 make sense, as 0 dimensional array is not really a thing, though it is interesting that the tests on my machine did not throw an error. 

Regarding the issue of `create` and `create_array`.  An upper bound for the version of Zarr might be a good idea, to stay on top of changes.

Berkant03 (approved):

Issue #1879, opened 2025-06-02
[Bug]: 0_setup-conda.sh may install into base environment


thawn:
### What happened?

if the line `conda activate heat-env` does not work, the script installs packages into the base environment.

it would be great if this was prevented

### Code snippet triggering the error

```shell

./0_setup_conda.sh

```

### Error message or erroneous outcome

```shell
conda: execute conda init first
```

### Version

main (development branch)

### Python version

None

### PyTorch version

None

### MPI version

irrelevant

PR #1880, opened 2025-06-02
Exit installation if conda environment cannot be activated

thawn:
otherwise the script might install into the base environment

## Due Diligence

- General:
- Implementation:

## Description

Issue/s resolved: #1879 

## Changes proposed:

- exit install script if environment activation fails
-
-
-

## Type of change

- Bug fix

## Memory requirements

none
no


ClaudiaComito (approved):
Thanks a lot @thawn !

mrfh92:
@JuanPedroGHM @ClaudiaComito just went through our PRs and found that only the benchmarks were failing; I guess that was due some experiment while migrating the runners? 

If yes, I guess we can merge this now?

PR #1917, opened 2025-07-22
Loading of folder containing hdf5 files as DNDarray

Berkant03:
- created tests for the function

## Due Diligence

- General:
- Implementation:

## Description

Implementation of a Function that allows loading of all hdf5 files inside the given folder. 
Created as pendant to the DistributedSampler and Imagenet1k Dataset loading.

Issue/s resolved: #1916 

## Type of change

no

### DIFF
--- a/heat/core/io.py
+++ b/heat/core/io.py
@@ -6,12 +6,13 @@ from functools import reduce
 import operator
 import os.path
 from math import log10
+from pathlib import Path
 import numpy as np
 import torch
 import warnings
 import fnmatch
 
-from typing import Dict, Iterable, List, Optional, Tuple, Union
+from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 from . import devices
 from . import factories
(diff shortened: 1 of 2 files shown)


mrfh92:
@Berkant03 if you are ready, feel free to request review from me because I already supervised the implementation of a similar function for npy and csv

mrfh92:
I have fixed a minor bug in the tests: 

* the direct combination of communication and assert likely caused some problem; moreover, I replaced assert with "our" standard construction 

Differences compared to existing loading functionality for multiple files (npy and csv): 

* in this PR we have proper load balancing: thus, good performance is expected even in the case the files to load are of very different size 
* in this PR we assume the overall array being split along axis 0 (w.r.t. the files and also w.r.t. the final DNDarray) which is different from the other implementations where we allowed for all splits

mrfh92:
Note to myself: 
* extend tests 
* consider the case of different splits

mrfh92:
Decision in PR-talk: merge with split=0 option only, create issue on remaining split cases

PR #1948, opened 2025-08-18
Set random seed as part of setup of basic test case class

mrfh92:
see title

### DIFF
--- a/heat/core/tests/test_suites/test_basic_test.py
+++ b/heat/core/tests/test_suites/test_basic_test.py
@@ -6,6 +6,9 @@ from .basic_test import TestCase
 
 
 class TestBasicTest(TestCase):
+    def test_random_seed(self):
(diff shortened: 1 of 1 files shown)


mrfh92:
Discuss: 
* set seed for every test (instead only for initializing the class)?
* set deterministic seed or include randomization by varying seeds?

JuanPedroGHM (changes requested):
I would move setting the seed to the  ```setup()``` method of the class, to guarantee that each test starts from with that seed, and not only the file or module.

See: https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp

mrfh92:
> I would move setting the seed to the `setup()` method of the class, to guarantee that each test starts from with that seed, and not only the file or module.
> 
> See: https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp

taken into account

JuanPedroGHM (approved):

PR #1954, opened 2025-08-24
feat: add ptp function to statistics module (#1897)

### LINKED ISSUE #1897: implement ptp()
**Feature functionality**
The [ptp](https://numpy.org/doc/2.2/reference/generated/numpy.ptp.html) function from numpy returns the range of values (maximum - minimum) along an axis.

ivansherbakov9:
## Due Diligence
- General:
    - [x] **title** of the PR is suitable to appear in the [Release Notes](https://github.com/helmholtz-analytics/heat/releases/latest)
- Implementation:
    - [ ] unit tests: MPS tested (1 MPI process, 1 GPU)

## Description

Implemented `ht.ptp`, the peak-to-peak function, consistent with `numpy.ptp`.
Supports `axis`, `out`, and `keepdims` arguments.

Closes #1897

Issue/s resolved: #1897

## Changes proposed:
- Added `ptp` function to `heat/core/statistics.py`
- Added `DNDarray.ptp` method binding
- Added unit tests for `ptp` in `heat/core/tests/test_statistics.py`
- Updated docstrings and examples

## Type of change

## Memory requirements
No additional memory requirements compared to equivalent `ht.max` and `ht.min`.

## Performance
Performance is expected to be similar to running `ht.max` and `ht.min` separately.  
Future optimization may reduce collective communications to a single pass.

#### Does this change modify the behaviour of other functions? 
No.

### DIFF
--- a/heat/core/statistics.py
+++ b/heat/core/statistics.py
@@ -40,6 +40,7 @@ __all__ = [
     "min",
     "minimum",
     "percentile",
+    "ptp",
     "skew",
     "std",
     "var",
@@ -1737,6 +1738,58 @@ def percentile(
         del sorted_x
 
     return percentile
+
+
+def ptp(
+    x: DNDarray,
+    axis: Optional[Union[int, Tuple[int, ...]]] = None,
+    out: Optional[DNDarray] = None,
+    keepdims: bool = False,
+) -> DNDarray:
+    """
+    Range of values (maximum - minimum) along a given axis.
+
+    Parameters
+    ----------
+    x : ht.DNDarray
+        Input array.
+    axis : None or int or Tuple[int,...], optional
+        Axis or axes along which to operate. By default, flattened input is used.
+        If this is a tuple of ints, the ptp is selected over multiple axes,
+        instead of a single axis or all the axes as before.
+    out : DNDarray, optional
+        Output array to place the result.
+    keepdims : bool, optional
+        If this is set to ``True``, the axes which are reduced are left in the result as dimensions with size one.
+
+    Returns
+    -------
+    ptp : ht.DNDarray
+        An array with the same shape as `x`, with the specified axis removed. If `keepdims` is True, the reduced axes are left in the result as dimensions with size one.
+
+    Examples
+    --------
+    >>> a = ht.array([[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]], dtype=ht.float32)
+    >>> ht.ptp(a)
+    DNDarray([11.], dtype=ht.float32, device=cpu:0, split=None)
+    >>> ht.ptp(a, axis=0)
+    DNDarray([9., 9., 9.], dtype=ht.float32, device=cpu:0, split=None)
+    >>> ht.ptp(a, axis=1)
+    DNDarray([2., 2., 2., 2.], dtype=ht.float32, device=cpu:0, split=None)
+    """
+    xmax = max(x, axis=axis, keepdims=keepdims)
+    xmin = min(x, axis=axis, keepdims=keepdims)
+
+    if out is not None:
+        arithmetics.subtract(xmax, xmin, out=out)
+        return out
+    return arithmetics.subtract(xmax, xmin)
+
+
+DNDarray.ptp: Callable[[DNDarray, Union[int, Tuple[int, ...]], DNDarray, bool], DNDarray] = (
+    lambda x, axis=None, out=None, keepdims=False: ptp(x, axis, out, keepdims)
+)
+DNDarray.ptp.__doc__ = ptp.__doc__
 
 
 def skew(x: DNDarray, axis: int = None, unbiased: bool = True) -> DNDarray:
--- a/heat/core/tests/test_statistics.py
+++ b/heat/core/tests/test_statistics.py
@@ -1327,6 +1327,75 @@ class TestStatistics(TestCase):
         with self.assertRaises(ValueError):
             ht.percentile(X, q, axis=axis, sketched=True, sketch_size=10)
 
+    def test_ptp(self):
+        # argument errors
+        x = ht.zeros((2, 3, 4))
+        with self.assertRaises((ValueError, IndexError)):
+            ht.ptp(x, axis=10)
+        with self.assertRaises(TypeError):
+            ht.ptp(x, axis="01")
+        with self.assertRaises(TypeError):
+            ht.ptp(x, axis=(0, "10"))
+
+        # simple deterministic example
+        a = ht.array([[1, 2, 3],
+                      [4, 5, 6],
+                      [7, 8, 9],
+                      [10, 11, 12]])
+        self.assertEqual(ht.ptp(a), 11)
+        self.assertEqual(a.ptp(), 11)
+
+        # helper to check split of result
+        def __split_calc(ht_split, axis):
+            sp = ht_split if axis > ht_split else ht_split - 1
+            if axis == ht_split:
+                sp = None
+            return sp
+
+        # 1D: comparison with NumPy and invariance to split
+        v = ht.random.rand(50)
+        v_np = v.copy().numpy()
+        self.assertAlmostEqual(ht.ptp(v), np.ptp(v_np), places=5)
+
+        v = ht.resplit(v, 0)
+        self.assertAlmostEqual(ht.ptp(v), np.ptp(v_np), places=5)
+
+        # 2D float (take into account mps as in other tests)
+        dtype = ht.float32 if self.is_mps else ht.float64
+        X = ht.random.rand(50, 30, dtype=dtype)
+        X_np = X.copy().numpy()
+
+        # without axis
+        self.assertAlmostEqual(ht.ptp(X) - np.ptp(X_np), 0.0, places=5)
+
+        # along axes and for different splits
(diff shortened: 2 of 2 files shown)


ivansherbakov9:
@JuanPedroGHM @ClaudiaComito Thanks, I've applied the suggested fix 👍

ClaudiaComito:
> @JuanPedroGHM @ClaudiaComito Thanks, I've applied the suggested fix 👍

@ivansherbakov9 thanks a lot for your contribution! Do you want to try and write the custom reduction operation to minimize communication?

ivansherbakov9:
@ClaudiaComito Thanks for the review!  

I'd like to try implementing the custom MPI reduction for ptp, to minimize communication as you suggested.  
I'll experiment with it and ask here (or in Mattermost) if I get stuck.

ivansherbakov9:
@ClaudiaComito
It implements a custom MPI reduction (MPI_MINMAX) for the ptp() operation. Instead of performing separate global min and max reductions, both values are packed and reduced in a single communication step. This minimizes communication overhead while preserving correctness across all dtypes and shapes.

ClaudiaComito:
@ivansherbakov9 just a quick note - we've just started testing the copilot PR reviews so please feel free to disagree and not apply the suggested changes - like for any review.

ivansherbakov9:
@ClaudiaComito Thanks for the suggestions. I applied a subset of Copilot's proposals:
* Changed callback annotation to memoryview for clarity
* Narrowed exception handling to (ValueError, TypeError).
* Replaced the lambda binding with an explicit _ptp wrapper for readability

I intentionally kept the candidate dtype list explicit (extended to common types) rather than generating it dynamically because the dynamic approach was brittle and hard to read. If you prefer a different ordering of dtypes (e.g. int64 first) I can adjust.

heat/core/statistics.py
```diff
@@ -1460,6 +1461,86 @@ def mpi_argmin(a: str, b: str, _: Any) -> torch.Tensor:
...
+            lhs_np = lhs_data
+            rhs_np = rhs_data
+            break
+
+    if chosen is None:
+        raise RuntimeError(
+            "mpi_minmax: could not autodetect buffer dtype/size. "
+            f"bytes-left={len(a)}, bytes-right={len(b)}. "
+            "Ensure local_minmax returns same dtype/layout on all ranks."
+        )
```
JuanPedroGHM: Hi, thank you in advance for the PR again. I have one small suggestion: instead of trying to autodetect the data type and size with exceptions, I would recommend using closures to capture the buffer sizes and data types.

This should speed up the function and eliminate the need to capture exceptions. The operation will have to be freed after each use, but that is a common practice due to the limited number of self-defined operations available. 

You can check a similar example [in our implementation of _derived_op to handle custom MPI data types on reduction operations.](https://github.com/helmholtz-analytics/heat/blob/main/heat/core/communication.py#L805)

ivansherbakov9:
@ClaudiaComito @JuanPedroGHM 
This commit implements the reviewers' suggestions:

- Moved the min/max MPI-op factory into the communications module.
- Implemented a closure-based `_minmax_op(dtype, count, offset)` that creates an MPI.Op specialized for the buffer dtype/count.
- ptp() now creates the MPI.Op on demand (and frees it after use) instead of relying on runtime autodetection within the callback.
- Added unit tests for the `_minmax_op` factory and made small doc updates.
- Ensured the reduction callback no longer raises exceptions at runtime and that dtypes/sizes are determined before registering the operation.

All local tests and pre-commit hooks ran successfully.

.github/workflows/codeql.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@f1f6e5f6af878fb37288ce1c627459e94dbf7d01 # v3.30.1
```
JuanPedroGHM: Those version number updates should be removed from the PR.

heat/core/communication.py
```diff
@@ -860,6 +861,46 @@ def op(sendbuf: MPI.memory, recvbuf: MPI.memory, datatype):
...
+            send_arr = (ctype * (2 * total_count + offset)).from_address(sendbuf.address)
+            recv_arr = (ctype * (2 * total_count + offset)).from_address(recvbuf.address)
+
+            # create torch views (count=2*total_count, offset=offset)
+            send_tensor = torch.frombuffer(
+                send_arr, dtype=dtype, count=2 * total_count, offset=offset
+            )
+            recv_tensor = torch.frombuffer(
+                recv_arr, dtype=dtype, count=2 * total_count, offset=offset
+            )
```
JuanPedroGHM, line 919: This won't be able to handle non-contiguous tensors, as no information on the stride is given. I would also pass that information in the same way as the _derived_op.

heat/core/statistics.py
```diff
@@ -1739,6 +1740,187 @@ def _create_sketch(
...
+            # keep MPS behavior consistent with other functions
+            if mins.is_mps:
+                mins = mins.float()
+                maxs = maxs.float()
+            else:
+                mins = mins.double()
+                maxs = maxs.double()
+        else:
+            mins = mins.to(torch.int64)
+            maxs = maxs.to(torch.int64)
```
JuanPedroGHM: Is this conversion to double precision necessary? The local arrays should already be cast to the same datatype, which would solve the MPS problem too. Unless torch functions unexpectedly change the datatype when calling amin and amax.

heat/core/statistics.py
```diff
@@ -1739,6 +1740,187 @@ def _create_sketch(
...
+            f"ptp: unexpected preview buffer size: preview.numel()={preview.numel()} "
+            "(must be non-zero and even)"
+        )
+    total_count = preview.numel() // 2
+
+    comm = x.comm
+
+    # Create a dtype/size-specific MPI.Op via the communicator's factory
+    # The communicator must provide `_minmax_op(dtype, total_count, offset=0)`
+    op = comm._minmax_op(preview_dtype, total_count, offset=0)
```
JuanPedroGHM: It should not be necessary to create a preview of the results to get the count and data type. It should be faster in most cases to calculate the resulting size (and the datatype should not change).

ivansherbakov9:
@JuanPedroGHM 
Thanks for the review!

Regarding the float -- double upcast that was present earlier:
I removed it because the MPI.Op is now created with the dtype taken from x.larray, so buffer interpretation is fixed up-front and extra upcasting is redundant. This way, dtype consistency is guaranteed across ranks without conversions inside the reduction callback.

heat/core/communication.py
ClaudiaComito: > This won't be able to handle non-contiguous tensors, as no information on the stride is given. I would also pass that information in the same way as the _derived_op.

@ivansherbakov9 thanks for your patience, is the non-contiguous case being handled?

ivansherbakov9: @ClaudiaComito 
Thanks for the review! Just letting you know that I won’t be able to continue work on this PR until next week. I’ll pick it up then.

JuanPedroGHM:
> @JuanPedroGHM Thanks for the review!
> 
> Regarding the float -- double upcast that was present earlier: I removed it because the MPI.Op is now created with the dtype taken from x.larray, so buffer interpretation is fixed up-front and extra upcasting is redundant. This way, dtype 
consistency is guaranteed across ranks without conversions inside the reduction callback.

Hi @ivansherbakov9,

I apologise for the long silence, I was on vacation. I will review this as soon as possible.

heat/core/statistics.py
```diff
@@ -1739,6 +1740,187 @@ def _create_sketch(
...
+
+    Examples
+    --------
+    >>> a = ht.array([[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]], dtype=ht.float32)
+    >>> ht.ptp(a)
+    DNDarray([11.], dtype=ht.float32, device=cpu:0, split=None)
+    >>> ht.ptp(a, axis=0)
+    DNDarray([9., 9., 9.], dtype=ht.float32, device=cpu:0, split=None)
+    >>> ht.ptp(a, axis=1)
+    DNDarray([2., 2., 2., 2.], dtype=ht.float32, device=cpu:0, split=None)
```
JuanPedroGHM, line 1778: An example of how the axis argument works with a tuple would be great.

heat/core/tests/test_statistics.py
```diff
@@ -1327,6 +1327,75 @@ def test_percentile_sketched(self):
...
+                sp = None
+            return sp
+
+        # 1D: comparison with NumPy and invariance to split
+        v = ht.random.rand(50)
+        v_np = v.copy().numpy()
+        self.assertAlmostEqual(ht.ptp(v), np.ptp(v_np), places=5)
+
+        v = ht.resplit(v, 0)
+        self.assertAlmostEqual(ht.ptp(v), np.ptp(v_np), places=5)
```
JuanPedroGHM: You can probably change all the ```assertAlmostEqual``` operations to ```assertEqual```. No floating point arithmetic happens.

JuanPedroGHM (commented):
Changes look good, there are just minor comments. I'll run the CI to see if something else comes up.

JuanPedroGHM (approved):

PR #2013, opened 2025-10-29
Tutorial for local outlier factor 

### LINKED ISSUE #1996: LOF tutorial
Create a tutorial that explains how the newly implemented local outlier factor works.

Hakdag97:
## Description
Establishment of a Jupyter notebook for a tutorial describing how to use the LOF.

## Type of change
- Documentation update

no


JuanPedroGHM (changes requested):
I could not run it, because the LOF code has not been merged yet. 

One thing that I would change is that I would save the notebook with some rendered results and plots. That way it will look complete on readthedocs.

JuanPedroGHM (approved):
Looks good :+1:

Issue #2004, opened 2025-10-23
[Bug]: by default save_hdf5 and load_hdf5 set the dtype to float32


Marc-Jindra:
### What happened?

When saving a dndarray with `dtype=ht.int32` using `save_hdf5`, the resulting h5 dataset has the type `float32`. This can be avoided by specifying `dtype=int` or `dtype=np.int32` (not `ht.int32` or `torch.int32` as those throw errors) but it should use the dtype of the dndarray by default.

The same issue exists for `load_hdf5`, where it does not use the type of the h5 dataset but sets it to `float32`.

### Code snippet triggering the error

```shell
import heat as ht
a = ht.arange(100, dtype=ht.int32)
ht.save_hdf5(a, 'test.h5', dataset='data')
b = ht.load_hdf5('test.h5', dataset='data')
print(b.dtype)
```

### Error message or erroneous outcome

```shell
No error just the wrong type
```

### Version

main (development branch)

### Python version

3.12

### PyTorch version

2.6

### MPI version

```shell

```

PR #2014, opened 2025-10-30
Set correct dtype when loading and saving hdf5

Marc-Jindra:
## Due Diligence

- General:
- Implementation:

## Description

Until now, when saving a dndarray with `dtype=ht.int32` using `save_hdf5`, the resulting h5 dataset has the type `float32`. This can be avoided by specifying `dtype=int` or `dtype=np.int32` (not `ht.int32` or `torch.int32` as those throw errors) but it should use the dtype of the dndarray by default.

The same issue exists for `load_hdf5`, where it does not use the type of the h5 dataset but sets it to `float32`.

This PR resolves those issues, while also solving the issue of not being able to specify `dtype=ht.float32` in `save_hdf5`.

Issue/s resolved: #2004 

## Changes proposed:

- Changed save_hdf5, so that by default, the hdf5 dataset has the same dtype as the saved data.
- Changed load_hdf5, so that by default, the dndarray has the same dtype as the loaded hdf5 dataset.
- Changed save_hdf5, so that specifying heat and float dtypes works.
-

## Type of change

yes / no

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.14.3
+    rev: v0.14.4
     hooks:
       # Run the linter.
       - id: ruff


ClaudiaComito (approved):
This looks great, thanks a lot @Marc-Jindra !

PR #2040, opened 2025-11-24
Add support for complex64 in QR decomposition

### LINKED ISSUE #2039: Make linalg.qr work with complex values
**Related**
Right now, when calling linalg.qr with a complex-valued DNDarray throws a TypeError saying:
`Array 'A' must have a datatype of float32 or float64, but has <class 'heat.core.types.complex64'>`

**Feature functionality**
The NumPy version is intended to work with both real and complex data. This would be beneficial for quantum mechanical calculations.

https://numpy.org/doc/2.3/reference/generated/numpy.linalg.qr.html

2400032275:
Fixes #2039

Updated the QR decomposition function to support complex64 data type in addition to float32 and float64.

## Due Diligence

- General:
- Implementation:

## Description

Issue/s resolved: #

## Changes proposed:

-
-
-
-

yes / no

### DIFF
--- a/.github/workflows/CommentPR.yml
+++ b/.github/workflows/CommentPR.yml
@@ -69,7 +69,7 @@ jobs:
         with:
           egress-policy: audit
 
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
       - name: 'Trigger Workflow'
         run: |
--- a/.github/workflows/ReceivePR.yml
+++ b/.github/workflows/ReceivePR.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           egress-policy: audit
 
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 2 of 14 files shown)


mrfh92:
@2400032275 thanks for taking up this issue :+1:. I have two suggestions if you want to complete this PR: 

* So far, you have added support for `complex64` (the complex counterpart of `float32`), but `complex128` (the complex double precision) is still excluded 
* In order to test whether QR really works with the additional datatypes you would need to add corresponding tests in `heat/core/linalg/tests/test_qr.py`, where in lines 14 and 81 the datatypes that are used for tests are defined. 

If you need any help, feel free to ask :)

heat/core/linalg/qr.py
```diff
@@ -94,11 +94,10 @@ def qr(
     if procs_to_merge == 0:
         procs_to_merge = A.comm.size
 
-    if A.dtype not in [float32, float64]:
+    if A.dtype not in [float32, float64, complex64]:
```
brownbaerchen, line 97: A better check would be
```python
from ..types import issubdtype, floating, complex
if not issubdtype(A.dtype, floating) and not issubdtype(A.dtype, complex):
```
which would automatically work with all precision versions of these types. In particular, ones that will be added in the future.

ClaudiaComito:
closing this, superseded by #2105

PR #2068, opened 2025-12-10
Update Acknowledgements in README

mrfh92:
* introduced two sections "current" and "past" 
* @ClaudiaComito can you please add the funding text for ScienceServe and (if necessary) the respective Helmholtz logo (if different from the current one)


mrfh92:
as discussed on mattermost, this will be part of refactoring of the docs

Issue #2084, opened 2026-01-05
[Bug]: Unable to install dependencies for running examples with Python>3.12


brownbaerchen:
### What happened?

Using `pip install .[examples]` with Python 3.13, I get `ModuleNotFoundError: No module named 'numpy.distutils'` while trying to install `scikit-learn~=0.24`. Apparently `numpy.distutils` was deprecated in Python 3.12, see [here](https://numpy.org/doc/stable/reference/distutils_status_migration.html).

The latest version of scikit-learn can be installed with no issues and removing the version specifier in `pyproject.toml` seems to do the trick. Then, I can use the above pip command to install the dependencies and, after setting up the datasets, run the examples.

I see no reason to restrict the version of scikit-learn here. If you agree I can do a quick PR, changing the dependency for the examples from `scikit-learn~=0.24` to `scikit-learn` to get the latest version.

### Code snippet triggering the error

```shell

```

### Error message or erroneous outcome

```shell
$ pip install ".[examples]"                                                                                                                                                                                                                         
Processing <path>/heat                                                                                                                                                                                                                    
  Installing build dependencies ... done                                                                                                                                                                                                                                       
  Getting requirements to build wheel ... done                                                                                                                                                                                                                                 
  Preparing metadata (pyproject.toml) ... done                                                                                                                                                                                                                                 
Requirement already satisfied: mpi4py>=3.0.0 in ./venvheat/lib/python3.13/site-packages (from heat==1.7.0.dev0) (4.1.1)                                                                                                                                                        
Requirement already satisfied: torch<2.10,~=2.0 in ./venvheat/lib/python3.13/site-packages (from heat==1.7.0.dev0) (2.9.1)                                                                                                                                                     
Requirement already satisfied: torchvision~=0.15 in ./venvheat/lib/python3.13/site-packages (from heat==1.7.0.dev0) (0.24.1)                                                                                                                                                   
Requirement already satisfied: scipy~=1.14 in ./venvheat/lib/python3.13/site-packages (from heat==1.7.0.dev0) (1.16.3)                                                                                                                                                         
Collecting scikit-learn~=0.24 (from heat==1.7.0.dev0)                                                                                                                                                                                                                          
  Using cached scikit-learn-0.24.2.tar.gz (7.5 MB)                                                                                                                                                                                                                             
  Installing build dependencies ... done                                                                                                                                                                                                                                       
  Getting requirements to build wheel ... done                                                                                                                                                                                                                                 
[...]

PR #2086, opened 2026-01-05
Removed version restriction for scikit-learn in `examples` dependencies

brownbaerchen:
## Due Diligence

- General:
- Implementation:

## Description

See the discussion in #2084 for the problem and solution. Also, I refactored how the paths to the datasets are set up as [`pkg_resources` is deprecated](https://setuptools.pypa.io/en/latest/pkg_resources.html) and the old solution actually did not work with my editable installation of heat. I did make sure that this works with an installation directly from PyPI. That is to say `pip install heat[h5py]; python examples/classification/demo_knn.py` does work.

Unrelated to the issue that this PR addresses: I see that the IO tests rely on being executed in a certain directory. As far as I can tell, this could be fixed by applying the same refactor from here, i.e. replacing
```
cls.HDF5_PATH = os.path.join(os.getcwd(), "heat/datasets/iris.h5")
```
by 
```
cls.HDF5_PATH = os.path.join(os.path.dirname(ht.__file__), "datasets/iris.h5")
```
let me know if I should do this or if there is any reason not to.

## Issue/s resolved:
#2084

## Changes proposed:

- Remove version restriction on scikit-learn in examples dependencies
- Refactor generation of paths to datasets in tutorials

## Type of change

Bug fix

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.14.10
+    rev: v0.14.11
     hooks:
       # Run the linter.
       - id: ruff
--- a/heat/core/tests/test_statistics.py
+++ b/heat/core/tests/test_statistics.py
@@ -1,5 +1,6 @@
 import numpy as np
 import torch
+import os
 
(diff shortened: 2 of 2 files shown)


mtar (approved):
Thank you. Looks fine to me as we are only using the dataset in scikit-learn now.

Issue #2069, opened 2025-12-12
[Bug]: errors in test_statistics.py


Hakdag97:
### What happened?

 test_statistics.py fails (in the main branch) if more than MPI process is used

### Code snippet triggering the error

```shell

```

### Error message or erroneous outcome

```shell

```

### Version

main (development branch)

### Python version

None

### PyTorch version

None

### MPI version

```shell

```

PR #2090, opened 2026-01-07
Using same random values on all MPI tasks in `test_statistics.py`

brownbaerchen:
## Due Diligence

- General:
- Implementation:

## Description
Running the file `test_statistics.py` from any directory had two issues: The paths to the datasets were incorrect and there is some funny business with the random seed where, depending on how you run the tests, the random seeds are set up differently.

I finally figured out why. `heat.random` has a global variable `__rng` which determines the behavior of the random generator on different MPI tasks. Specifically, when set to `Threefry`, `heat.random.seed` gives the same random seed on all ranks, whereas with `batchparallel`, you get the seed plus the rank.
So if you want to be sure the code is doing what you want, irrespective of what has been run before, you need to use
```
ht.random.set_state((<'Threefry' or 'batchparallel'>, seed, local_seed)) 
```
rather than just
```
ht.random.seed(seed)
```

I have to say I am not super happy with this because clearly it can lead to confusion. The names of allowed states of the `__rng` variable are not helpful in understanding the parallel behavior. I understand that both modes are needed, but maybe the API should be overhauled. If whoever reviews this agrees, feel free to open an issue or ping me to do so.

Issue/s resolved: #2069

## Changes proposed:

- Set random seed at the beginning of each test
- Made paths to datasets relative to heat installation rather than current working directory

## Type of change

 - Bug fix

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.14.10
+    rev: v0.14.11
     hooks:
       # Run the linter.
       - id: ruff
--- a/contributing.md
+++ b/contributing.md
@@ -50,10 +50,10 @@ git checkout main
 git pull upstream main
 ```
 
-* Install Heat from the checked out sources with:
+* Install Heat in editable mode from the checked out sources with:
 
 ```
-pip install .[hdf5,netcdf,dev]
+pip install -e .[hdf5,netcdf,dev]
 ```
 
 * The extra `dev` dependency pulls in additional tools to support the enforcement
@@ -149,7 +149,7 @@ If GitHub indicates that the branch of your PR can no longer be merged automatic
 * To measure the test coverage, install [codecov](https://github.com/codecov/codecov-python) and then run:
 
 ```
-mpirun -np <PROCESSES coverage run --source=heat --parallel-mode -m pytest heat/ && \
+mpirun -np <PROCESSES> coverage run --source=heat --parallel-mode -m pytest heat/ && \
     coverage combine && \
     coverage report && \
     coverage xml'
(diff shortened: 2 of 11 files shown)


mrfh92:
Regarding the tests, one also needs to be aware of line 22 in  `heat/core/tests/test_suites/basic_test.py` which sets a seed whenever starting a new test.

The rationale behind "batchparallel" was: threefry is very cool but has two main limitations: 
- The limit of possible generations is reached fast
- quite slow

The batchparallel option is also closer to Heats rationale to take process-local operations from Torch whenever possible. Only disadvantage: The actual random numbers depend on the number of processes, not only on the seed.

mrfh92:
Indeed, setting a seed requires differences for the two options: 
- one (global) seed for threefry
- one (global) seed that produces local seeds for Torch for the batchparallel option. 

In the current implementation, full reproducibility should be given for threefry and partial reproducibility (under same seed and number of processes/larray shapes) should be given for batchparallel. 

In my opinion, the larger problem with randomized tests is likely due to the fact that sometimes we set seeds, sometimes not, and additionally due to the recent change mentioned above the test class sets a seed as well.

brownbaerchen:
I am not disagreeing with having both Threefry and batchparallel. In fact, I don't know anything about generating random numbers or these methods in particular.
Actually, after giving it some thought, I think the current API seems fine and it was just used a bit sillily in the test. The `setUp` method @mrfh92 mentions sets a seed for heat and then the tests generates some random numbers with torch. `heat.seed` sets a seat for torch, but depending on the `__rng` global variable, it's either the same or not.
So, the "correct" way of seeding is: seed torch if you want to use torch directly or seed heat if you want to use heat.
Having different torch seeds on different ranks is perfectly reasonable. I remember running a simulation with random perturbations and thinking I have a bug because the solution repeated in space. But then it was just the same seed on all tasks...

ClaudiaComito:
> I think the current API seems fine and it was just used a bit sillily in the test. 

I think the team has settled on the formulation "for historical reasons" 😝 

These must be some of the first tests we implemented, even before the ht.random module existed. A nice refactoring was long overdue. Thanks @brownbaerchen !

Issue #2018, opened 2025-11-05
[Bug]: ```__repr__``` vs ```__str__``` behaviour.


JuanPedroGHM:
### What happened?

Our latest implementation of ```__repr__``` is making it difficult to work on interactive environments (IPython, Jupyter).

Goal should be that ```__repr__``` behaves like ```__str__```, if it detects that it is running on either of those. 

### Code snippet triggering the error

```shell

```

### Error message or erroneous outcome

```shell

```

### Version

main (development branch)

### Python version

None

### PyTorch version

None

### MPI version

```shell

```

PR #2091, opened 2026-01-07
Print contents of array in `__repr__`

brownbaerchen:
## Due Diligence

- General:
- Implementation:

## Description
Python uses `__repr__` and `__str__` to give textual information about objects. The intention is that `__repr__` gives output for developers and `__str__` gives output for users. In heat arrays, this means `__repr__` tells you, for each MPI rank, what kind of data is contained in the array, while `__str__` tells you the actual data.

This is all well and good but the problem is when people are using heat interactively, in which case they get the `__repr__` if they just print the array, but if they are users, they want to know the data and not the local shape.
For example:
```
python
import heat as ht
a = ht.arange(4)
a  # prints a.__repr__()
f"{a}"  # prints a.__str__()
```
when doing something like this, you tend to want to know the contents of the array in both cases.

In this PR, information about the contents of the array is appended. Before you got
```
$ mpirun -np 2 python -c "import heat as ht; a = ht.arange(3 * 4, split=0).reshape((3, 2, 2)); print(a.__repr__())"
<DNDarray(MPI-rank: 0, Shape: (3, 2, 2), Split: 0, Local Shape: (2, 2, 2), Device: cpu:0, Dtype: int32)>
<DNDarray(MPI-rank: 1, Shape: (3, 2, 2), Split: 0, Local Shape: (1, 2, 2), Device: cpu:0, Dtype: int32)>
```
with the changes introduced here, you get
```
$ mpirun -np 2 python -c "import heat as ht; a = ht.arange(3 * 4, split=0).reshape((3, 2, 2)); print(a.__repr__())" 
DNDarray(MPI-rank: 1, Shape: (3, 2, 2), Split: 0, Local Shape: (1, 2, 2), Device: cpu:0, Dtype: int32
         [[[ 8,  9],
           [10, 11]]])
DNDarray(MPI-rank: 0, Shape: (3, 2, 2), Split: 0, Local Shape: (2, 2, 2), Device: cpu:0, Dtype: int32
         [[[0, 1],
           [2, 3]],

          [[4, 5],
           [6, 7]]])
```

Compare this to the `__str__` output
```
$ mpirun -np 2 python -c "import heat as ht; a = ht.arange(3 * 4, split=0).reshape((3, 2, 2)); print(a)"           
DNDarray([[[ 0,  1],
           [ 2,  3]],

          [[ 4,  5],
           [ 6,  7]],

          [[ 8,  9],
           [10, 11]]], dtype=ht.int32, device=cpu:0, split=0)

```

Let me know what you think about this.

Issue/s resolved: #2018 

## Changes proposed:

- Print both debug information and content information in interactive sessions
- Test for correct output in `__repr__` with multiple processes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.14.10
+    rev: v0.14.11
     hooks:
       # Run the linter.
       - id: ruff
--- a/contributing.md
+++ b/contributing.md
@@ -50,10 +50,10 @@ git checkout main
 git pull upstream main
 ```
 
-* Install Heat from the checked out sources with:
+* Install Heat in editable mode from the checked out sources with:
 
 ```
-pip install .[hdf5,netcdf,dev]
+pip install -e .[hdf5,netcdf,dev]
 ```
 
 * The extra `dev` dependency pulls in additional tools to support the enforcement
@@ -149,7 +149,7 @@ If GitHub indicates that the branch of your PR can no longer be merged automatic
 * To measure the test coverage, install [codecov](https://github.com/codecov/codecov-python) and then run:
 
 ```
-mpirun -np <PROCESSES coverage run --source=heat --parallel-mode -m pytest heat/ && \
+mpirun -np <PROCESSES> coverage run --source=heat --parallel-mode -m pytest heat/ && \
     coverage combine && \
     coverage report && \
     coverage xml'
--- a/examples/classification/demo_knn.py
+++ b/examples/classification/demo_knn.py
@@ -5,13 +5,8 @@ import random
 import heat as ht
 from heat.classification.kneighborsclassifier import KNeighborsClassifier
 
-import pkg_resources
-
 # Load dataset from hdf5 file
-iris_path = pkg_resources.resource_filename(
-    pkg_resources.Requirement.parse("heat"), "heat/datasets/iris.h5"
-)
-
+iris_path = f'{os.path.dirname(ht.__file__)}/datasets/iris.h5'
 X = ht.load_hdf5(iris_path, dataset="data", split=0)
 
 # Generate keys for the iris.h5 dataset
--- a/examples/lasso/demo.py
+++ b/examples/lasso/demo.py
@@ -10,15 +10,11 @@ import heat.regression.lasso as lasso
 
 import plotfkt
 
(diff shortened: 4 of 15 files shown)


brownbaerchen:
I am kind of lost why this fails. On Jureca I get
```
(heat_venv) [me@jrlogin02 heat]$ python
Python 3.12.3 (main, Jul 26 2024, 17:40:49) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import heat as ht
/path/heat/heat/core/_config.py:108: UserWarning: Heat has CUDA GPU-support (PyTorch version 2.5.1+cu124 and `torch.cuda.is_available() = True`), but CUDA-awareness of MPI could not be detected. This may lead to performance degradation as direct MPI-communication between GPUs is not possible.
  warnings.warn(
>>> a = ht.arange(4, device='gpu')
>>> a
DNDarray([0, 1, 2, 3], dtype=ht.int32, device=gpu:0, split=None)
>>> a.__repr__(force_debug_output=True)
'<DNDarray(MPI-rank: 0, Shape: (4,), Split: None, Local Shape: (4,), Device: gpu:0, Dtype: int32)>'
>>> a.__str__() == a.__repr__()
True
```
Same result with cpu array and the tests pass both in serial as well as in parallel.

So it seems to work also on GPUs. Anyone any ideas? I will not spend time on this until we decide that we even want this behavior.

brownbaerchen:
We decided that we will try for a compromise. Rather than either printing the content or only the debug information about an array, in `__repr__`, we want both debug information and some content of the array. In particular, we need to watch out that no communication happens in `__repr__` to make sure we don't add deadlocks.

brownbaerchen:
> My minor suggestion is to add the tag 'data:' before showing the local data.

I added the tag, such that the example from the description now is:
```
$ mpirun -np 2 python -c "import heat as ht; a = ht.arange(3 * 4, split=0).reshape((3, 2, 2)); print(a.__repr__())"
DNDarray(MPI-rank: 1, Shape: (3, 2, 2), Split: 0, Local Shape: (1, 2, 2), Device: cpu:0, Dtype: int32, Data:
         [[[ 8,  9],
           [10, 11]]])
DNDarray(MPI-rank: 0, Shape: (3, 2, 2), Split: 0, Local Shape: (2, 2, 2), Device: cpu:0, Dtype: int32, Data:
         [[[0, 1],
           [2, 3]],

          [[4, 5],
           [6, 7]]])
```

Putting it in the line where the data begins somehow looked odd to me.

Issue #2097, opened 2026-01-13
[Bug]: CI does not test with the dependency versions it promises to


brownbaerchen:
### What happened?

The problem is that the specific module versions are set beforehand and then heat is installed, which may reinstall a different version of the dependency. See, for instance [here](https://github.com/helmholtz-analytics/heat/actions/runs/20914119951/job/60083420272#step:7:125).

This should be resolved by merging the installation of the dependencies and heat itself into a single command such that pip respects the constraints, i.e. sth like
```
pip install numpy==1.26 torch==2.2.2 torchvision==0.17.2 torchaudio==2.2.2 . --extra-index-url https://download.pytorch.org/whl/cpu
```

### Code snippet triggering the error

```shell

```

### Error message or erroneous outcome

```shell

```

### Version

main (development branch)

### Python version

None

### PyTorch version

None

### MPI version

```shell

```

PR #2098, opened 2026-01-13
Remove updating dependency versions when installing heat in CI

brownbaerchen:
## Description

As discussed in #2097, installing heat like
```
pip install numpy=1.26.0
pip install .
```
can lead to a newer version of the dependency being installed while installing heat, which is not what we want in the CI.
Instead, merging the two commands to 
```
pip install numpy=1.26.0 .
```
leads to pip respecting all the listed constraints in the installation and does not install a newer version of the dependencies.

Issue/s resolved: #2097 

## Changes proposed:

- Merge installation of dependencies with specific version and heat in CI

## Type of change

Bug fix

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -67,7 +67,6 @@ jobs:
       - name: Test
         run: |
           pip install pytest
-          pip install ${{ matrix.pytorch-version }} --extra-index-url https://download.pytorch.org/whl/cpu
-          pip install ${{ matrix.install-options }}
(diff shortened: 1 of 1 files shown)


ClaudiaComito:
OK this indeed solves the CI matrix problem, thanks @brownbaerchen , I'll merge and update the rest

brownbaerchen:
Seems like this accidentally fixed the deadlock with old dependency versions in the CI. So we don't need to drop support or do anything. 🥳

PR #2105, opened 2026-01-15
Added support for complex values in QR

### LINKED ISSUE #2039: Make linalg.qr work with complex values

brownbaerchen:
## Due Diligence

- General:
- Implementation:

## Description
As torch supports complex values in QR decomposition, only minimal changes are needed to make heat support it as well. Only the type check needs to be altered and the complex conjugate needs to be taken in a few places, that's it.

There is already a very similar PR open for this (#2040). However, it lacks testing. Here, I added tests. The diff is quite large because I split the test in subtests, but I actually changed very little there.

Issue/s resolved: #2039, #2040 

## Changes proposed:

- Allow complex values in QR

## Type of change

 - New feature

### DIFF
--- a/heat/core/linalg/qr.py
+++ b/heat/core/linalg/qr.py
@@ -10,7 +10,7 @@ from ..dndarray import DNDarray
 from ..manipulations import concatenate
 from .. import factories
 from .. import communication
-from ..types import float32, float64
+from ..types import issubdtype, floating, complex
 
 __all__ = ["qr"]
 
@@ -31,7 +31,7 @@ def qr(
     Parameters
     ----------
     A : DNDarray of shape (M, N), of shape (...,M,N) in the batched case
-        Array which will be decomposed. So far only arrays with datatype float32 or float64 are supported
+        Array which will be decomposed.
     mode : str, optional
         default "reduced" returns Q and R with dimensions (M, min(M,N)) and (min(M,N), N). Potential batch dimensions are not modified.
         "r" returns only R, with dimensions (min(M,N), N).
@@ -93,9 +93,10 @@ def qr(
         )
     if procs_to_merge == 0:
         procs_to_merge = A.comm.size
-
-    if A.dtype not in [float32, float64]:
-        raise TypeError(f"Array 'A' must have a datatype of float32 or float64, but has {A.dtype}")
+    if not issubdtype(A.dtype, floating) and not issubdtype(A.dtype, complex):
+        raise TypeError(
+            f"QR decomposition is only implemented for floating point or complex numbers, not {A.dtype}"
+        )
 
     QR = collections.namedtuple("QR", "Q, R")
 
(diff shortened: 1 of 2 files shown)


heat/core/linalg/qr.py
```diff
@@ -93,9 +93,10 @@ def qr(
         )
     if procs_to_merge == 0:
         procs_to_merge = A.comm.size
-
-    if A.dtype not in [float32, float64]:
-        raise TypeError(f"Array 'A' must have a datatype of float32 or float64, but has {A.dtype}")
+    if not issubdtype(A.dtype, floating) and not issubdtype(A.dtype, complex):
```
brownbaerchen: Turns out torch doesn't support float16, raising `NotImplementedError: "geqrf_cpu" not implemented for 'Half'` in `torch.linalg.qr`.
Do you want an explicit type check for this here or let torch raise the error?

ClaudiaComito: > Turns out torch doesn't support float16, raising `NotImplementedError: "geqrf_cpu" not implemented for 'Half'` in `torch.linalg.qr`. Do you want an explicit type check for this here or let torch raise the error?

try: except: and have Heat raise torch's error message?

brownbaerchen: I find torch's error message a bit cryptic, so I raise the error with different message that might be more expressive for heat users.

brownbaerchen: The issue with this `try ... except` stuff is that sometimes the local QR factorization is computed at one rank at a time, which causes deadlock if this fails. Therefore, I now check if the type is among a list specifically supported by heat. If that is not the case, I try computing the local QR factorization on all tasks. If this works, I raise a `NotImplementedError` that heat doesn't support the type, but torch does. If this doesn't work, I raise a `TypeError` as before.

ClaudiaComito: Nice!

heat/core/linalg/qr.py
```diff
@@ -94,8 +94,16 @@ def qr(
...
-        raise TypeError(f"Array 'A' must have a datatype of float32 or float64, but has {A.dtype}")
+    if A.dtype not in [float32, float64, complex64, complex128]:
+        try:
+            torch.linalg.qr(A.larray)
+        except (NotImplementedError, RuntimeError) as E:
+            raise TypeError(
+                f"Array 'A' must have datatype of float32, float64, complex64, or complex128, but has {A.dtype}"
+            ) from E
+        raise NotImplementedError(
+            f"`heat.linalg.qr` is not implemented for dtype {A.dtype}, but is supported by torch. Please open an issue on GitHub to get this implemented"
```
ClaudiaComito:
```suggestion
            f"`heat.linalg.qr` is not implemented for dtype {A.dtype}. Please open an issue on GitHub to get this implemented"
```

that a dtype is supported by torch might be TMI for the average user.  I'd leave that out

ClaudiaComito (dismissed):
This looks good, thanks @brownbaerchen .

ClaudiaComito (approved):
Looks good, thanks a lot!

PR #2113, opened 2026-01-22
Implement heat.linalg.solve

### LINKED ISSUE #2036: Implement linalg.solve from NumPy
This function is missing, but it should be straightforward to implement. It would be needed for our quantum mechanical simulations

https://numpy.org/devdocs/reference/generated/numpy.linalg.solve.html

brownbaerchen:
## Due Diligence

- General:
- Implementation:

## Description
This function directly solves linear systems $Ax=b$ for $x$. It supports batched solving and $A$ and $b$ may only be distributed in non-batched axes.
The function simply passes the local data to pytorch and makes sure the input and output make sense for heat.

Note that pytorch allows to pass `left=False` in `torch.linalg.solve` in order to solve $xA=b$ instead. I didn't implement this here because I wasn't sure if it's needed and didn't want to bother. If you want to see this here, I suggest splitting #2036 into sub issues and not doing this as part of this PR. @GioPede, please let us know if you need this. As `numpy.linalg.solve` does not have this option, I assumed you don't.

I basically copy pasted the docstring from pytorch. I have a question @ClaudiaComito: I left some stuff in that is relevant not for the heat code, but only for the pytorch code that is called underneath. Specifically, I am thinking about the docstring including limitation of datatypes and raising `RuntimeError` if the matrix is not invertible. I don't like this because if it changes in pytorch, our documentation is just lying. On the other hand, users don't want to have to look up the torch documentation for the heat function they are using. What is the policy for this in heat?

Issue/s resolved: #2036 

## Changes proposed:

- add heat.linalg.solve

## Type of change

 - New feature

no

### DIFF
--- a/heat/core/linalg/solver.py
+++ b/heat/core/linalg/solver.py
@@ -10,7 +10,7 @@ from .. import factories
 
 import torch
 
-__all__ = ["cg", "lanczos", "solve_triangular"]
+__all__ = ["cg", "lanczos", "solve", "solve_triangular"]
 
 
 def cg(A: DNDarray, b: DNDarray, x0: DNDarray, out: Optional[DNDarray] = None) -> DNDarray:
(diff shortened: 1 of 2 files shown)


heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+        >>> x.split
+        0
+        >>> ht.allclose((A @ x).resplit_(None), b, atol=1e-5)
+        True
+
+    .. _invertible:
+        https://en.wikipedia.org/wiki/Invertible_matrix#The_invertible_matrix_theorem
+    """
+    if not isinstance(A, DNDarray) or not isinstance(b, DNDarray):
+        raise TypeError(f"A and b need to be of type ht.DNDarray, but were {type(A)}, {type(b)}")
```
ClaudiaComito: see `sanitization` module (`sanitize_in()` in this case I believe)

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+    - If :attr:`A` has shape `(*, n, n)` and :attr:`B` has shape `(*, n)` (a batch of vectors) or shape
+      `(*, n, k)` (a batch of matrices or "multiple right-hand sides"), this function returns `X` of shape
+      `(*, n)` or `(*, n, k)` respectively.
+    - Otherwise, if :attr:`A` has shape `(*, n, n)` and  :attr:`B` has shape `(n,)`  or `(n, k)`, :attr:`B`
+      is broadcast to have shape `(*, n)` or `(*, n, k)` respectively.
+      This function then returns the solution of the resulting batch of systems of linear equations.
+
+    .. note::
+        A and b may only be distributed in the batch dimensions. If both are split, they must be split in the same axis.
+
```
ClaudiaComito: they must be split... along the same batch axis?

ClaudiaComito: or along matching batch axes?

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+        raise ValueError(f"b has incorrect shape of {b.shape} for A of shape {A.shape}")
+    elif b.shape[-2] == A.shape[-1]:
+        b_non_batched_axis = b.ndim - 2
+    else:
+        raise ValueError(f"b has incorrect shape of {b.shape} for A of shape {A.shape}")
+
+    # raise error if b is distributed in disallowed way
+    if b.is_distributed() and b.split == b_non_batched_axis:
+        raise ValueError(
+            f"b of shape {b.shape} with A of shape {A.shape} is split in {b.split} but may not be distributed in non-batched axis {b_non_batched_axis}"
```
ClaudiaComito: This might be a bit obscure. How about "A and b are not distributed along the same batch axis: A.shape, A.split, b.shape, b.split...."

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+        )
+
+    # raise errors if A is distributed in disallowed way
+    if A.is_distributed():
+        if A.split > A.ndim - 3:
+            raise ValueError(
+                f"A of dimension {A.ndim} is split in {A.split} but must not be distributed in the (non-batched) last two axes"
+            )
```
ClaudiaComito: I would raise NotimplementedError here (ht.linalg.solve not implemented yet on input distributed along the linalg axis: A.split = ...)

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+        elif A.split != b.split and b.is_distributed():
+            raise ValueError(f"Split of A and b must match, but got {A.split} and {b.split}")
```
ClaudiaComito, line 383: I'm not sure this holds as b.ndim < A.ndim, isn't it?

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+        0
+        >>> ht.allclose((A @ x).resplit_(None), b, atol=1e-5)
+        True
+
+    .. _invertible:
+        https://en.wikipedia.org/wiki/Invertible_matrix#The_invertible_matrix_theorem
+    """
+    if not isinstance(A, DNDarray) or not isinstance(b, DNDarray):
+        raise TypeError(f"A and b need to be of type ht.DNDarray, but were {type(A)}, {type(b)}")
+
```
ClaudiaComito, line 359: I would insert an early out for non-distributed `A` and `b` here, just passing to the torch operation

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+        if not isinstance(out, DNDarray):
+            raise TypeError(f"out needs to be of type ht.DNDarray, but is {type(out)}")
+        if out.shape != out_initalization["shape"]:
+            raise ValueError(
+                f"Expect out to have shape {out_initalization['shape']} with A of shape {A.shape} but got {out.shape}"
+            )
+        if out.split != out_initalization["split"]:
+            raise ValueError(
+                f"Expect out to be split along {out_initalization['split']} but got {out.split}"
+            )
```
ClaudiaComito: check out `sanitation.sanitize_out()`

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+    Letting :math:`\mathbb{K}` be :math:`\mathbb{R}` or :math:`\mathbb{C}`,
+    this function computes the solution :math:`X \in \mathbb{K}^{n \times k}` of the **linear system** associated to
+    :math:`A \in \mathbb{K}^{n \times n}, B \in \mathbb{K}^{n \times k}`, which is defined as
+
+    .. math:: AX = B
+
+    This system of linear equations has one solution if and only if :math:`A` is `invertible`_.
+    This function assumes that :math:`A` is invertible.
+
+    Supports inputs of float, double, cfloat and cdouble dtypes.
```
ClaudiaComito: We want to support int as well for consistency with numpy

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+    if A.is_distributed():
+        if A.split > A.ndim - 3:
+            raise ValueError(
+                f"A of dimension {A.ndim} is split in {A.split} but must not be distributed in the (non-batched) last two axes"
+            )
+        elif A.split != b.split and b.is_distributed():
+            raise ValueError(f"Split of A and b must match, but got {A.split} and {b.split}")
+
+    # figure out what the output vector looks like
+    out_initalization = {"dtype": b.dtype, "device": b.device}
```
ClaudiaComito: dtype should probably be `types.promote_types(a.dtype, b.dtype)`

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+            raise TypeError(f"out needs to be of type ht.DNDarray, but is {type(out)}")
+        if out.shape != out_initalization["shape"]:
+            raise ValueError(
+                f"Expect out to have shape {out_initalization['shape']} with A of shape {A.shape} but got {out.shape}"
+            )
+        if out.split != out_initalization["split"]:
+            raise ValueError(
+                f"Expect out to be split along {out_initalization['split']} but got {out.split}"
+            )
+
```
ClaudiaComito, line 419: Do we need to check that distribution is aligned? i.e. `sanitation.sanitize_distribution()`

tests/core/linalg/test_solver.py
```diff
@@ -136,6 +136,140 @@ def test_lanczos(self):
             A = ht.random.randn(10, 10, split=1)
             V, T = ht.lanczos(A, m=3)
 
+    def test_solve(self):
+        torch.manual_seed(42 + ht.communication.MPI_WORLD.rank)
+
+        types = [ht.float32, ht.float64, ht.complex64, ht.complex128]
```
ClaudiaComito, line 142: pls test with int as well

heat/core/linalg/solver.py
```diff
@@ -272,6 +272,157 @@ def lanczos(
...
+    # do the actual solving of local matrices in torch
+    torch.linalg.solve(A.larray, b.larray, left=True, out=out.larray)
```
ClaudiaComito: we should catch the potential RuntimeError here (not invertible matrices) and return a proper Heat error. Esp. because torch's RuntimeError might only happen on a subset of the processes

ClaudiaComito (changes requested):
Thanks a lot for addressing this @brownbaerchen . I have a few comments, see below. In general I'd rather avoid copy-pasting documentation from numpy or torch, although obviously they will be very similar. If we are copy-pasting, we should point out where from.


Also tagging @GioPede who requested the feature and might want to check it out as well.

heat/core/linalg/solver.py
brownbaerchen: `torch.linag.solve`, which this function relies on, does not support integers. I will raise a `NotImplementedError` in the heat function telling people to open an issue on GitHub if they need this.

ClaudiaComito: I get it, but this is not how Heat works. We cast any integer input to float and pass it to the torch function.

heat/core/linalg/solver.py
brownbaerchen: I don't really see a way to make this more simple because of the `out` argument.
What the code is really doing, apart from just calling the torch operation, is make sure `out` has correct shape and that all the splits are compatible. If everything is non-distributed, we can forgo the split checks but we still need to make sure we have a proper out. Only if no out is passed, we can just wrap the torch output in a DNDarray.
I am not convinced that we really gain anything from treating this special case separately.

heat/core/linalg/solver.py
brownbaerchen: There are a few cases:
 - 1: A has shape (n, n) and b has shape (n, k)
 - 2: A has shape (k, n, n) and b has shape (n, k)
 - 3: A has shape (k, n, n) and b has shape (n, p)
 - 4: A has shape (k, n, n) and b has shape (k, n)
 - 5: A has shape (k, n, n) and b has shape (k, n, p)

What this is targeting is cases four. A has one more dimension in a non-batched axis which it must not be distributed in, both may only be distributed in the first axis here. (there could be more batched dimensions, which doesn't change this)
It probably is possible to have A split in 0 and b in 1 in case two or A split in 0 and b in 2 in case 5 if k = p. But I thought that is so niche that we can just disallow it.

heat/core/linalg/solver.py
brownbaerchen: Does it ever happen that two arrays of the same shape are split in the same axis but distributed differently? If so, we may need to do this. But it's non-trivial because `out` doesn't necessarily have the same shape as `b`, so it's not obvious to me what lshape_map to distribute this to.

heat/core/linalg/solver.py
brownbaerchen: That is a good point. But: What is a proper heat error?

heat/core/linalg/solver.py
brownbaerchen: I don't really know what to compare `out` to in metrics other than shape and split which are covered by `sanitize_out`. I think this is probably fine..?

ClaudiaComito (dismissed):
👍🏼

brownbaerchen:
@ClaudiaComito you need to reapprove after I resolved the merge conflict from moving the tests.

ClaudiaComito (approved):

PR #2114, opened 2026-01-22
Added `heat.linalg.matrix_exp`

### LINKED ISSUE #2037: Implement scipy.linalg.expm from SciPy
This is a function needed for many quantum mechanical simulations. 

https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.expm.html

brownbaerchen:
## Due Diligence

- General:
- Implementation:

## Description
Added function to compute matrix exponential. May only be distributed in batched dimensions and simply calls torch equivalent on the local data underneath.

I wasn't sure where to put this, so I put it in a new file `heat/core/linalg/other.py` I can also put in `heat/core/linalg/basics` if you prefer.

The matrix exponential is not listed in the python array API and `scipy.linalg.expm` and `torch.linalg.matrix_exp` have different names for this. <del>I chose the torch version since heat is based on torch and I find the name more descriptive, even though @GioPede mentioned the scipy name in #2037. </del> heat supports both, i.e. the function is implemented as `heat.linalg.matrix_exp` and has alias `heat.linalg.expm`.

Note that the same questions about documentation as in #2113 apply here. Feel free to answer in one of the PRs and I assume it applies to both.

Issue/s resolved: #2037

## Changes proposed:

- Added `heat.linalg.matrix_exp`
- Added `heat.linalg.expm` as an alias to match scipy naming scheme 

## Type of change

 - New feature

no


mtar:
Hi, in other cases where the the functions mostly differ by name, we try to support both by using an alias or a call to the other one. Is it also an option here?

brownbaerchen:
> Hi, in other cases where the the functions mostly differ by name, we try to support both by using an alias or a call to the other one. Is it also an option here?

Indeed it is an option and I added it. So now you can do
```python
import heat as ht
import scipy
import torch
import numpy as np

A = ht.random.randn(2, 2, dtype=ht.float64)
torch.allclose(ht.linalg.matrix_exp(A).larray, torch.linalg.matrix_exp(A.larray))
np.allclose(ht.linalg.expm(A).larray, scipy.linalg.expm(A.larray))
```

heat/core/linalg/other.py
```diff
@@ -0,0 +1,62 @@
...
+        >>> A[1, :, :] = 2 * ht.eye((2, 2))
+        >>> ht.linalg.matrix_exp(A)
+        DNDarray([[[2.7183, 0.0000],
+           [0.0000, 2.7183]],
+
+          [[7.3891, 0.0000],
+           [0.0000, 7.3891]]], dtype=ht.float32, device=cpu:0, split=0)
+    """
+    if not isinstance(A, DNDarray):
+        raise TypeError(f"A needs to be of type ht.DNDarray, but was {type(A)}")
```
mtar: We have [sanitize_in](https://heat.readthedocs.io/en/stable/_modules/heat/core/sanitation.html#sanitize_in) that does the same job.

heat/core/linalg/other.py
```diff
@@ -0,0 +1,62 @@
...
+    if A.is_distributed() and A.split >= A.ndim - 2:
+        raise ValueError(
+            f"A of shape {A.shape} may only be distributed in batched dimensions but is distributed in {A.split}"
+        )
+    out = ht.empty_like(A)
+    out.larray[...] = torch.linalg.matrix_exp(A.larray)
+    return out
+
+
+expm = matrix_exp  # provide alias with name of scipy equivalent
```
mtar: By adding a docstring like """Alias for :py:func:`matrix_exp`""" on the next line, it will also be listed in the documentation.

brownbaerchen: Nice! I didn't know this.

heat/core/linalg/other.py
```diff
@@ -0,0 +1,64 @@
...
+          [[7.3891, 0.0000],
+           [0.0000, 7.3891]]], dtype=ht.float32, device=cpu:0, split=0)
+    """
+    sanitize_in(A)
+
+    if A.is_distributed() and A.split >= A.ndim - 2:
+        raise ValueError(
+            f"A of shape {A.shape} may only be distributed in batched dimensions but is distributed in {A.split}"
+        )
+    out = ht.empty_like(A)
```
mtar: Please import the function directly from the relative path such that the import of the whole library can be avoided.

mtar (changes requested):
I think the function can be put in basics. I don't see a reason to create a new file for it.

mtar (dismissed):
Looks good, thanks.

heat/core/linalg/basics.py
```diff
@@ -1429,6 +1431,59 @@ def matrix_norm(
...
+        raise ValueError(
+            f"A of shape {A.shape} may only be distributed in batched dimensions but is distributed in {A.split}"
+        )
+    out = factories.empty_like(A)
+    out.larray[...] = torch.linalg.matrix_exp(A.larray)
+    return out
+
+
+expm = matrix_exp  # provide alias with name of scipy equivalent
+"""Alias for :py:func:matrix_exp"""
```
mtar: The rendering in the doc is wrong. Likely because of the missing backticks
```suggestion
"""Alias for :py:func:`matrix_exp`"""
```

mtar (approved):

PR #2115, opened 2026-01-23
Discussion template for using Heat as backend

brownbaerchen:
I want to start bringing the discussions page back to life and I think a good place to start is some templates.
In this PR, I add a template for porting a code to heat.

The way I imagine this is that anyone who thinks their code is slow or they cannot do the science they want because of memory limitations or whatever and they hear about heat, they can just use this template to reach out. Then, we get in touch with them and either find out quickly that heat is not the cure for their ailment with the information they put in the template, or help them out.

I don't really know what people want to say / what we need to hear. So this is a draft that I would like to receive comments on. If you don't like this at all, we can also not start using discussions for this. I am aware that some people might be hesitant to enter all this information publicly, for instance.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -34,7 +34,7 @@ body:
       description: What version of Heat are you running?
       options:
         - main (development branch)
-        - 1.7.x
+        - 1.8.x
         - other
     validations:
       required: true
@@ -44,10 +44,10 @@ body:
       label: Python version
       description: What Python version?
       options:
-        - 3.14
-        - 3.13
-        - 3.12
-        - 3.11
+        - '3.14'
+        - '3.13'
+        - '3.12'
+        - '3.11'
     validations:
       required: true
   - type: dropdown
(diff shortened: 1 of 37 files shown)


.github/DISCUSSION_TEMPLATE/heat_as_backend.yaml
```diff
@@ -0,0 +1,50 @@
...
+      label: What motivates you to switch to heat?
+      description: |
+        Please explain if you run out of memory or want to achieve speedup or any other way in which heat might help.
+        If there is a critical part of your code that is currently the main bottleneck, please explain it briefly.
+  - type: textarea
+    id: current-base
+    attributes:
+      label: What is your code currently built on?
+      description: numpy, scipy, pytorch, cupy, ...? Please sketch out what functionality you use from these libraries.
+  - type: textarea
```
brownbaerchen, line 29: @JuanPedroGHM: Add field on indented hardware accelerator

brownbaerchen:
After incorporating the feedback from @JuanPedroGHM, I think we can now merge this. Unless @ClaudiaComito wants to give feedback, that is.

Issue #2135, opened 2026-02-19
[Bug]: Unexpected masking behavior


brownbaerchen:
### What happened?

When masking a heat array with only one True entry, heat returns the entire (aka unmasked) array rather than the one entry that the mask specifies. This is, of course, unlike the behavior of numpy and pytorch. See the code snippet below for a minimal failing example and comparison with numpy and pytorch.

### Code snippet triggering the error

```shell
import numpy as np
import heat as ht

a = np.arange(12).reshape((4, 3))  # some numpy data
a_heat = ht.array(a)  # same data in heat

# generate a mask with two True entries and zeros else
mask = np.zeros_like(a).astype(bool)
mask[0, 1] = True
mask[2, 0] = True

mask_heat = ht.array(mask)  # make a heat copy of the mask
print(mask)
"""
[[False  True False]
 [False False False]
 [ True False False]
 [False False False]]
"""

print(a[mask])
"""
[1 6]
"""

print(a_heat[mask_heat])
"""
DNDarray([1, 6], dtype=ht.int64, device=cpu:0, split=None)
"""

# so far, so good. Now on to the bug

# now generate a mask with only one True entry and zeros else
mask = np.zeros_like(a).astype(bool)
mask[0, 1] = True

mask_heat = ht.array(mask)  # make a heat copy of the mask
print(mask)
"""
[[False  True False]
 [False False False]
 [False False False]
 [False False False]]
"""

print(a[mask])
"""
[1]
"""

print(a_heat[mask_heat])
"""
DNDarray([[[ 0,  1,  2],
           [ 3,  4,  5],
           [ 6,  7,  8],
           [ 9, 10, 11]]], dtype=ht.int64, device=cpu:0, split=None)
"""

print(a_heat.larray[mask_heat.larray])
"""
tensor([1])
"""
```

### Error message or erroneous outcome

```shell

```

### Version

main (1.8.0-dev)
### Python version

3.13.11

### PyTorch version

2.10

### MPI version

```shell

```

brownbaerchen:
Would this be addressed by #938, @ClaudiaComito? Not sure if it's a good idea for me to debug this now when we want to discuss the advanced indexing stuff soon.

ClaudiaComito:
this will be solved in #938

PR #2138, opened 2026-02-20
Fix bug in `Heat.indexing.nonzero`

### LINKED ISSUE #2136: [Bug]: Bug in `heat.indexing.nonzero`
### What happened?

The function `heat.indexing.non_zero` can return strange data in some cases. In the example below, it returns a DNDarray of shape (2,), with an larray of shape (1,2). This would correspond to kind of reverse distribution where the total global data is smaller than local data and is totally wrong.

It seems to me that this bug is causing the bug documented in #2135, so fixing this one could hopefully also fix that one.

```shell
import heat as ht

a = ht.zeros((4, 3), dtype=ht.bool)
a[1, 2] = True
non_zero = ht.indexing.nonzero(a)
print(non_zero.__repr__())
"""
DNDarray(MPI-rank: 0, Shape: (2,), Split: None, Local Shape: (1, 2), Device: cpu:0, Dtype: int64, Data:
         [[1, 2]])
"""
```

```shell

```

### Version

main (development branch)

### Python version

3.13

### PyTorch version

2.10

### MPI version

```shell

```

brownbaerchen:
## Due Diligence

- General:
- Implementation:

## Description
This fixes the bugs documented in #2136 and #2135. It makes the function `heat.indexing.nonzero` compatible with the sanity checks proposed in #2137.

Issue/s resolved: #2136, #2135 

## Changes proposed:

- Refactor and bugfix in `heat.indexing.nonzero`

## Type of change

 - bug fix

no

### DIFF
--- a/heat/core/tests/test_indexing.py
+++ b/heat/core/tests/test_indexing.py
@@ -23,6 +23,15 @@ class TestIndexing(TestCase):
         a[nz] = 10.0
         self.assertEqual(ht.all(a[nz] == 10), 1)
 
+        # edge case: single non-zero element
+        for split in [None, 1]:
+            a = ht.zeros((4, 3), dtype=ht.bool, split=split)
+            a[1, 2] = True
+            nz = ht.indexing.nonzero(a)
+            self.assertEqual(nz.gshape, (1, 2))
+            self.assertTrue(ht.allclose(a[nz], a[a]))
+
+
     def test_where(self):
         # cases to test
         # no x and y
--- a/heat/core/tests/test_indexing.py
+++ b/heat/core/tests/test_indexing.py
@@ -24,10 +24,12 @@ class TestIndexing(TestCase):
         self.assertEqual(ht.all(a[nz] == 10), 1)
 
         # edge case: single non-zero element
-        for split in [None, 1]:
+        for split in [None, 0, 1]:
             a = ht.zeros((4, 3), dtype=ht.bool, split=split)
             a[1, 2] = True
             nz = ht.indexing.nonzero(a)
+            a.resplit_(None)
+            nz.resplit_(None)
             self.assertEqual(nz.gshape, (1, 2))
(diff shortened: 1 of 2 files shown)


heat/core/tests/test_indexing.py
```diff
@@ -23,6 +23,15 @@ def test_nonzero(self):
         a[nz] = 10.0
         self.assertEqual(ht.all(a[nz] == 10), 1)
 
+        # edge case: single non-zero element
+        for split in [None, 1]:
+            a = ht.zeros((4, 3), dtype=ht.bool, split=split)
+            a[1, 2] = True
+            nz = ht.indexing.nonzero(a)
+            self.assertEqual(nz.gshape, (1, 2))
+            self.assertTrue(ht.allclose(a[nz], a[a]))
```
mtar: The edge case for split=1 doesn't work. All local tensors have no data after indexing.
```
print(a.comm.rank, a[nz].larray)
print(a.comm.rank, a[a].larray)
```
```
0 tensor([], dtype=torch.bool)
1 tensor([], dtype=torch.bool)

0 tensor([], dtype=torch.bool)
1 tensor([], dtype=torch.bool)
```

mtar (changes requested):
Thanks, the PR looks fine. The added edge case doesn't work properly. Something's wrong with the indexing.

heat/core/tests/test_indexing.py
brownbaerchen: That's true, good catch!

This PR is kind of working on two issues. The first is in `heat.indexing.nonzero` and the second is in the advanced indexing. Turns out this fixes the advanced indexing bug only when the array is non-distributed. 
However, the primary function of this PR, which is fixing the bug in nonzero is still valid. I modified the test to index only non-distributed data. Note that this test fails on the current main.

I suggest we merge this anyways and hope that #938 will fix the indexing.

mtar (approved):

PR #2133, opened 2026-02-18
Skip readthedocs CI if codebase is not affected by PR

ClaudiaComito:
## Due Diligence

- General:
    - [x]  **title** of the PR is suitable to appear in the [Release Notes](https://github.com/helmholtz-analytics/heat/releases/latest)
- Implementation:
    - [ ] unit tests: all split configurations tested
    - [ ] unit tests: multiple dtypes tested
    - [ ] **NEW** unit tests: MPS tested (1 MPI process, 1 GPU)
    - [ ] benchmarks: created for new functionality
    - [ ] benchmarks: performance improved or maintained
    - [x] documentation updated where needed

## Description

This PR modifies `.readthedocs.yaml` to skip the ReadTheDocs test build for those PRs that don't modify the codebase (dependencies updates, GH actions etc.)

Background: merging dependabot PRs has become a huge pain because of (among other things) the multiple ReadTheDocs builds trying and failing to run concurrently.  

Issue/s resolved: #2070 

## Changes proposed:

- searches diff for changes in relevant folders (heat/, doc/, examples/, tutorials/). If no changes, cancel readthedocs CI 
-
-
-

#### Does this change modify the behaviour of other functions? If so, which?
no

PR #2144, opened 2026-02-23
Revert "Skip readthedocs CI if codebase is not affected by PR"

ClaudiaComito:
Reverts helmholtz-analytics/heat#2133

That change a) doesn't work and b) anyway would still take a too long runtime to figure out that it should skip the build

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -14,28 +14,6 @@ build:
     - pandoc
     - libopenmpi-dev
 
-  jobs:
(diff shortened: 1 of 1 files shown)


JuanPedroGHM (approved):

brownbaerchen:
Skipping doesn't work, but we don't care..

ClaudiaComito:
Closed as .readthedocs.yaml has been modified in the meantime

PR #2139, opened 2026-02-21
Leaner CI matrix + `test_io` deadlock

ClaudiaComito:
## Due Diligence

- General:
    - [x]  **title** of the PR is suitable to appear in the [Release Notes](https://github.com/helmholtz-analytics/heat/releases/latest)
- Implementation:
    - [x] unit tests: all split configurations tested
    - [x] unit tests: multiple dtypes tested
    - [ ] **NEW** unit tests: MPS tested (1 MPI process, 1 GPU)
    - [ ] benchmarks: created for new functionality
    - [ ] benchmarks: performance improved or maintained
    - [ ] documentation updated where needed

## Description

This PR implements a leaner CI matrix as discussed in #2134. The race condition + deadlock arising in test_io with Python 3.14 has also been solved.  

Issue/s resolved: #2134 

## Changes proposed:

- refactor `ci.yaml` as per #2134 . Runs for every commit as requested by @brownbaerchen 
- keep the full matrix in `ci_full.yaml`. This will run when a PR is merged into `main`, if the PR affects the codebase
- in `test_io.test_load_zarr_group` prevent race condition by inserting a Barrier before starting to write out the zarr group. Might be related to [PEP 703](https://peps.python.org/pep-0703/) in combination with Zarr V3's asynchronous I/O. 
-

## Type of change

- Bug fix (non-breaking change which fixes an issue) + GH action refactoring

## Memory requirements

DNA
## Performance

DNA
#### Does this change modify the behaviour of other functions? If so, which?
no

PR #2148, opened 2026-02-24
Add MPI barrier before all tests

brownbaerchen:
The CI currently runs into deadlocks with some IO tests. We have not understood why, but seemingly adding some MPI barriers helps. In this PR, I simply add a global MPI barrier before every test that is derived from the basic `TestCase` using the `setUp` function.
Since I cannot reproduce the problem locally and we don't understand the problem, I have no clue if this fixes the issue. Nevertheless, I don't think this hurts and I'd rather do this in a centralized place for all tests than randomly spamming barriers in the code.
Schauen wir mal was wird.

Looking forward to being able to just do this on my fork without requiring a review from somebody else once we have merged #2139 ;)

### DIFF
--- a/heat/core/tests/test_suites/basic_test.py
+++ b/heat/core/tests/test_suites/basic_test.py
@@ -19,6 +19,7 @@ class TestCase(unittest.TestCase):
     envar: Optional[str] = None
 
     def setUp(self) -> None:
+        ht.MPI_WORLD.Barrier()
         seed(42)
 
     @classmethod


JuanPedroGHM:
Should we remove some other barriers at the start of tests within this PR?

brownbaerchen:
> Should we remove some other barriers at the start of tests within this PR?

Actually, I didn't find any places where there is a barrier at the **start** of the test. Only at the end. This is very similar, but since we don't understand the problem, I don't want to do everything at once. Can you approve this PR, such that the CI runs and we see if this fixes the problem at all? Then we can try removing the barriers at the end of IO tests.

brownbaerchen:
So, this is not the solution, sadly. Let's go with the barrier in a specific place suggested in #2147 then..

PR #2150, opened 2026-02-24
Fix: Missing sync for sent messages in `get_halo`

### LINKED ISSUE #2108: [Bug]: Halo synchronization not fully finished - random numbers in halo
### What happened?

My convolution2d tests with larger signals and kernels failed inconsistently during array comparison to ground truth, due to unrealisticly large integer numbers in the output array, unrelated to the input signal and kernel.
Numbers were traced back to halo communications, where they appear for the first time, propagating through the convolution:
Thus, I conclude, that the halo computation is potentially incomplete leaving leftovers from a torch.empty() initalization or something similar.
Might be related to a padding call earlier in the code.

Image
Image

Branch: 
feature920/distributed-2D-convolution
Commit Hash: [03a4662](https://github.com/helmholtz-analytics/heat/pull/1937/commits/03a4662f3fe9e2ef21a8e2bb7fd77b8a89b56def)

Only (but still inconsistently) appears if run on 3 processes

Files:
heat/core/signal.py Lines 720 - 743
heat/core/tests/test_signal.py `test_convolve2d_large_signal_and_kernel_modes`

Code in heat/core/signal.py
```shell
# pad signal with zeros
    pad_array = ((pad_size[0], pad_size[0]), (pad_size[1], pad_size[1]))
    a = pad(a, pad_array)
    print(a.comm.rank, "Padd area halo prev", a.larray[-20:, :])
    print(a.comm.rank, "Padd area halo next", a.larray[0:20, :])
    # CF: Necessary for convolution2d
    a.comm.Barrier()
    # print("lshape map", a.comm.rank, a.lshape_map, v.lshape_map)
    # no batch processing
    if a.is_distributed():
        if (v.lshape_map[:, a.split] > a.lshape_map[:, a.split]).any():
            raise ValueError(
                "Local chunk of filter weight is larger than the local chunks of signal"
            )

        # compute halo size
        halo_size = int(v.lshape_map[0][a.split]) // 2

        # fetch halos and store them in a.halo_next/a.halo_prev
        a.get_halo(halo_size)

        print(a.comm.rank, "Halo prev", a.halo_prev)
        print(a.comm.rank, "Halo next", a.halo_next)
        # apply halos to local array
        signal = a.array_with_halos
```

```shell
heat/core/tests/test_signal.py:1523: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
heat/core/tests/test_suites/basic_test.py:158: in assert_array_equal
    self.assertTrue(is_allclose == heat_array.comm.size)
E   AssertionError: tensor(False) is not true
```

### Version

other

### Python version

3.12

### PyTorch version

2.7

### MPI version

```shell
OpenMPI: 4.1.6
mpi4py: 4.1.1
```

JuanPedroGHM:
## Due Diligence

- General:
- Implementation:

## Description

One `Isend` request object was not being waited on, leading to raise conditions when immediately overwriting the sent buffer.

Issue/s resolved: #2108

## Changes proposed:

- Added the Request object to the `req_list`, to ensure waiting.
- Refactored `test_get_halo` for easier debugging
- Changed the recv buffers to `empty` instead of `zero` (better performance)

## Type of change

## Performance

Might affect performance, as waiting needs to take place. If we want to improve performance, we need to implement a truly `async_get_halo`. But that is a different issue, and there is no need at the moment.

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -436,7 +436,6 @@ class DNDarray:
             a_next = self.__prephalo(-halo_size, None)
             res_prev = None
             res_next = None
-
             req_list = []
 
             # exchange data with next populated process
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -442,9 +442,9 @@ class DNDarray:
             # exchange data with next populated process
             if prev:
                 if rank != last_rank:
-                    self.comm.Isend(a_next, next_rank)
+                    req_list.append(self.comm.Isend(a_next, next_rank))
                 if rank != first_rank:
-                    res_prev = torch.zeros(
+                    res_prev = torch.empty(
                         a_prev.size(), dtype=a_prev.dtype, device=self.device.torch_device
                     )
                     req_list.append(self.comm.Irecv(res_prev, source=prev_rank))
@@ -453,7 +453,7 @@ class DNDarray:
                 if rank != first_rank:
                     req_list.append(self.comm.Isend(a_prev, prev_rank))
                 if rank != last_rank:
-                    res_next = torch.zeros(
+                    res_next = torch.empty(
                         a_next.size(), dtype=a_next.dtype, device=self.device.torch_device
                     )
                     req_list.append(self.comm.Irecv(res_next, source=next_rank))
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -387,7 +387,7 @@ class DNDarray:
 
         return self.__array[tuple(ix)].clone()
 
-    def get_halo(self, halo_size: int, prev: bool = True, next: bool = True) -> torch.Tensor:
+    def get_halo(self, halo_size: int, prev: bool = True, next: bool = True):
         """
         Fetch halos of size ``halo_size`` from neighboring ranks and save them in ``self.halo_next/self.halo_prev``.
 
@@ -404,12 +404,12 @@ class DNDarray:
             raise TypeError(
                 f"halo_size needs to be of Python type integer, {type(halo_size)} given"
             )
-        if halo_size < 1:
+        if halo_size < 0:
             raise ValueError(
-                f"halo_size needs to be a positive Python integer, {type(halo_size)} given"
+                f"halo_size needs to be a non-negative Python integer, {halo_size} given"
             )
 
-        if self.is_distributed():
+        if self.is_distributed() and halo_size > 0:
             # gather lshapes
             lshape_map = self.lshape_map
             rank = self.comm.rank
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -404,12 +404,12 @@ class DNDarray:
             raise TypeError(
                 f"halo_size needs to be of Python type integer, {type(halo_size)} given"
             )
-        if halo_size < 0:
+        if halo_size < 1:
             raise ValueError(
                 f"halo_size needs to be a positive Python integer, {type(halo_size)} given"
             )
 
-        if self.is_distributed() and halo_size > 0:
+        if self.is_distributed():
             # gather lshapes
             lshape_map = self.lshape_map
             rank = self.comm.rank
--- a/heat/core/tests/test_dndarray.py
+++ b/heat/core/tests/test_dndarray.py
@@ -26,14 +26,14 @@ class TestDNDarray(TestCase):
             ht.equal(int16_tensor & int16_vector, ht.bitwise_and(int16_tensor, int16_vector))
         )
 
-    def test_gethalo_split1_random(self):
-        """Test get_halo with randomly generated data split along axis 1."""
+    def test_gethalo_split1(self):
+        """Test get_halo with data split along axis 1."""
         rows, cols = 4, 6 * max(ht.MPI_WORLD.size, 2)
         data_np = np.arange(rows * cols).reshape((rows, cols))
         data = ht.array(data_np, split=1)
 
         if not data.is_distributed():
-            return
+            self.skipTest("Need data to be distributed")
 
         halo_size = min(2, data.lshape[1])
         data.get_halo(halo_size)
@@ -68,14 +68,14 @@ class TestDNDarray(TestCase):
             expected_cols += halo_size
         self.assertEqual(data.array_with_halos.shape, (rows, expected_cols))
 
-    def test_gethalo_split0_random(self):
-        """Test get_halo with randomly generated data split along axis 0."""
+    def test_gethalo_split0(self):
+        """Test get_halo with data split along axis 0."""
         rows, cols = 6 * max(ht.MPI_WORLD.size, 2), 4
         data_np = np.arange(rows * cols).reshape((rows, cols))
         data = ht.array(data_np, split=0)
 
         if not data.is_distributed():
-            return
+            self.skipTest("Need data to be distributed")
 
         halo_size = min(2, data.lshape[0])
         data.get_halo(halo_size)
@@ -156,7 +156,7 @@ class TestDNDarray(TestCase):
     def test_gethalo_no_data_on_process(self):
         """Test get_halo when some processes have no local data."""
         if ht.MPI_WORLD.size < 3:
-            return
+            self.skipTest("Need at least 3 processes")
 
         data_np = np.arange(2 * 12).reshape(2, 12)
         data = ht.array(data_np, split=0)
--- a/heat/core/tests/test_dndarray.py
+++ b/heat/core/tests/test_dndarray.py
@@ -303,6 +303,7 @@ class TestDNDarray(TestCase):
 
         halo_size = 50
         data.get_halo(halo_size)
+        array_with_halos = data.array_with_halos
 
         rank = data.comm.rank
         lshape_map = data.lshape_map
@@ -313,7 +314,7 @@ class TestDNDarray(TestCase):
                 data_np[displacements[rank] - halo_size : displacements[rank], :],
                 device=data.device.torch_device,
             )
-            self.assertTrue(torch.all(data.halo_prev == expected_prev))
+            self.assertTrue(torch.all(array_with_halos[:halo_size, :] == expected_prev))
         else:
             self.assertIsNone(data.halo_prev)
 
@@ -322,7 +323,7 @@ class TestDNDarray(TestCase):
                 data_np[displacements[rank + 1] : displacements[rank + 1] + halo_size, :],
                 device=data.device.torch_device,
             )
-            self.assertTrue(torch.all(data.halo_next == expected_next))
+            self.assertTrue(torch.all(array_with_halos[-halo_size:, :] == expected_next))
         else:
             self.assertIsNone(data.halo_next)
 
@@ -331,7 +332,7 @@ class TestDNDarray(TestCase):
             expected_rows += halo_size
         if data.halo_next is not None:
             expected_rows += halo_size
-        self.assertEqual(data.array_with_halos.shape, (expected_rows, cols))
+        self.assertEqual(array_with_halos.shape, (expected_rows, cols))
 
         # Large split=1 array
         rows, cols = 500, 1000 * size
@@ -340,6 +341,7 @@ class TestDNDarray(TestCase):
 
         halo_size = 100
         data.get_halo(halo_size)
+        array_with_halos = data.array_with_halos
 
         rank = data.comm.rank
         lshape_map = data.lshape_map
@@ -350,7 +352,7 @@ class TestDNDarray(TestCase):
                 data_np[:, displacements[rank] - halo_size : displacements[rank]],
                 device=data.device.torch_device,
             )
-            self.assertTrue(torch.all(data.halo_prev == expected_prev))
+            self.assertTrue(torch.all(array_with_halos[:, :halo_size] == expected_prev))
         else:
             self.assertIsNone(data.halo_prev)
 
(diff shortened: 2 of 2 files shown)


heat/core/dndarray.py
brownbaerchen:
```suggestion
                f"halo_size needs to be a non-negative Python integer, {type(halo_size)} given"
```

heat/core/dndarray.py
```diff
@@ -442,9 +442,9 @@ def get_halo(self, halo_size: int, prev: bool = True, next: bool = True) -> torc
             # exchange data with next populated process
             if prev:
```
brownbaerchen, line 442:
```suggestion
            if prev:
                a_prev = self.__prephalo(0, halo_size)
                res_prev = None
```

heat/core/dndarray.py
brownbaerchen, line 451:
```suggestion
            if next:
                a_next = self.__prephalo(-halo_size, None)
                res_next = None
```

brownbaerchen:
```suggestion
```
I suggest doing this a bit more lazily and keeping the definitions closer to where they are used.

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
             ht.equal(int16_tensor & int16_vector, ht.bitwise_and(int16_tensor, int16_vector))
         )
 
-    def test_gethalo(self):
-        data_np = np.array([[1, 2, 3, 4, 5, 6], [7, 8, 9, 10, 11, 12]])
+    def test_gethalo_split1_random(self):
+        """Test get_halo with randomly generated data split along axis 1."""
```
brownbaerchen:
```suggestion
        """Test get_halo with data split along axis 1."""
```
The data is just `np.arange`, no?

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
             ht.equal(int16_tensor & int16_vector, ht.bitwise_and(int16_tensor, int16_vector))
         )
 
-    def test_gethalo(self):
-        data_np = np.array([[1, 2, 3, 4, 5, 6], [7, 8, 9, 10, 11, 12]])
+    def test_gethalo_split1_random(self):
```
brownbaerchen:
```suggestion
    def test_gethalo_split1(self):
```

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
+        # Verify array_with_halos shape
+        expected_cols = data.lshape[1]
+        if data.halo_prev is not None:
+            expected_cols += halo_size
+        if data.halo_next is not None:
+            expected_cols += halo_size
+        self.assertEqual(data.array_with_halos.shape, (rows, expected_cols))
+
+    def test_gethalo_split0_random(self):
+        """Test get_halo with randomly generated data split along axis 0."""
```
brownbaerchen:
```suggestion
    def test_gethalo_split0(self):
        """Test get_halo with data split along axis 0."""
```
same as above

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
+
+        expected_rows = data.lshape[0]
+        if data.halo_prev is not None:
+            expected_rows += halo_size
+        if data.halo_next is not None:
+            expected_rows += halo_size
+        self.assertEqual(data.array_with_halos.shape, (expected_rows, cols))
+
+    def test_gethalo_float_random(self):
+        """Test get_halo with randomly generated float data."""
```
brownbaerchen:
```suggestion
    def test_gethalo_float(self):
        """Test get_halo with float data."""
```

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
+            self.assertIsNone(data.halo_next)
+
+        expected_rows = data.lshape[0]
+        if data.halo_prev is not None:
+            expected_rows += halo_size
+        if data.halo_next is not None:
+            expected_rows += halo_size
+        self.assertEqual(data.array_with_halos.shape, (expected_rows, cols))
+
+    def test_gethalo_float_random(self):
```
brownbaerchen: Maybe give datatype as an argument with defaupt to the previous test functions and call them from here with different datatype?

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
-                self.assertEqual(data.halo_next, None)
+        halo_size = 2
+        data.get_halo(halo_size)
 
-            data = ht.ones((10, 2), split=0)
+        rank = data.comm.rank
+        lshape_map = data.lshape_map
+        populated_ranks = torch.nonzero(lshape_map[:, 1]).squeeze().tolist()
+        if not isinstance(populated_ranks, list):
+            populated_ranks = [populated_ranks]
```
brownbaerchen: what's going on here? Since the data is `np.arange`, aren't all ranks populated?

heat/core/tests/test_dndarray.py
brownbaerchen: I see this is a different test than the others, but it's not obvious how it differs.

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
-                new_split_size = data.larray.shape[0] + 1
+        if rank in populated_ranks:
+            offsets = [0] + torch.cumsum(lshape_map[:, 1], dim=0).tolist()
+            if rank != first_rank:
+                expected_prev = torch.tensor(
+                    data_np[:, offsets[rank] - halo_size : offsets[rank]],
+                    device=data.device.torch_device,
+                )
+                self.assertTrue(
+                    np.isclose(((data.halo_prev - expected_prev) ** 2).mean().item(), 0.0)
```
brownbaerchen:
```suggestion
                    torch.allclose(data.halo_prev.larray - expected_prev)
```
why so complicated?

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
-                new_split_size = data.larray.shape[1] + 1
+                self.assertIsNone(data.halo_prev)
+
+            if rank != last_rank:
+                expected_next = torch.tensor(
+                    data_np[:, offsets[rank + 1] : offsets[rank + 1] + halo_size],
+                    device=data.device.torch_device,
+                )
+                self.assertTrue(
+                    np.isclose(((data.halo_next - expected_next) ** 2).mean().item(), 0.0)
```
brownbaerchen: again, this seems unnecessarily complicated.
```suggestion
                    torch.allclose(data.halo_next.larray, expected_next)

File: heat/core/tests/test_dndarray.py [hunk 10]
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
-                new_split_size = data.larray.shape[1] + 2
-            self.assertEqual(data_with_halos.shape, (5, new_split_size))
-            self.assertTrue(data.halo_prev is prev_halo or (data.halo_prev == prev_halo).all())
-            self.assertTrue(data.halo_next is next_halo or (data.halo_next == next_halo).all())
+                self.assertIsNone(data.halo_next)
+
+    def test_gethalo_no_data_on_process(self):
+        """Test get_halo when some processes have no local data."""
+        if ht.MPI_WORLD.size < 3:
+            return
```
brownbaerchen:
```suggestion
            self.skipTest("Need at least 3 processes")
```

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
+        """Test get_halo with randomly generated data split along axis 1."""
+        rows, cols = 4, 6 * max(ht.MPI_WORLD.size, 2)
+        data_np = np.arange(rows * cols).reshape((rows, cols))
         data = ht.array(data_np, split=1)
 
-        if data.comm.size == 2:
-            halo_next = torch.tensor(np.array([[4, 5], [10, 11]]), device=data.device.torch_device)
-            halo_prev = torch.tensor(np.array([[2, 3], [8, 9]]), device=data.device.torch_device)
+        if not data.is_distributed():
+            return
```
brownbaerchen:
```suggestion
            self.skipTest("Need data to be distributed")
```

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
+        self.assertEqual(data.array_with_halos.shape, (rows, expected_cols))
+
+    def test_gethalo_split0_random(self):
+        """Test get_halo with randomly generated data split along axis 0."""
+        rows, cols = 6 * max(ht.MPI_WORLD.size, 2), 4
+        data_np = np.arange(rows * cols).reshape((rows, cols))
+        data = ht.array(data_np, split=0)
+
+        if not data.is_distributed():
+            return
```
brownbaerchen:
```suggestion
            self.skipTest("Need data to be distributed")
```

heat/core/tests/test_dndarray.py
brownbaerchen: Since the data size is not relative to the number of tasks, maybe this test needs skipping if too many tasks are used as well?

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
+            self.assertEqual(data_with_halos.shape, (12, 2))
+        if data.comm.rank == 2:
+            self.assertIsNone(data.halo_prev)
+            self.assertIsNone(data.halo_next)
+            self.assertEqual(data_with_halos.shape, (12, 0))
+
+    def test_gethalo_imbalanced(self):
+        """Test get_halo with imbalanced data distribution across processes."""
+        if ht.MPI_WORLD.size < 3:
+            return
```
brownbaerchen:
```suggestion
            self.skipTest("Need at least 3 processes")
```

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
+        # exception for halo larger than local chunk size
+        if data.is_distributed():
+            with self.assertRaises(ValueError):
+                data.get_halo(data.lshape[1] + 1)
+
+    def test_gethalo_large_array(self):
+        """Test get_halo with very large arrays and large halo sizes."""
+        size = ht.MPI_WORLD.size
+        if size < 2:
+            return
```
brownbaerchen:
```suggestion
            self.skipTest("Need at least 2 processes")
```

heat/core/tests/test_dndarray.py
```diff
@@ -26,212 +26,351 @@ def test_and(self):
...
+            data.get_halo("wrong_type")
+        # exception on negative halo size
+        with self.assertRaises(ValueError):
+            data.get_halo(-99)
+        # exception for halo larger than local chunk size
+        if data.is_distributed():
+            with self.assertRaises(ValueError):
+                data.get_halo(data.lshape[1] + 1)
+
+    def test_gethalo_large_array(self):
```
brownbaerchen, line 253: Is this test needed? What is different when there is a lot of data compared to when there is little data?

brownbaerchen (changes requested):
Looks good overall. The waiting makes sense, the tests are comprehensive. (Even though I don't understand the unbalanced one) I just marked a few things to be refactored, but should be quick. Thanks a lot!

heat/core/tests/test_dndarray.py
JuanPedroGHM: The issue with the convolution only appeared on large halos, because the last elements on some of the written in time (running theory), because we were not waiting for some Isend messages to wait.

This test tries to recreate that error to catch it in the future.

heat/core/dndarray.py
JuanPedroGHM: ?

JuanPedroGHM:
> Looks good overall. The waiting makes sense, the tests are comprehensive. (Even though I don't understand the unbalanced one) I just marked a few things to be refactored, but should be quick. Thanks a lot!

Thanks for the review. Most of the changes have been accepted. Unbalanced one tests that the halos are still correct, even when the dndarray is not balanced, not equal splits (or close to equal) between the ranks

heat/core/tests/test_dndarray.py
brownbaerchen: Does the test fail on the current main? If so, this seems like a good test 👍

heat/core/dndarray.py
brownbaerchen: I don't know why the diff looks so weird and I know that you are not modifying this line in this PR.
Nevertheless, a halo size of 0 is allowed and is not a positive integer, right? This was more of a Boy Scout rule type suggestion..

brownbaerchen (dismissed):
I don't have a strong opinion about the remaining comments. If you decide to merge the ones with `a_next` etc., don't forget to delete it above. This comment is somehow disappeared.
Anyways, if you want to merge like this, go ahead.

heat/core/tests/test_dndarray.py
JuanPedroGHM: It should, that's how I chose the sizes.

brownbaerchen (dismissed):

JuanPedroGHM:
'get_halo(0)' is relevant for halo calculations with unbalanced arrays. Changed wording to have a consistent error message.

brownbaerchen (approved):

Issue #2146, opened 2026-02-23
Run CI on forks


brownbaerchen:
In order to make contributions easier, it would be fantastic if we had a clean CI that people can run on their forks. The current workflows are specific to the main repository and fail when run on a fork.
We are currently adding a workflow in #2134 that would run on push on any branch, enabling people to use CI for developments on their own forks, which is great.
So I am happy if we
 - [x] Merge #2134 to get a CI workflow that people can run on their forks
 - [ ] Add something like `if: ${{ !github.event.repository.fork }}` to workflows that rely on secrets / HPC resources and can't run on forks

This is a crucial step for the Heat hackathon that will be organized as part of HeatHub such that people use their own GitHub CI resources during development on their own forks and do not exhaust the resources of the main repository. Gotta love forks.

PR #2152, opened 2026-02-25
Fix CI on forks and run on draft PRs

brownbaerchen:
## Description
This PR disables workflows that trigger workflows on HPC resources if run on forks, as forks are not allowed and able to trigger them.
Instead, whenever people push branches to their forks, the new lean CI will run. It gives you quick feedback if something failed and runs on your own resources.

Side note: The fork workflow is a bit counter to the special branches that are created on issue assignment. When working with these branches, you push to the upstream repo rather than the fork, which clutters the actions tab and sends emails on failure to a bunch of people who don't care. Also, you need write permission to the main repo in order to use them, so outside contributors can't and we just create a bunch of branches that nobody will remember to delete. We may want to discuss getting rid of these branches later on.

Issue/s resolved: #2146 

## Changes proposed:

- Skip CI workflows triggering workflows outside of GitHub if run on fork
- Run CI on draft PRs

## Type of change

tending the garden

### DIFF
--- a/.github/workflows/CIBase.yml
+++ b/.github/workflows/CIBase.yml
@@ -21,7 +21,7 @@ jobs:
           fi
 
       - name: Harden Runner
-        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
+        uses: step-security/harden-runner@58077d3c7e43986b6b15fba718e8ea69e387dfcc # v2.15.1
         with:
           egress-policy: audit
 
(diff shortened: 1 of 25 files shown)


brownbaerchen:
@mtar, I merged main to trigger the CI in my fork. Indeed, it does what it's supposed to as you can see in the [actions log](https://github.com/brownbaerchen/heat/actions). In the meeting we saw failing actions from a previous state that didn't work. In the current log, the respective actions are not shown because they are exited early enough.

Maybe my rambling is confusing, but the point is that this PR is doing what it's supposed to. When I push to my fork, only the lean CI runs and I only get an email if this failed. The CI on the main repository is not affected as far as I can tell.

PR #2154, opened 2026-02-26
Getting started on interoperability tutorial

brownbaerchen:
## Description

This is a start of the interoperability tutorial. I don't know if I am using heat the proper way yet, so I thought it best to start with basic numpy and then continue with other stuff. This is also why this PR is not targeting main, but a different branch.

Maybe I have mentioned once or twice that I am a big fan of jupytext. I have written the tutorial as a python file and set up a GitHub workflow that translates any python file in heat/tutorials/scripts to a notebook and uploads the resulting notebook as artifacts. These will then have to be included in the website.

This has the following advantages:
 - Easy to use with git: Just comment the python file while reviewing the PR
 - Easy to test: The GitHub workflow runs the python files before converting them to notebooks which is an easy smoke test
 - Users can copy-paste directly from the plain python file to get started. No need to copy from multiple cells in the notebook

Please give feedback on:
 - Is this really how to interoperate with numpy?
 - Do you like the jupytext workflow? (Don't forget to download the artifact with the finished notebook)

Finally, don't forget to squash and merge in the end in order to keep commit history clean! I needed quite a few tries to get the workflow to do what I want..

## Changes proposed:

- Tutorial on interoperating with numpy
- GitHub workflow for use of jupytext in tutorials

### DIFF
--- a/.github/workflows/generate_tutorial_notebooks.yaml
+++ b/.github/workflows/generate_tutorial_notebooks.yaml
@@ -32,6 +32,9 @@ jobs:
           pip install '.[hdf5,netcdf,pandas,zarr]'
           pip install jupytext notebook jupyter_client
 
+      - name: Install jupyter kernel
+        run: python -m ipykernel install --name heat_kernel --user
+
       - name: Run and convert python files
         run: |
           set -e
--- a/.github/workflows/generate_tutorial_notebooks.yaml
+++ b/.github/workflows/generate_tutorial_notebooks.yaml
@@ -6,7 +6,7 @@ jobs:
   generate_notebooks:
     runs-on: ubuntu-latest
     timeout-minutes: 120
-    name: Setup
+    name: Generate Notebooks
     steps:
       - name: Harden Runner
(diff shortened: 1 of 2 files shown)


brownbaerchen:
Issues we have discussed in today's meeting:
 - [ ] Users need to be able to use ipyparallel kernel to run notebooks in parallel -> Check out how easy that is to do with a jupytext-generated notebook
 - [ ] Jupytext-generated notebooks have output from random MPI rank. Need to take care that relevant heat output, which is only printed on rank 0 by default shows up

I will post updates about these issues here.

ClaudiaComito:
Thanks for getting this started @brownbaerchen . I think the content will require more work from all of us, so if you don't mind I'm moving this PR to Milestone 1.9.  

I'll add some examples of numpy/pytorch interoperability later

brownbaerchen:
> Thanks for getting this started @brownbaerchen . I think the content will require more work from all of us, so if you don't mind I'm moving this PR to Milestone 1.9.
> 
> I'll add some examples of numpy/pytorch interoperability later

That's why this PR is not targeting main. Can you maybe provide some quick feedback on the little stuff that is in here so far? Just so I make sure that I actually understand the "proper" way.

PR #2172, opened 2026-03-04
Put tests to new directory

### LINKED ISSUE #1924: Separate tests from code
It would be nice to have our unit tests separated (in a separate folder) from the actual code.

### LINKED ISSUE #2094: [Bug]: Paths to datasets in tests assume specific working directory.
### What happened?

Paths in tests typically look like
```python
os.path.join(os.getcwd(), "heat/datasets/iris.h5")
```
which assumes that the test is run from the main heat directory. As mentioned in #2086, an easy remedy is
```python
os.path.join(os.path.dirname(ht.__file__), "datasets/iris.h5")
```
 which makes the path relative to the heat installation rather than the current working directory.
This should be adopted in all tests.

```shell

```

```shell

```

### Version

main (development branch)

### Python version

None

### PyTorch version

None

### MPI version

```shell

```

mtar:
## Due Diligence

- General:
- Implementation:

## Description

This PR introduces a movement of the tests files bundled to a new directory.
Issue/s resolved: #1924
Closes: #2094 

## Changes proposed:

- move test files to `tests/`
- update ci
- move `basic_test.py` from `heat/core/tests/test_suites` to `heat/testing`

## Type of change

Testing

no


mtar:
I had to modify the codebase ci: [status](https://codebase.helmholtz.cloud/helmholtz-analytics/ci/-/pipelines/696407)

.github/workflows/ci.yaml
```diff
@@ -55,5 +55,5 @@ jobs:
           pip install pytest
           pip install ${{ matrix.pytorch-version }} ${{ matrix.install-options }} --extra-index-url https://download.pytorch.org/whl/cpu
           # use pytest -vv -x for debugging
-          pytest heat/
-          mpirun -n 4 pytest -vv heat/
+          pytest
+          mpirun -n 4 pytest -vv
```
brownbaerchen, line 57:
```suggestion
          pytest tests
          mpirun -n 4 pytest -vv tests
```

.github/workflows/ci_full.yaml
```diff
@@ -81,5 +81,5 @@ jobs:
         run: |
           pip install pytest
           pip install ${{ matrix.pytorch-version }} ${{ matrix.install-options }} --extra-index-url https://download.pytorch.org/whl/cpu
-          mpirun -n 3 pytest heat/
-          mpirun -n 4 pytest heat/
+          mpirun -n 3 pytest
+          mpirun -n 4 pytest
```
brownbaerchen, line 82:
```suggestion
          mpirun -n 3 pytest tests
          mpirun -n 4 pytest tests
```

heat/classification/tests/__init__.py
brownbaerchen, line 1: will the tests directories be deleted in git like this? We should get rid of them.

brownbaerchen (changes requested):
Thanks a lot! I think we need to remove all the files and directories that are no longer needed from the repository. But we should merge this soon..

heat/classification/tests/__init__.py
mtar: I will remove them again. After changing branches they got back again.

mtar: On the branch itself I don't see any of the old test directories.

brownbaerchen:
Does it make sense to fix the paths like suggested in #2094?

So the git diff is wrong / confusing? The files are actually gone? If we're not sure, we can just merge and clean up later if needed.

ClaudiaComito:
> Does it make sense to fix the paths like suggested in #2094?
> 
> So the git diff is wrong / confusing? The files are actually gone? If we're not sure, we can just merge and clean up later if needed.

Not before the March 26th release IMO.

mtar:
> So the git diff is wrong / confusing? The files are actually gone? If we're not sure, we can just merge and clean up later if needed.

The diff in the files tab links to an older commit hash when you try to view it. Edit and delete leads to an 404 error.

> Does it make sense to fix the paths like suggested in #2094?

Good point, I did it in the previous PR, but I forgot to do it in this one :sweat_smile:

brownbaerchen:
> > Does it make sense to fix the paths like suggested in #2094?
> > So the git diff is wrong / confusing? The files are actually gone? If we're not sure, we can just merge and clean up later if needed.
> 
> Not before the March 26th release IMO.

The clean-up should be very quick to do since it's just deleting some empty directories and init files. As this is touching so many things that other PRs are concerned with, I am afraid of the many merge conflicts we would create by postponing a merge of this PR..

ClaudiaComito:
> > > Does it make sense to fix the paths like suggested in #2094?
> > > So the git diff is wrong / confusing? The files are actually gone? If we're not sure, we can just merge and clean up later if needed.
> > 
> > 
> > Not before the March 26th release IMO.
> 
> The clean-up should be very quick to do since it's just deleting some empty directories and init files. As this is touching so many things that other PRs are concerned with, I am afraid of the many merge conflicts we would create by postponing a merge of this PR..

Alright then  @mtar @brownbaerchen go for it! I've updated the branch so no Python 3.10 attempts

mtar:
There are still some specific issues with the ci action on github that needs do be solved:
- tests fail on data set reading if test directory is given. No directory works.
- npy io tests run forever. problem with with statements or temporary directories?

brownbaerchen:
> There are still some specific issues with the ci action on github that needs do be solved:
> 
> * tests fail on data set reading if test directory is given. No directory works.
> * npy io tests run forever. problem with with statements or temporary directories?

I have some issues with the io tests locally as well since a few days. I am getting:
```
FAILED heat/core/tests/test_io.py::TestIO::test_load_multiple_csv - FileExistsError: [Errno 17] File exists: '...
FAILED heat/core/tests/test_io.py::TestIO::test_load_npy_float - ValueError: all the input array dimensions except for the concatenation axis must match exactly...
FAILED heat/core/tests/test_io.py::TestIO::test_load_npy_int - ValueError: all the input array dimensions except for the concatenation axis must match exactly...
```
This leads to a deadlock when run in parallel. I don't know if this is related to the issues you are having.

.github/workflows/ci.yaml
```diff
@@ -53,5 +53,5 @@ jobs:
...
```
brownbaerchen, line 57:
```suggestion
          pytest heat/tests
          mpirun -n 4 pytest -vv heat/tests
```
do we maybe need to do this?

mtar: Doesn't work, I had to revert it back.

.github/workflows/ReceivePR.yml
```diff
@@ -33,8 +33,9 @@ jobs:
       - name: Test
         run: |
           pip install .[dev]
+          pip install pytest
           pre-commit run --all-files
-          python -m unittest
+          pytest
```
brownbaerchen, line 38: Why are the tests run here?

heat/testing/basic_test.py
```diff
@@ -10,15 +12,19 @@
 from heat.core import MPI, MPICommunication, dndarray, factories, types, Device
 from heat.core.random import seed
 
+
 # TODO adapt for GPU once this is working properly
 class TestCase(unittest.TestCase):
+    """Helper functions for unit tests"""
+
```
brownbaerchen, line 19:
```suggestion
```
I'd rather avoid comments that don't add information.

brownbaerchen (dismissed):

brownbaerchen (dismissed):

brownbaerchen (dismissed):

heat/testing/basic_test.py
mtar: Neither do I. Ruff was complaining about a "Missing docstring in public package". We can add D104 to the ignore list instead.

brownbaerchen: Yes, let's add that to the global ignore list! Ideally the code speaks for itself.

.github/workflows/ReceivePR.yml
mtar: This was some simple test as we had not checked the commit earlier. I think it should be addressed in #2152

brownbaerchen (approved):

PR #2174, opened 2026-03-04
Minimum Supported Dependencies for v1.8.0: drop Python 3.10

ClaudiaComito:
## Due Diligence

- General:
- Implementation:

## Description

Prepping for 1.8.0 release, adopting [Spec 0](https://scientific-python.org/specs/spec-0000/). The only significant change is that we're dropping Python 3.10 support.

Issue/s resolved: #2164 

## Changes proposed:

- updated dependencies in pyproject.toml
- dropped python 3.10 test runs from ci.yaml and ci_full.yaml
- updated Dependencies section in README
- added Spec 0 badge on readme
- updated Automated Release Schedule automation to reference Spec 0 instead of the outdated NEP 29

## Type of change

Documentation
## Memory requirements

NA
## Performance

NA
no

### DIFF
--- a/.github/workflows/release-schedule.yml
+++ b/.github/workflows/release-schedule.yml
@@ -82,8 +82,8 @@ jobs:
 
             const issueData = [
               {
-                title: `[Release Prep]: NEP 29 Compliance Check for v${version}`,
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -147,7 +147,7 @@ ignore_errors=false
 
 # Ruff
 [tool.ruff]
-target-version = "py310"
+target-version = "py311"
```
brownbaerchen, line 150:
```suggestion
target-version = "py314"
```
?

ClaudiaComito: > Suggested change
> target-version = "py311"
> target-version = "py314"
> 
> ?

Shouldn't the target version be aligned to the oldest supported version? We might get syntax fixes that aren't supported on Python 3.11.

Issue #2052, opened 2025-12-01
Device detection for factories.array()


ClaudiaComito:
**Related**
#2044  #2045 #2046 

**Feature functionality**
`ht.array(obj, device=None)` should default to the object's device if `obj` is a torch tensor.

Currently, it defaults to CPU.

Thanks @GioPede for bringing this up.

brownbaerchen:
Note that it is recommended to use the global control `heat.use_device('cuda')` rather than adding `device='cuda'` to all array instantiation commands which also takes care of these issues.

PR #2191, opened 2026-03-10
Remove CPU as default device in `ht.array` with non-DNDarray input

brownbaerchen:
Removes CPU as default device in `ht.array` with non-DNDarray input. In the relevant call to `torch.as_tensor`, it is left up to torch to decide the device. This will be the device the data that is converted to tensor is on, as torch attempts do construct the tensor without copy. Later, if a device is supplied, the data is moved.

I did want to refactor this function. And I still think it's needed, but I ran out of patience because there seem to be many special cases, making this refactor a pretty large task.

Issue/s resolved: #2052

### DIFF
--- a/.github/workflows/ReceivePR.yml
+++ b/.github/workflows/ReceivePR.yml
@@ -2,7 +2,7 @@ name: Receive PR
 
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


brownbaerchen:
This is not the solution to this issue..

PR #2193, opened 2026-03-11
Fix for `test_load_multiple_csv`

brownbaerchen:
The test uses `os.mkdir` on rank 0 to create a directory in persistent storage. This fails when the directory already exists. This issue is currently preventing me from running the tests locally in parallel since only rank 0 fails and the test results in a deadlock. 
This test runs only if the optional dependency `pandas` is installed which is why I didn't run into this issue sooner and why maybe you are not affected.

### DIFF
--- a/heat/testing/basic_test.py
+++ b/heat/testing/basic_test.py
@@ -444,5 +444,4 @@ class TestCase(unittest.TestCase):
             path = None
 
         path = ht.MPI_WORLD.bcast(path, root=0)
-        ht.comm.Barrier()
         return path, tmpdir  # need to return tmpdir here so that its not cleaned up at this point
--- a/tests/test_basic_test.py
+++ b/tests/test_basic_test.py
@@ -97,6 +97,7 @@ class TestBasicTest(TestCase):
         path, tmpdir = self.get_tmpdir()
         assert os.path.exists(path), f'No directory at {path}'
         assert os.path.isdir(path), f'No directory at {path}'
+        ht.comm.Barrier()
 
         # test cleanup
         del tmpdir


brownbaerchen:
Hm. The CI on Helmholtz cloud has not been triggered because I merged main locally and then pushed rather than synching here. That's not ideal.

tests/core/test_io.py
```diff
@@ -886,11 +886,11 @@ def test_load_multiple_csv(self):
 
         import pandas as pd
 
-        csv_path = os.path.join(os.getcwd(), "heat/datasets/csv_tests")
+        csv_path = ht.comm.bcast(tempfile.mkdtemp())
```
mtar: This will create many temporary directories on multiprocess, but only the broadcasted one will be deleted. Furthermore, the default temporary directory is often leads to `/tmp` if not set otherwise. This can cause issues on HPC if more than one node is used and no environment variable is set.

brownbaerchen: Alright, I am now creating the temporary directory only on rank 0 and within the current working directory of where the test is run from.

brownbaerchen:
I did some testing on Juwels booster. On main, I get:
```
$ pytest tests/core/test_io.py
...
FAILED tests/core/test_io.py::TestIO::test_load_multiple_csv - FileExistsError: [Errno 17] File exists: './heat/heat/datasets/csv_tests'
```

I then did
```
git checkout fix_io_tests
```
and appended `print(ht.MPI_WORLD.rank, csv_path, flush=True)` in line 895 after communicating the local path.

Then I ran
```
srun -A ... --nodes=2 -n 4 pytest tests/core/test_io.py
```
Which actually seems to deadlock since I had to cancel this after ~10mins without any meaningful output.

Running only the test modified by this PR, however using
```
srun -A ... -N 2 -n 4 pytest tests/core/test_io.py::TestIO::test_load_multiple_csv
```
yields
```
3 /p/project1/.../heat/tmpqg171j5l
1 /p/project1/.../heat/tmpqg171j5l
0 /p/project1/.../heat/tmpqg171j5l
2 /p/project1/.../heat/tmpqg171j5l
```
and passes, meaning that the temporary directories are indeed created in a directory shared among the nodes and all is fine.

So I see that there are issues, but as far as I can tell, this PR does not contribute to them. Note that on main this test deadlocks because it fails only on rank 0.
So my recommendation is to merge this now so we remove some of the deadlocks..

mtar:
I tried to reproduced it on juwels, but didn't get quite the same result.

> I did some testing on Juwels booster. On main, I get:
> 
> ```
> $ pytest tests/core/test_io.py
> ...
> FAILED tests/core/test_io.py::TestIO::test_load_multiple_csv - FileExistsError: [Errno 17] File exists: './heat/heat/datasets/csv_tests'
> ```

I couldn't reproduce it. It works when I run it in the heat directory and outside it fails with FileNotFoundError when it tries to create the directory as the parent directory doesn't exist. However, no FileExistError.


> ```
> srun -A ... --nodes=2 -n 4 pytest tests/core/test_io.py
> ```
Same result, it hangs on a test when outside heat directory. Inside the directory, it runs.

> 
> ```
> srun -A ... -N 2 -n 4 pytest tests/core/test_io.py::TestIO::test_load_multiple_csv
> ```
> 

Same experience. This runs through.

brownbaerchen:
At this point I simply uninstalled pandas locally so I can finally run the tests on my laptop again... From your tests, is there any reason not to merge this? You don't find any issues with this fix, right? Weird that you couldn't reproduce the error. Have you tried on your laptop?

tests/core/test_io.py
mtar: My guess is that this part hangs on multiprocess. It creates the directory the same way like the code you have changed.

brownbaerchen: Do you want me to switch to temporary paths in all of these tests? How do we continue with this PR?

mtar: Only in **test_load_multiple_csv_exception** . It is related to the test function you edited.

brownbaerchen:
I added a function `get_tmpdir` to the basic `TestCase` class to make this easier in the future. I also added a test for this that checks if the directory is created and cleaned up properly. 
Indeed, after switching to this temporary directory in the `test_load_mulitple_csv_exception` test there are no more deadlocks for me.

heat/testing/basic_test.py
```diff
@@ -421,3 +422,26 @@ def __create_random_np_array(
             )
         array = array.astype(dtype)
         return array
+
+    def get_tmpdir(self):
+        """
+        Create a temporary directory in the directory the current working directory.
```
mtar:
```suggestion
        Create a temporary directory in the current working directory.
```

tests/core/test_io.py
mtar:
```suggestion
path, _ = self.get_tmpdir()
  with self.assertRaises(TypeError):
      ht.load_npy_from_path(path=1, split=0)
  with self.assertRaises(TypeError):
      ht.load_npy_from_path(path, split="ABC")
  with self.assertRaises(ValueError):
      ht.load_npy_from_path(path="heat", dtype=ht.int64, split=0)
  if ht.MPI_WORLD.size > 1:
      if ht.MPI_WORLD.rank == 0:
          x = np.random.rand(2, random.randint(1, 10), 11)
          np.save(os.path.join(os.getcwd(), path, "float_data"), x)
      ht.MPI_WORLD.Barrier()
      with self.assertRaises(RuntimeError):
          ht.load_npy_from_path(path, dtype=ht.int64, split=0)
      ht.MPI_WORLD.Barrier()
      if ht.MPI_WORLD.rank == 0:
          os.remove(os.path.join(os.getcwd(), path, "float_data.npy"))
```
This part was not working for me.

brownbaerchen: I did not apply your suggestion directly, because I removed the deletion of the data in the end. This happens automatically when the temporary directory is garbage collected at the end of the test.

brownbaerchen:
I had to add change a few minor things to fix the tests on all versions, but this is working and ready, I think.

Issue #2171, opened 2026-03-04
[Bug] For some matrices, the subspace iterations in `eigh` fail due to the matrix shape


GioPede:
During the call to the `_subspaceiteration` from the `eigh` function a variable `idx` is computed based on the properties of the matrix. When the input matrix is perfectly uniform this index leads to the matrix `X` being a square matrix. The following line thus produces a matrix with one extent being 0, as `X.shape[0] - X.shape[1]` for a square matrix is 0. The function `randn` then throws an error.

https://github.com/helmholtz-analytics/heat/blob/9f57145d8854aa24bb898e0e4b9aee7756990243/heat/core/linalg/eigh.py#L90

In my own branch I have fixed this issue by simply using `max(1, X.shape[0] - X.shape[1])` as the second parameter in `randn`, thus ensuring that the matrix can execute properly. The results are correct.

brownbaerchen:
Thanks for reporting all these bugs! In cases like this where you have already fixed it on your fork, we welcome a PR of your fix. It doesn't need to be perfect and you don't have to take of all the testing etc. If necessary, we can merge your PR to a development branch and take it from there.

ClaudiaComito:
@GioPede can you post a code snippet to reproduce the error and test the fix? Thanks!

brownbaerchen:
For the record: The reproducing example in #2194 can be reduced to the following:
```python
import heat as ht
mat = ht.ones((4, 4), split=0, dtype=ht.float64)
S, V = ht.linalg.eigh(mat)  # This fails when run in parallel
```
Note that this passes with float32. I am not an expert on the algorithm, but my feeling is that maybe we should skip the padding with random values rather than padding with 1 row in this case.

PR #2194, opened 2026-03-11
Fix bug for split SVD/eigh

GioPede:
## Description
This PR fixes a small bug for some split matrices that causes SVD and eigh to fail.

Issue/s resolved: #2171 

## Changes proposed:

- ensure that the second axis size is at least 1 in the case where ` X.shape[0] == X.shape[1]` (this function is only called for split matrices)
## Type of change

## Memory requirements
- Only happens on split matrices

Below is a minimal reproducer:
```python
import heat as ht
import numpy as np

# Constants
nx, ny = 4, 4
Nt = 32
Lambda = 2*nx*ny
beta=8
U=2
mu=0
delta = beta / Nt

# Heat settings
default_device = "cpu"
default_type = ht.float64

kappa=ht.diag(ht.zeros(Lambda, device=default_device, dtype=default_type))
for ix in range(nx):  # only loop over unit cells
    for iy in range(ny):
        ixy = 2 * (ix*ny + iy)
        kappa[ixy, ixy+1] += 1   # unit cell nearest neighbor
        kappa[ixy+1, ixy] += 1
        
        ixy1 = 2 * (((ix-1) % nx) * ny+iy)
        kappa[ixy, ixy1+1] += 1   # second nearest neighbor
        kappa[ixy1+1, ixy] += 1

        ixy2 = 2 * (ix*ny + (iy-1) % ny)
        kappa[ixy, ixy2+1] += 1   # third nearest neighbor
        kappa[ixy2+1, ixy] += 1
print(kappa)

mutilde = ht.diag(ht.ones(Lambda, device=default_device, dtype=default_type) * delta * mu)
M = delta * kappa - mutilde 
print(M)


# Self made exponential function
eig, eigv = ht.linalg.eigh(M)
inv_eigv = ht.linalg.inv(eigv)
expM = eigv @ ht.diag(np.exp(eig)) @ inv_eigv
print(expM)

print(ht.expm(M)) ## Note that expm does not return the same split as the original matrix

s, v, d = ht.linalg.svd(expM) # This fails on a split matrix
```

### DIFF
--- a/heat/core/linalg/eigh.py
+++ b/heat/core/linalg/eigh.py
@@ -87,7 +87,7 @@ def _subspaceiteration(
                 X,
                 randn(
                     X.shape[0],
-                    X.shape[0] - X.shape[1],
+                    max(1, X.shape[0] - X.shape[1]),
                     dtype=X.dtype,
                     device=X.device,
                     comm=X.comm,


ClaudiaComito:
Thanks a lot @GioPede, we fixed it via #2195, pls let us know if you still have problems. Closing this PR for now.

PR #2195, opened 2026-03-11
Fix edge case in distributed eigenvalue computation

### LINKED ISSUE #2171: [Bug] For some matrices, the subspace iterations in `eigh` fail due to the matrix shape

brownbaerchen:
## Description
See #2171 for details. Note that I am not an expert on the algorithm, but all the tests pass, so...

Issue/s resolved: #2171 
Supersedes: #2194

## Changes proposed:

- Fix bug by forgoing padding with random values if more padding is not possible
- Add test for the relevant edge case

## Type of change

Bug fix


ClaudiaComito (changes requested):
Thanks @brownbaerchen , yes I prefer this fix too. 

The 6-process run is failing (test-cuda but it's a CPU run), as soon as that's fixed, this can be merged.

brownbaerchen:
> Thanks @brownbaerchen , yes I prefer this fix too.
> 
> The 6-process run is failing (test-cuda but it's a CPU run), as soon as that's fixed, this can be merged.

The test was failing because of a different bug in matrix multiplication. I simply restrict the test to running only on two tasks because fixing matrix multiplication is out of scope of this PR. It is urgently needed, though!

tests/core/linalg/test_eigh.py
```diff
@@ -24,15 +24,20 @@ def test_eigh(self):
...
             for dtype in dtypes:
                 with self.subTest(split=split, dtype=dtype):
                     X = ht.random.randn(100, 100, split=split, dtype=dtype)
                     X = X + X.T.resplit_(X.split)
                     Lambda, H = ht.linalg.eigh(X)
                     self._check_eigh_result(X, Lambda, H)
-                    i += 1
+
+        # test edge case discussed in #2171
+        if ht.comm.size == 2:  # with some communicator sizes this fails because of a different bug
```
ClaudiaComito: can you specify the Issue? Otherwise this is good

brownbaerchen: The error is
```
>           if b.lshape[-1] % nB != 0 or (kB == 1 and b.lshape[-1] != 1):                                                                                                                                                                                                                                                                                                 
               ^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                          
E           ZeroDivisionError: integer modulo by zero 
```
raised [in this line](https://github.com/helmholtz-analytics/heat/blob/cc983dbe19a354349630fb6fa6320568a4db4c63/heat/core/linalg/basics.py#L846) in matrix multiplication.

ClaudiaComito:
```suggestion
        if ht.comm.size == 2:  # with some communicator sizes this fails, see Issue #2093
```

I meant something like this

ClaudiaComito (approved):
Thanks @brownbaerchen

mrfh92:
@ClaudiaComito @brownbaerchen sry for being that late on this issue. 

In my understanding, your bug fix now checks whether padding is necessary (because $X$ is actually non-square with more rows than columns) and then does this padding, whereas if $X$ is square already (and thus no padding necessary) no padding is performed. The original code likely attempted to do padding with an empty array in the latter case. 

So from my point of view this fix should be fine :+1:

brownbaerchen:
> @ClaudiaComito @brownbaerchen sry for being that late on this issue.
> 
> In my understanding, your bug fix now checks whether padding is necessary (because X is actually non-square with more rows than columns) and then does this padding, whereas if X is square already (and thus no padding necessary) no padding is performed. The original code likely attempted to do padding with an empty array in the latter case.
> 
> So from my point of view this fix should be fine 👍

Yes that is exactly right. Thanks for confirming that this makes sense from the maths side!

PR #2196, opened 2026-03-11
Skip tests without optional dependencies in lean CI

brownbaerchen:
## Description
The CI can be more lean! Afaik, when installing optional dependencies, we only run more tests than without. So there is no need to also run without optional dependencies in the lean CI in my opinion. This halves the number of jobs spawned by the lean CI.
On the other hand, I do think the lean CI should run the linter.

ReceivePR should not run any tests or linters in my opinion. The workflows should be as simple as possible.

## Changes proposed:

- Run lean CI only with all optional dependencies
- Run linter in lean CI
- Small refactor of lean CI
- Do not run tests in ReceivePR
- Do not run linter in ReceivePR

### DIFF
--- a/.github/workflows/ReceivePR.yml
+++ b/.github/workflows/ReceivePR.yml
@@ -30,13 +30,6 @@ jobs:
         with:
           mpi: openmpi
 
-      - name: Test
-        run: |
-          pip install .[dev]
-          pip install pytest
-          pre-commit run --all-files
-          pytest
-
       - name: Save PR number
         env:
           PR_NUMBER: ${{ github.event.number }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,8 +7,7 @@ on:
       - dependabot/**
 
 jobs:
-  approved:
-    if: github.event.pull_request.draft == false
+  tests:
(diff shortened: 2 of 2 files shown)


.github/workflows/ReceivePR.yml
```diff
@@ -30,13 +30,6 @@ jobs:
         with:
           mpi: openmpi
```
mtar, line 32: These steps are not necessary any longer without the test.

.github/workflows/ReceivePR.yml
mtar, line 1: The script should run after the lean CI.

.github/workflows/ci.yaml
```diff
@@ -18,7 +17,7 @@ jobs:
           - '3.11' # Oldest supported
           - '3.14' # Latest stable
         mpi: [ 'openmpi' ]
-        install-options: [ '.', '.[hdf5,netcdf,pandas,zarr]' ]
+        install-options: [ '.[hdf5,netcdf,pandas,zarr,dev]' ]
```
mtar, line 20: While we don't need to test the all combinations without the optional dependencies, I still recommend a single one. In the past, we had to fix some bugs because a non-dependency was overlooked at some point.

Issue #2190, opened 2026-03-10
Implement `silhouette_score`


ClaudiaComito:
**Related**
#2189 

**Feature functionality**

The Silhouette Score validation metrics to allow users to evaluate clustering quality (e.g., for `KMeans` or `SpectralClustering`) on distributed datasets. The implementation should follow the `scikit-learn` API [sklearn.metrics.silhouette_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.silhouette_score.html) but must be engineered for heat.DNDarray input (large distributed memory arrays) and MPI communication.

**Technicalities**
- target module: `heat.cluster.metrics` for now (potentially `heat.metrics` eventually)
- following the sklearn implementation, the core functionality will reside in `silhouette_score` and `silhouette_samples`
- within `silhouette_samples`, utilize [`heat.spatial.distance.py`](https://github.com/helmholtz-analytics/heat/blob/main/heat/spatial/distance.py) for optimized distance calculation on distributed arrays (`cdist` or `_dist`).
- results must match `sklearn.metrics.silhouette_score` exactly when executed on a single node (non-distributed)
- benchmark function  should be added to [heat/benchmarks/cb/cluster.py](https://github.com/helmholtz-analytics/heat/blob/main/benchmarks/cb/cluster.py)

**Additional context**
See  https://github.com/Astroua/SCIMES/blob/master/scimes/scimes.py for an application.

mtar:
Is the issue completed?

brownbaerchen:
Yes, this was completed in #2199.

PR #2199, opened 2026-03-13
Implement silhouette_score from scikit-learn

Shabasovich:
## Description

- tried to implement silhouette_samples from scikit-learn

## Type of change

- new function

#2190 

## Additional
Explanation of calculation
### Main goal
- Calculation Silhouette score using Formula $$s(i) = \frac{b(i)-a(i)}{max\{ a(i),b(i) \}}$$
##### Example
When I refer to example I am using following values:
- `X=[[0, 0], [10, 10], [20, 20], [1, 1]]`
- `labels=[10, 30, 20, 10]` 
	- `unique_labels = [10,20,30]`
	- `labels_encoded = [0, 2, 1, 0]` 

### Explanation of a(i) calculation
Formula: $$a(i) = \frac{1}{|C_{i}|-1} \cdot \sum_{j \in C_{i}, j \neq i} d(i,j)$$

#### Mask
- Goal: isolate distances within the same cluster with boolean matrix
- Uses `==` operator, that compares vector with its transpose. If labels are the same, has 1, meaning $M_{i,j}=1$ if $v_{i}=v_{j}$ 
##### Example
$$\begin{bmatrix} 0 \\ 
2 \\
 1 \\ 
0  \end{bmatrix}
== \begin{bmatrix}0  & 2 & 1 & 0 \end{bmatrix} \implies \begin{bmatrix} (0==0)=1 & (0==2)=0 & (0==1)=0 & (0==0)=1 \\
0 & 1 & 0 & 0\\
0 & 0 & 1 & 0\\
1 & 0 &0&1\end{bmatrix}$$

#### Distance Filtering
Multiplikation with Distance Mask $D$ gives a Matrix where in row $i$ every column $j$ that is not in i's cluster is set to 0.
##### Example
$$
D = \begin{bmatrix}
d_{0,0} & d_{0,1} & d_{0,2} & d_{0,3} \\
 d_{1,0} & d_{1,1} & d_{1,2} & d_{1,3} \\
 d_{2,0} & d_{2,1} & d_{2,2} & d_{2,3} \\
 d_{3,0} & d_{3,1} & d_{3,2} & d_{3,3} \\
\end{bmatrix} \times \begin{bmatrix}
1 & 0 & 0 & 1 \\
 0 & 1 & 0 & 0 \\
 0 & 0 & 1 & 0 \\
1 & 0 & 0 & 1
\end{bmatrix}  = \begin{bmatrix}
d_{0,0}  & 0  & 0 & d_{0,3} \\
0  & d_{1,1}  &  0  & 0 \\
0 & 0 & d_{2,2} & 0  \\
d_{3,0} & 0 & 0  & d_{3,3}
\end{bmatrix}
$$ 
#### Sum
- Since $d(i,i)=0$, the self-distance naturally drops out
- So if we are summing among the rows, we get exactly $\sum_{j \in C_{i}} d(i,j)$ with $C_{i}$ is cluster of $i$, $d(i,j)$ is distance between samples $i$ and $j$ 
- We can put it in array of sums like `sums`: `[sum_dist_0, sum_dist_1, sum_dist_2, sum_dist_3]` 
##### Example
- We get  $d_{0,3}$ from 1 row

#### Denominator
- Now we need a denominator for every point $i$ , so we need to determine the value of $|C_{i}|-1$ for each sample
- `labels_freqs` has all the sizes, you just need to map them to each sample with `labels_encoded` and subtract 1

### b(i) calculation
Formula:
$$b(i)= \min_{j \neq i} \frac{1}{|C_{j}|} \sum_{l \in C_{j}} d(i,l) $$
#### Mask
- Same as in calculation of $a(i)$ but we use vector of unique labels
- gives a Matrix where `mask[i, k] = 1` if sample $i$ belongs to cluster $k$ 
##### Example

$$
\begin{bmatrix}
0 \\
2 \\
1 \\
0
\end{bmatrix} == \begin{bmatrix}
0 & 1 & 2\end{bmatrix} \implies mask= \begin{bmatrix}
1 & 0 & 0 \\
0 & 0 & 1 \\
0 & 1 & 0 \\
1 & 0 & 0
\end{bmatrix}$$ 
#### Distance Filtering
- Multiplikation with Distance Mask $D$ gives a Matrix where entry $(i,k)$ is the sum of distances from sample i to all points in cluster k. --> äquivalent to $\sum_{l\in C_{i}} d(i,l)$ 
- we want neighbor cluster, so we set the distance to points in own cluster to infinity to force `min` to choose neighbors
##### Example

$$
D = \begin{bmatrix}
d_{0,0} & d_{0,1} & d_{0,2} & d_{0,3} \\ 
d_{1,0} & d_{1,1} & d_{1,2} & d_{1,3} \\ 
d_{2,0} & d_{2,1} & d_{2,2} & d_{2,3} \\ 
d_{3,0} & d_{3,1} & d_{3,2} & d_{3,3} \\
\end{bmatrix} 
\times \begin{bmatrix}
1 & 0 & 0  \\ 
0  & 0 & 1 \\ 
0 & 1 & 0\\
1 & 0 & 0 
\end{bmatrix}  = \begin{bmatrix}
d_{0,0}+d_{0,3}  & d_{0,2}  & d_{0,1} \\
d_{1,0}+d_{1,3}  & d_{1,2}   & d_{1,1} \\
d_{2,0}+d_{2,3} &  d_{2,2} & d_{2,1}  \\
d_{3,0}+d_{d_{3,3}} & d_{3,2} & d_{3,1}  
\end{bmatrix}
$$
After setting inf:
$$
\begin{bmatrix}
\infty  & d_{0,2}  & d_{0,1} \\
d_{1,0}+d_{1,3}  & d_{1,2}   & \infty \\
d_{2,0}+d_{2,3} &  \infty & d_{2,1}  \\
\infty & d_{3,2} & d_{3,1}  
\end{bmatrix}
$$


tests/cluster/test_metrics.py
brownbaerchen, line 1: We have moved the tests to a different directory in #2172, please move this file there.

heat/cluster/tests/test_metrics.py
```diff
@@ -0,0 +1,68 @@
...
+    # Comparison
+
+    ht_results_np = ht_results.numpy()
+
+    if np.allclose(sk_results, ht_results_np, atol=1e-5):
+        print("✅ Test Passed: HeAT matches Scikit-Learn results.")
+    else:
+        max_diff = np.max(np.abs(sk_results - ht_results_np))
+        print(f"❌ Test Failed: Results differ. Max diff: {max_diff}")
+        #print(f"sk_results are {np.abs(sk_results)}; ht_results are {np.abs(ht_results_np)}")
```
brownbaerchen:
```suggestion
    ht_results_np = ht_results.resplit(None).numpy()
    
    assert np.allclose(sk_results, ht_results, atol=1e-5), f'Max diff between Heat and scipy: np.max(np.abs(sk_results - ht_results_np))'
```

heat/cluster/tests/test_metrics.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # Single sample in a cluster
+    labels_edge = np.array([0, 0, 0, 1])
+    X_edge = np.random.rand(4, n_features)
+
+    res_edge = silhouette_samples(ht.array(X_edge), ht.array(labels_edge))
+    if res_edge[3] == 0:
+        print("✅ Edge Case Passed: Single-sample cluster correctly assigned 0.0")
```
brownbaerchen:
```suggestion
    assert res_edge[3] == 0:
```

heat/cluster/tests/test_metrics.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Expected value for i=0 (Cluster 0)
+        # a = dist((0,0), (1,1)) = 1.414
+        # b = dist((0,0), (10,10)) = 14.14
+        # sil = (14.14 - 1.414) / 14.14 = 0.9
+
+    if res_np[0] > 0.8:
+        print(f"✅ Success! Point 0 is {res_np[0]:.4f}")
+    else:
+        print(f"❌ Failure! Point 0 is {res_np[0]:.4f}")
```
brownbaerchen:
```suggestion
assert res_np[0] > 0.8, f"Point 0 is {res_np[0]:.4f}"
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,136 @@
...
+        Number of samples.
+    """
+    if not 1 < n_labels < n_samples:
+        raise ValueError(
+            "Number of labels is %d. Valid values are 2 to n_samples - 1 (inclusive)" % n_labels
+        )
+
+
+def silhouette_samples(X, labels, *, metric="euclidean", **kwds):
+    X_distributed = ht.array(X, split=0)
```
brownbaerchen: You want the input to be a Heat array that is split according to the user's need. Have a look at `heat.sanitation.sanitize_in` to make sure the input is a valid heat DNDarray and then don't split.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,136 @@
...
+    labels_distributed = ht.array(labels, split=0)
```
brownbaerchen: Same comment as for `X`

heat/cluster/metrics.py
```diff
@@ -0,0 +1,136 @@
...
+        # mb write a function to fill diag with 0, like np.fill_diagonal(X, 0)
+        diag_elements = ht.diag(X_distributed)
+
+        if X_distributed.dtype.kind == "f":
+            atol = ht.finfo(X_distributed.dtype).eps * 100  # tolerance based on machine accuracy
+
+            if ht.any(ht.abs(diag_elements) > atol):
+                raise error_msg
+        elif ht.any(diag_elements != 0):  # integral dtype
+            raise error_msg
```
brownbaerchen: I don't think single precision deserves special treatment here. You can check against the machine eps for any datatype.

brownbaerchen (commented):
This is already looking quite good. I had a brief look and highlighted some general things. Mainly using `assert` statements in the tests.

Since you mentioned you had trouble debugging, I want to advertise the [debugger](https://docs.python.org/3/library/pdb.html) once more. Just write `breakpoint()` in the code whenever you want to stop and explore. There can be trouble with parallel debugging. So when running in parallel, I only call a breakpoint on rank 0:
```
if ht.comm.rank == 0:
    breakpoint()
```
If you're already familiar with this stuff, just ignore this. But I know many python programmers, myself included, who were introduced to this way too late, so it doesn't hurt to mention it :D

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+
+    check_consistent_length(X, y)
+
+    return X, y
+
+
+def check_array(X, accept_sparse=False, input_name="X"):
+    if not isinstance(X, ht.DNDarray):
+        # Convert to heat array
+        X = ht.array(X, split=0)
```
ClaudiaComito: In general I wouldn't impose `split=0`  by default, as it comes with all kind of communication overhead downstream (i.e. every operation on the array will be performed in distributed mode). 

Unless you've assessed that it has advantages from an algorithmic point of view.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+
+    if not accept_sparse and X.is_sparse:
+        raise TypeError(f"{input_name} is sparse, but sparse input is not accepted.")
+
+    return X
+
+
+def _check_y(y):
+    if not isinstance(y, ht.DNDarray):
+        y = ht.array(y, split=0)
```
ClaudiaComito: same as above

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+        return ht.random
+
+    raise ValueError(f"Type {type(seed)} cannot be used to seed ht.random")
+
+
+# Functions
+def silhouette_samples(X, labels, *, metric="euclidean", **kwds):
+    X, labels = check_X_y(
+        X, labels, accept_sparse=["csr"]
+    )  # think about accept_sparse, i have no idea what it is and what csr means
```
ClaudiaComito: hints:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html

We also support sparse csr matrices with very few arithmetic operations, see ht.sparse .

But let's leave the sparse functionality for later, it's lower priority for now

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+    sil_samples = ht.where(labels_freqs[labels_encoded] > 1, sil_samples, 0.0)
+    sil_samples = ht.nan_to_num(sil_samples)
+
+    return sil_samples
+
+
+def silhouette_score(X, labels, *, metric="euclidean", sample_size=None, random_state=None, **kwds):
+    if sample_size is not None:
+        X, labels = check_X_y(X, labels, accept_sparse=["csc", "csr"])  # same as silhouette_samples
+        random_state = check_random_state(random_state)
```
ClaudiaComito: I'm not sure we need this whole abstraction, you could replace this simply with `ht.random.seed(random_state)`

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+        if ht.any(ht.abs(diag_elements) > atol):
+            raise error_msg
+        elif ht.any(diag_elements != 0):  # integral dtype
+            raise error_msg
+
+    unique_labels, labels_encoded = ht.unique(
+        labels, return_inverse=True
+    )  # f.e. labels = [10, 30, 20, 10], then unique_labels = [10,20,30] and labels_encoded = [0, 2, 1, 0]
+    unique_labels.resplit_(None)
+    labels_encoded.resplit(None)
```
ClaudiaComito: `DNDarray.resplit(None)` is probably the most dangerous method in the codebase. It gathers all the distributed slices of the global DNDarray onto each process, i.e. each process will try to hold the full DNDarray in memory. 

In this case `unique_labels` might be relatively small (we don't know!), but `labels_encoded` has the same size as the original input data, so `labels_encoded.resplit(None)` will most certainly fail, as in, the MPI processes will run out of memory.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+    labels_freqs = ht.bincount(labels_encoded)
+    labels_freqs.resplit_(None)
```
ClaudiaComito: vade retro resplit(None)!

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+    )  # reshape((-1,1)) transposes labels_encoded
+    a_mask = a_mask.astype(ht.float32)
+
+    a_clust_dists = ht.sum(ht.mul(D, a_mask), axis=1)
+    denominator_a = labels_freqs.larray[labels_encoded.larray] - 1
+    denominator_a = ht.where(
+        ht.array(denominator_a, split=0).astype(ht.float32) > 0,
+        ht.array(denominator_a, split=0),
+        1.0,
+    )
```
ClaudiaComito: with `.larray` you're operating directly on the underlying torch tensor (a slice of the global array). In this snippet of code:

- on each process, we define `denominator_a` by indexing the **local slice** of `label_freqs` by the **local slice** of `labels_encoded`. Is that what we want, no info required from the other processes? 
- in the following line, I think we want to read the local `denominator_a` tensors into a global distributed DNDarray. The argument for that operation is `is_split =  0`. Check out the [ht.array()](https://heat.readthedocs.io/en/stable/autoapi/heat/core/factories/index.html#heat.core.factories.array) docs.   
- because `ht.array(is_split=0)` must exchange information on slice size among processes (communication overhead),  better define it only once.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+
+    a = ht.div(a_clust_dists, ht.array(denominator_a, split=0))
```
ClaudiaComito: same here, should be ht.array(denominator_a, is_split=0) and should be defined once before all these calls

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # b(i) calculation
+    b_mask = ht.reshape(labels, (-1, 1)) == ht.reshape(unique_labels, (1, -1))
+    full_b_mask = (labels_encoded.reshape((-1, 1)) == unique_labels.reshape((1, -1))).astype(
+        ht.float32
+    )
+
+    b_clust_dists = ht.matmul(D, full_b_mask)
+    labels_freqs.resplit_(None)
```
ClaudiaComito: vade retro resplit(None)

heat/cluster/metrics.py
```diff
@@ -0,0 +1,215 @@
...
+    denominator_b = labels_freqs.astype(ht.float32)
+    denominator_b.resplit_(None)
```
ClaudiaComito: bad bad resplit_(None)

ClaudiaComito (dismissed):
Thanks a lot @Shabasovich ! I have a few high-level comments :

- the fundamental requirement for any Heat functionality, is that it must support massive memory-distributed arrays;
- this PR is calling [`DNDarray.resplit_(None)`](https://heat.readthedocs.io/en/stable/autoapi/heat/core/dndarray/index.html#heat.core.dndarray.DNDarray.resplit_) often in the code, which means that the DNDarray gets gathered onto each MPI process. After a `resplit_(None)`, any operation on that array will be single-process. 
- most importantly, for our use cases there won't be an "after resplit_(None)", because the process runs out of memory.  
 
At the current stage, the code might be passing the tests, but doesn't exploit memory distribution. The good thing is that all of the Heat  operations you're using do support memory distribution, so, next steps:

- find the spots in the code where arrays are being "gathered" (all the `resplit_(None)` calls)
- ask yourself: what's size do we expect this array to be if the original input data have, say, ~1e6 or 1e7 rows? does it make sense to try to copy it entirely on each process?
- if not: don't resplit. Just apply the follow up Heat operation to the distributed array.

By the way, check out our new [LLM and AI usage](https://github.com/helmholtz-analytics/heat/blob/main/CONTRIBUTING.md#llm-and-ai-usage) guidelines. It's really common for the LLMs to not grasp the memory-distributed requirement for Heat operations. 

**My recommendation is to not use the AI to write code for the next iteration**, but to work on adapting the existing code. Since you've just taken the MPI course, you might want to refresh the [tutorials](https://github.com/helmholtz-analytics/heat/blob/main/doc/source/tutorials/notebooks/README.md) again, especially check out [Internals](https://github.com/helmholtz-analytics/heat/blob/main/doc/source/tutorials/notebooks/Internals.ipynb)

Thanks again for all the work!

heat/cluster/metrics.py
Shabasovich: i guess i will do a length check before converting it to heat array, because communication overhead is then reasonable

brownbaerchen: Personally, I would either not distribute in this case or require the input to be a heat array. 
I am not sure what the workflow would be where you pass a torch tensor or whatever and use multiple tasks. But in this case, I would not be surprised if different tasks do different things. Since heat uses the world communicator within the function, this can lead to deadlocks.

heat/cluster/metrics.py
brownbaerchen: Sparse matrices are matrices with few entries relative to the overall size. This is the case for the masking matrix, where some clusters only contain few entries. It is notably not the case for the distance matrix because every point has a distance wrt to any other point. So it doesn't make sense to me to accept sparse matrices as input and I agree that we should just remove it from this PR. 
Having a sparse masking matrix would be nice, but I don't know if there is much to be gained when multiplying the dense distance matrix with a sparse masking matrix. Let's worry about this once it becomes a problem.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    >>> import heat as ht
+    >>> X = ht.array([[1, 2], [1, 1], [4, 4], [4, 5]], split=0)
+    >>> labels = ht.array([0, 0, 1, 1], split=0)
+    >>> ht.cluster.silhouette_score(X, labels)
+    0.76439
+    """
+    if sample_size is not None:
+        X, labels = check_X_y(X, labels, accept_sparse=["csc", "csr"])  # same as silhouette_samples
+        if random_state is not None:
+            ht.random.seed(random_state)
```
brownbaerchen, line 351: I think we need an else here that defines the random state that is subsequently used.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+        indices = random_state.permutation(X.shape[0])[
+            :sample_size
+        ]  # selecs a subset of random samples, but all ranks need same indices
+
+        if metric == "precomputed":  # precomputed means here distance matrix for some reason?
```
brownbaerchen:
```suggestion
        if metric == "precomputed":  # input is distance matrix
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    >>> ht.cluster.silhouette_score(X, labels)
+    0.76439
+    """
+    if sample_size is not None:
+        X, labels = check_X_y(X, labels, accept_sparse=["csc", "csr"])  # same as silhouette_samples
+        if random_state is not None:
+            ht.random.seed(random_state)
+        indices = random_state.permutation(X.shape[0])[
+            :sample_size
+        ]  # selecs a subset of random samples, but all ranks need same indices
```
brownbaerchen:
```suggestion
        ]  # selects a subset of random samples, but all ranks need same indices
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+                "Local shapes of X and y do not match. Ensure they are partitioned identically."
+            )
+
+
+# Functions
+def silhouette_samples(X, labels, *, metric="euclidean", **kwds):
+    """
+    Compute the Silhouette Coefficient for each sample.
+
+    The Silhouette Coefficient is a measure of how similar an object is to its own cluster
```
brownbaerchen:
```suggestion
    The Silhouette Coefficient is a measure of how close an object is to its own cluster
```
?

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+
+    * The score is 0 for clusters with only a single sample.
+    * The calculation involves computing the mean intra-cluster distance (a) and
+      the mean nearest-cluster distance (b) for each sample.
+
+    Parameters
+    ----------
+    X : DNDarray
+        An array of pairwise distances between samples, or a feature array.
+        If `metric='precomputed'`, X is assumed to be a distance matrix.
```
brownbaerchen:
```suggestion
        If `metric='precomputed'`, X is assumed to be a distance matrix and a feature array otherwise.
```

pyproject.toml
```diff
@@ -51,6 +51,7 @@ dependencies = [
     "torch~=2.3,<2.11.1",
     "torchvision~=0.18",
     "scipy~=1.14",
+    "scikit-learn~=1.6",
```
brownbaerchen, line 54: We need to remember to fix this before merging. But leave as is for now.

tests/cluster/test_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    print(f"Labels: {labels_np}")
+    print(f"HeAT Results: {res_np}")
+    print(f"SK Results: {sk_results}")
+
+        # Expected value for i=0 (Cluster 0)
+        # a = dist((0,0), (1,1)) = 1.414
+        # b = dist((0,0), (10,10)) = 14.14
+        # sil = (14.14 - 1.414) / 14.14 = 0.9
+
+    assert res_np[0] > 0.8, f"Point 0 is {res_np[0]:.4f}"
```
brownbaerchen:
```suggestion
    assert np.isclose(res_np[0], 0.9), f"Point 0 is {res_np[0]:.4f}"
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+
+    # a(i) calculation
+    a_mask = ht.reshape(labels_encoded, (1, -1)) == ht.reshape(
+        labels, (-1, 1)
+    )  # reshape((-1,1)) transposes labels_encoded
+    a_mask = a_mask.astype(ht.float32)
+
+    a_clust_dists = ht.sum(ht.mul(D, a_mask), axis=1)
+    denominator_a = labels_freqs.larray[labels_encoded.larray] - 1
+    denominator_a = ht.array(denominator_a, split=0)
```
brownbaerchen:
```suggestion
    denominator_a = ht.array(denominator_a, is_split=0)
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    denominator_a = ht.where(
+        denominator_a.astype(ht.float32) > 0,
```
brownbaerchen:
```suggestion
        denominator_a > 0,
```
?

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+        denominator_a,
+        1.0,
```
brownbaerchen:
```suggestion
        1,
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    )
+
+    a = ht.div(a_clust_dists, denominator_a)
+
+    # b(i) calculation
+    b_mask = ht.reshape(labels, (-1, 1)) == ht.reshape(unique_labels, (1, -1))
+    full_b_mask = (labels_encoded.reshape((-1, 1)) == unique_labels.reshape((1, -1))).astype(
+        ht.float32
```
brownbaerchen:
```suggestion
    full_b_mask = (labels_encoded.reshape((-1, 1)) == unique_labels.reshape((1, -1)))
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    )
+
+    b_clust_dists = ht.matmul(D, full_b_mask)
+
+    # labels_freqs.resplit_(None)
+    denominator_b = labels_freqs.astype(ht.float32)
```
brownbaerchen:
```suggestion
    denominator_b = labels_freqs
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    # denominator_b.resplit_(None)
+    b_clust_means = b_clust_dists / denominator_b
+
+    # we want neighbor cluster, so set the distance to points in own cluster to infinity
+    b_clust_means.larray[b_mask.larray > 0] = float("inf")
+    b = ht.min(b_clust_means, axis=1)
+
+    sil_samples = ht.div(ht.sub(b, a), ht.maximum(a, b))
+
+    sil_samples = ht.where(labels_freqs[labels_encoded] > 1, sil_samples, 0.0)
```
brownbaerchen:
```suggestion
    sil_samples = ht.where(labels_freqs[labels_encoded] > 1, sil_samples, 0)
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    if sample_size is not None:
+        X, labels = check_X_y(X, labels, accept_sparse=["csc", "csr"])  # same as silhouette_samples
+        if random_state is not None:
+            ht.random.seed(random_state)
+        indices = random_state.permutation(X.shape[0])[
+            :sample_size
+        ]  # selecs a subset of random samples, but all ranks need same indices
+
+        if metric == "precomputed":  # precomputed means here distance matrix for some reason?
+            X, labels = X[indices].T[indices].T, labels[indices]
```
brownbaerchen, line 357: @ClaudiaComito, is there a way to do this without communication and advanced indexing?

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    b_mask = ht.reshape(labels, (-1, 1)) == ht.reshape(unique_labels, (1, -1))
+    full_b_mask = (labels_encoded.reshape((-1, 1)) == unique_labels.reshape((1, -1))).astype(
+        ht.float32
+    )
+
+    b_clust_dists = ht.matmul(D, full_b_mask)
+
+    # labels_freqs.resplit_(None)
+    denominator_b = labels_freqs.astype(ht.float32)
+    # denominator_b.resplit_(None)
```
brownbaerchen:
```suggestion
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+
+    # b(i) calculation
+    b_mask = ht.reshape(labels, (-1, 1)) == ht.reshape(unique_labels, (1, -1))
+    full_b_mask = (labels_encoded.reshape((-1, 1)) == unique_labels.reshape((1, -1))).astype(
+        ht.float32
+    )
+
+    b_clust_dists = ht.matmul(D, full_b_mask)
+
+    # labels_freqs.resplit_(None)
```
brownbaerchen:
```suggestion
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    ht.sanitize_in(X)
+    ht.sanitize_in(labels)
+
+    unique_labels, labels_encoded = ht.unique(
+        labels, return_inverse=True
+    )  # f.e. labels = [10, 30, 20, 10], then unique_labels = [10,20,30] and labels_encoded = [0, 2, 1, 0]
+    unique_labels.resplit_(None)
+    # labels_encoded.resplit(None)
+    labels_freqs = ht.bincount(labels_encoded)
+    # labels_freqs.resplit_(None)
```
brownbaerchen:
```suggestion
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    X, labels = check_X_y(X, labels, accept_sparse=["csr"], metric=metric)
+
+    ht.sanitize_in(X)
+    ht.sanitize_in(labels)
+
+    unique_labels, labels_encoded = ht.unique(
+        labels, return_inverse=True
+    )  # f.e. labels = [10, 30, 20, 10], then unique_labels = [10,20,30] and labels_encoded = [0, 2, 1, 0]
+    unique_labels.resplit_(None)
+    # labels_encoded.resplit(None)
```
brownbaerchen:
```suggestion
```

tests/cluster/test_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+
+    res_edge = silhouette_samples(ht.array(X_edge), ht.array(labels_edge))
+    assert res_edge[3] == 0
+
+
+def test_minimal_silhouette():
+    X_np = np.array([[0, 0], [10, 10], [20, 20], [1, 1]], dtype=np.float32)
+    labels_np = np.array([0, 2, 1, 0], dtype=np.int32)
+
+    # HeAT values
```
brownbaerchen:
```suggestion
    # Heat values
```

tests/cluster/test_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    labels_ht = ht.array(labels_np, split=0)
+
+    ht_res = silhouette_samples(X_ht, labels_ht)
+
+    res_np = ht_res.numpy()
+
+    sk_results = sk_silhouette(X_np, labels_np)
+
+    print(f"Labels: {labels_np}")
+    print(f"HeAT Results: {res_np}")
```
brownbaerchen:
```suggestion
    print(f"Heat Results: {res_np}")
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+        An array of pairwise distances between samples, or a feature array.
+        If `metric='precomputed'`, X is assumed to be a distance matrix.
+    labels : DNDarray
+        Labels for each sample.
+    metric : str, optional
+        The metric to use when calculating distance between instances in a feature array.
+        If metric is "precomputed", X is assumed to be a distance matrix.
+        Default is "euclidean".
+    **kwds : optional
+        Additional keyword arguments are ignored (reserved for compatibility).
```
brownbaerchen, line 217: Please add information about the return values here

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    labels_freqs = ht.bincount(labels_encoded)
+    # labels_freqs.resplit_(None)
+    n_samples = labels.shape[0]
+    n_labels = unique_labels.shape[0]
+    check_number_of_labels(n_labels, n_samples)
+
+    if metric == "precomputed":
+        D = X
+    else:
+        D = ht.spatial.cdist(X, X)
```
brownbaerchen:
```suggestion
    elif metric == 'euclidean':
        D = ht.spatial.cdist(X, X)
     else:
         raise NotImplementedError(f'{metric=} not implemented')
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+
+    # a(i) calculation
+    a_mask = ht.reshape(labels_encoded, (1, -1)) == ht.reshape(
+        labels, (-1, 1)
+    )  # reshape((-1,1)) transposes labels_encoded
```
brownbaerchen, line 266:
```suggestion
    ) 
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    a_mask = a_mask.astype(ht.float32)
```
brownbaerchen:
```suggestion
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+
+    a_clust_dists = ht.sum(ht.mul(D, a_mask), axis=1)
+    denominator_a = labels_freqs.larray[labels_encoded.larray] - 1
```
brownbaerchen: We are all very confused why this is working since it's not really supposed to :D

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    )
+
+    b_clust_dists = ht.matmul(D, full_b_mask)
+
+    # labels_freqs.resplit_(None)
+    denominator_b = labels_freqs.astype(ht.float32)
+    # denominator_b.resplit_(None)
+    b_clust_means = b_clust_dists / denominator_b
+
+    # we want neighbor cluster, so set the distance to points in own cluster to infinity
```
brownbaerchen:
```suggestion
    # we want neighboring clusters, so set the distance to points in own cluster to infinity
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+Cluster metrics for the HeAT library.
+"""
+
+import heat as ht
+import numpy as np
+
+
+BORDER_LENGTH = 1e6  # value determines when array will be transformed to heat array
+
+
```
brownbaerchen:
```suggestion


```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+        )
+        # mb write a function to fill diag with 0, like np.fill_diagonal(X, 0)
+        diag_elements = ht.diag(X)
+
+        atol = ht.finfo(X.dtype).eps * 100  # tolerance based on machine accuracy
+
+        if ht.any(ht.abs(diag_elements) > atol):
+            raise error_msg
+        elif ht.any(diag_elements != 0):  # integral dtype
+            raise error_msg
```
brownbaerchen, line 105:
```suggestion
```
Not sure this is needed. I definitely wouldn't change the input. If we really want to make sure it's a valid distance matrix, we should also check if it's symmetric. What do you think, @ClaudiaComito?

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    --------
+    >>> import heat as ht
+    >>> X = [[0, 1], [1, 0]]
+    >>> check_array(X, metric="precomputed")
+    DNDarray([[0., 1.], [1., 0.]], dtype=ht.float32, device=cpu:0, split=0)
+    """
+    # Convert to heat array if input big enough --> overhead acceptable if array big enough
+    if X.shape[0] > BORDER_LENGTH or X.shape[1] > BORDER_LENGTH:
+        if not isinstance(X, ht.DNDarray):
+            X = ht.array(X, split=0)
```
brownbaerchen:
```suggestion
    ht.sanitize_in(X)
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    Examples
+    --------
+    >>> import heat as ht
+    >>> y = ht.array([[1], [2], [3]])
+    >>> _check_y(y)
+    DNDarray([1, 2, 3], dtype=ht.int64, device=cpu:0, split=0)
+    """
+    if y.shape[0] > BORDER_LENGTH:
+        if not isinstance(y, ht.DNDarray):
+            y = ht.array(y, split=0)
```
brownbaerchen:
```suggestion
    ht.sanitize_in(y)
```

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+
+    # need 1D labels
+    if len(y.shape) > 1:
+        if y.shape[1] == 1:
+            y = y.reshape(-1)
+        else:
+            raise ValueError("y should be a 1D array or a column vector.")
+
+    if y is None:
+        raise ValueError(" requires y to be passed, but the target y is None")
```
brownbaerchen, line 146:
```suggestion
```
this can never be reached since `None` has no `len`

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+        raise ValueError(
+            f"Found input variables with inconsistent numbers of samples: [{X.shape[0]}, {y.shape[0]}]"
+        )
+
+    if isinstance(X, ht.DNDarray) and isinstance(y, ht.DNDarray):
+        # ensure split axis is the same
+        if X.split != y.split:
+            y.resplit_(X.split)
+        # Check if local chunks match
+        if X.lshape[0] != y.lshape[0]:
```
brownbaerchen, line 187:
```suggestion
        if X.lshape[X.split] != y.lshape[y.split]:
```
right? Not sure about this.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,397 @@
...
+    >>> X = ht.array([[1, 2], [1, 1], [4, 4], [4, 5]], split=0)
+    >>> labels = ht.array([0, 0, 1, 1], split=0)
+    >>> ht.cluster.silhouette_samples(X, labels)
+    DNDarray([0.7452, 0.7836, 0.7452, 0.7836], dtype=ht.float64, device=cpu:0, split=0)
+    """
+    # Sanitation and checks
+    X, labels = check_X_y(X, labels, accept_sparse=["csr"], metric=metric)
+
+    ht.sanitize_in(X)
+    ht.sanitize_in(labels)
```
brownbaerchen:
```suggestion
```

tests/cluster/test_metrics.py
brownbaerchen: actually, the test should be in `../../tests/test_cluster`, relative to where this file is now.

heat/tests/test_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+def test_silhouette_implementation():
+    n_samples = 100
+    n_features = 5
+    centers = 3
+    X_np, labels_np = make_blobs(n_samples=n_samples, n_features=n_features, centers=centers, random_state=42)
+
+    # Reference values from Scikit-Learn
+    sk_results = sk_silhouette(X_np, labels_np)
+
+    # HeAT values
```
brownbaerchen:
```suggestion
    # Heat values
```

heat/tests/test_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+from sklearn.metrics import silhouette_samples as sk_silhouette
+from sklearn.metrics import silhouette_score as sk_silhouette_score
+from sklearn.datasets import make_blobs
+from heat.cluster.metrics import silhouette_samples, silhouette_score
+
+
+def test_silhouette_implementation():
+    n_samples = 100
+    n_features = 5
+    centers = 3
```
brownbaerchen: Can you please run tests with different combinations of these parameters? Since we fix the random seed, this is only testing a single configuration ever to medium tolerance, which is not conclusive evidence that this actually works.

heat/tests/test_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    X_ht = ht.array(X_np, split=0)
+    labels_ht = ht.array(labels_np, split=0)
+
+    ht_results = silhouette_samples(X_ht, labels_ht)
+
+    # Comparison
+
+    ht_results_np = ht_results.resplit(None).numpy()
+
+    assert np.allclose(sk_results, ht_results_np, atol=1e-5), f'Max diff between Heat and scipy: np.max(np.abs(sk_results - ht_results_np))'
```
brownbaerchen:
```suggestion
    assert np.allclose(sk_results, ht_results_np, atol=1e-9), f'Max diff between Heat and scipy: np.max(np.abs(sk_results - ht_results_np))'
```
Would be good to have a tighter tolerance here. Maybe the input needs to be cast to float64 if that is not already the case.

brownbaerchen (changes requested):
This review mostly contains what we discussed in the meeting. It may look like a lot of comments, but most of them should be very quick to implement and some of them maybe you choose not to implement, so don't worry :D

heat/tests/test_metrics.py
Shabasovich: i tried:

- n_samples=20000, n_features = 100, centers = 200
- n_samples=10000, n_features = 5000, centers = 5
- n_samples=10000, n_features = 5, centers = 5000

Everything works with tolerance 1e-9 except last one. It works for 1e-5 though.

heat/cluster/metrics.py
Shabasovich: It turns out you can completely remove larray and it still works. I also deleted line `denominator_a = ht.array(denominator_a, split=0)` and it also works fine

heat/tests/test_metrics.py
brownbaerchen: Can you push these tests here and maybe find a compromise tolerance like 1e-8?

Shabasovich: What do you mean push them here? So the same code of the test but with different values? I mean the tests that i ran are already here in `tests/test_metrics`

brownbaerchen: Hm? I don't see it there. But yes, you can just do nested four loops to try a bunch of combinations.

Shabasovich: Isn't it what you want?: https://github.com/helmholtz-analytics/heat/blob/features/2190-Implement_silhouette_score/tests/cluster/test_metrics.py
(without loop in test yet)

brownbaerchen: Ah yes, that is fine! Sorry, seems there were some problems with what GitHub was showing me. I made an effort to look at the latest version, but I can't say for sure that GitHub was the problem and not me :D

brownbaerchen:
I ran some scaling tests on JUSUF with your implementation and they look really good, but maybe too good?
The more tasks we use, the larger a problem we need in order to remain computation bound and benefit from parallelization, which is to be expected. So for very small problems, serial sklearn is best. But as soon as we increase the size of data, we benefit from parallelization and heat wins.
However: We win by too much in heat. We would expect to get speedup of at most two if we distribute the work among two tasks, but instead we get speedup of around 6 for the largest problem size (3s on one task vs 0.5s on two tasks).
I am not sure what's going on here, but maybe the algorithm does "too much work" per task?
In any event, your code significantly outperforms sklearn, which is good. For all the experiments shown here, the score is also the same. However, there were some edge cases where the score differed. Not sure what happened there.

Figure_1

Seems this super-optimal scaling is coming from functions like heat.spatial.cdist. Am I missing something @ClaudiaComito? I probably am just doing something silly and I don't think you should worry about it, @Shabasovich.

heat/cluster/metrics.py
brownbaerchen, line 1: Before we merge, we need to adapt these to the tests to the test class structure of unittest like the other heat tests.

tests/cluster/test_metrics.py
brownbaerchen, line 1: Before we merge, we need to adapt these to the tests to the test class structure of unittest like the other heat tests.

tests/cluster/test_metrics.py
```diff
@@ -0,0 +1,237 @@
...
+
+        # Single sample in a cluster
+        labels_edge = np.array([0, 0, 0, 1])
+        X_edge = np.random.rand(4, n_features)
+
+        res_edge = silhouette_samples(ht.array(X_edge), ht.array(labels_edge))
+        assert res_edge[3] == 0
+
+
+    def test_edge_case(self):
```
brownbaerchen: I don't know what's special about this case, but it does fail...

tests/cluster/test_metrics.py
brownbaerchen, line 1: I wonder if we should stop doing the k means classification here. I think it suffices to choose the labels randomly. After all, we don't care if the classification is good, we just want to know if the score of any given classification is correct. Using random labels should make these tests much faster to run.
What do you think @Shabasovich @ClaudiaComito? Any reasons to prefer non-random labels in the tests?

heat/cluster/metrics.py
```diff
@@ -0,0 +1,400 @@
...
+    >>> ht.cluster.silhouette_score(X, labels)
+    0.76439
+    """
+    if sample_size is not None:
+        X, labels = check_X_y(X, labels, accept_sparse=["csc", "csr"])  # same as silhouette_samples
+        if random_state is not None:
+            ht.random.seed(random_state)
+        else:
+            ht.random.seed(None)
+        # indices = random_state.permutation(X.shape[0])[:sample_size]
```
brownbaerchen:
```suggestion
```
Given that we don't use the random state anymore, the previous comment is outdated and there is no need to set a seed here.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,390 @@
...
+    >>> import heat as ht
+    >>> X = ht.array([[1, 2], [1, 1], [4, 4], [4, 5]], split=0)
+    >>> labels = ht.array([0, 0, 1, 1], split=0)
+    >>> ht.cluster.silhouette_samples(X, labels)
+    DNDarray([0.7452, 0.7836, 0.7452, 0.7836], dtype=ht.float64, device=cpu:0, split=0)
+    """
+    # Sanitation and checks
+    X, labels = check_X_y(X, labels, accept_sparse=["csr"], metric=metric)
+
+    unique_labels, labels_encoded = ht.unique(
```
brownbaerchen, line 247: I figured out what's wrong with the special case below. k means returns the labels `[0, 1, 3]`, but the way the masking matrices are set up, it assumes the labels go continuously, i.e. `[0, 1, 2]`, because you compare the unique labels directly with the labels encoded. So, in the matrix, nothing is ever assigned to cluster 3 because the maximal value in labels encoded is 2.
I think this can be easily fixed by replacing `unique_labels.resplit_(None)` with `unique_labels = ht.arange(ht.max(labels_encoded)+1)`.

tests/cluster/test_metrics.py
brownbaerchen: The problem here was that some clusters were assigned no values, i.e. the unique labels where 0, 1, and 3. The implementation required them to be continuous. I think I fixed this and implemented a cleaner test elsewhere.

brownbaerchen:
I removed the check for the number of labels. It required to have between 2 and n-1 distinct labels. I realize a clustering with a single label and with n labels is not very useful because it means either everything is assigned to the same cluster or everything is assigned to its own cluster. Nevertheless, I don't see a reason not to compute the silhouette score for these clusterings. They are valid clusterings and I wouldn't want to tell anybody that they can't compute a score for them.
I know you copied this from the [scikit learn implementation](https://github.com/scikit-learn/scikit-learn/blob/fe2edb3cdbd75ae4e662fda67dcb19277258792b/sklearn/metrics/cluster/_unsupervised.py#L31), @Shabasovich, but I think here we can deviate from their interface this much.
Note that the score is 0 in these cases which is nice because it's an obviously bad score and an appropriate output.

tests/cluster/test_metrics.py
brownbaerchen: Ups. I confused myself. There is already none of that here. Please ignore this.. Sorry!

pyproject.toml
```diff
@@ -84,6 +84,16 @@ dev = [
     "perun",
 ]
 
+## dependencies used for IO and verification in test suite
+tests = [
```
brownbaerchen, line 88: @JuanPedroGHM, @ClaudiaComito, @mtar what do you think about dedicated test dependencies? Here, we want to verify against scikit-learn, so we need to install it, but we don't want to add it as dependency. If we do this, we should add it to the codebase CI as well.

tests/core/test_manipulations.py
brownbaerchen, line 1: This is part of #2253, which should be merged before this PR.

heat/core/manipulations.py
brownbaerchen, line 1: This is part of #2253, which should be merged before this PR.

benchmarks/cb/cluster.py
brownbaerchen, line 1: Do we want to add benchmarks of this feature at all? Do we want to split that in another PR? Or do we clean this up? :D

brownbaerchen (dismissed):
There are three things left to do before merging this from my point of view
 - [x] Figure out how to handle the scikit-learn dependency of the tests
 - [x] Remove or clean up the benchmarks
 - [x] Merge #2253, which this depends on
 - [x] Merge #2262, which this depends on

I think the implementation and the tests are ready to be merged. I did a lot of refactoring since the scikit-learn code that some of this is based on could be more expressive and concise in my opinion.

Note that the tests are probably skipped on codebase so far because there scikit-learn is not installed and I am using `pytest.importorskip` here.

I am approving this PR in this review even though it still needs a little work since my previous one was blocking and I don't need to be blocking this PR at this time.

Again, I want to thank @Shabasovich for the work and applaud the creative implementation that works in parallel with no MPI calls at all.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,272 @@
...
+    if metric == "precomputed":
+        D = X
+    elif metric == "euclidean":
+        D = ht.spatial.cdist(X, X)
+    else:
+        raise NotImplementedError(
+            f"{metric=} not implemented, choose from 'precomputed' and 'euclidean'."
+        )
+
+    # a(i) calculation
```
JuanPedroGHM, line 159: Duplicate reshapes could cause unnecessary communication. Saving the duplicate reshapings for reuse is probably useful.

heat/cluster/metrics.py
```diff
@@ -0,0 +1,272 @@
...
+    # b(i) calculation
+    b_mask = ht.reshape(labels, (-1, 1)) == ht.reshape(unique_labels, (1, -1))
+    full_b_mask = labels.reshape((-1, 1)) == unique_labels.reshape((1, -1))
+
+    b_clust_dists = ht.matmul(D, full_b_mask)
+
+    denominator_b = labels_freqs
+    b_clust_means = ht.div(b_clust_dists, denominator_b)
+
+    # we want neighboring clusters, so set the distance to points in own cluster to infinity
```
JuanPedroGHM, line 181: Could `a_clust_dists` be calculated from this same matrix, by instead setting `b_clust_means.larray[b_mask.larray == 0] == 0`, and then `ht.sum(..., axis=1)`?

It would replace the element-wise product, mask creation, and reduction with a copy, local write, and a reduction. It might require less communication and computation, so it would be worth testing this.

brownbaerchen: I don't really understand, but I don't think so. Looking at the example Sofya shows in the PR description, I don't see a straightforward way to extract the a distance from the b one.

JuanPedroGHM: Based on the example of the $D$ calculation for $b$, each row $i$ contains the sum of the distances of element $i$ to all the elements of the cluster in column $j$. So, the distance to the elements of its own cluster can be extracted by using the mask that was used to compute this, taking the non-zero element instead of the min of zero elements. And that should be the same value that is needed for $a$. 

I tried quickly, and it works, here is the modified code. Caveat: tolerance has to be raised for `test_silhouette_random_sampling` for the line 149 to `1e-6`

```python
    # a(i) calculation
    labels_column = labels.reshape((-1, 1))  # reshape((-1,1)) transposes the labels

    unique_labels_row = unique_labels.reshape((1, -1))
    full_mask = labels_column == unique_labels_row

    clust_dists = ht.matmul(D, full_mask)
    a_clust_dists = ht.sum(ht.mul(clust_dists, full_mask), axis=1)

    denominator_a = labels_freqs[labels_encoded] - 1
    denominator_a = ht.where(
        denominator_a > 0,
        denominator_a,
        1,
    )

    a = ht.div(a_clust_dists, denominator_a)

    # b(i) calculation
    denominator_b = labels_freqs
    b_clust_means = ht.div(clust_dists, denominator_b)

    # we want neighboring clusters, so set the distance to points in own cluster to infinity
    b_clust_means.larray[full_mask.larray > 0] = float("inf")
    b = ht.min(b_clust_means, axis=1)

```

brownbaerchen: Oh yes, you're right! Good catch!

JuanPedroGHM (dismissed):

brownbaerchen (approved):

PR #2208, opened 2026-03-17
Update README and sync with Quick Start, contributing.md

ClaudiaComito:
## Due Diligence

- General:
    - [x]  **title** of the PR is suitable to appear in the [Release Notes](https://github.com/helmholtz-analytics/heat/releases/latest)
- Implementation:
    - [ ] unit tests: all split configurations tested
    - [ ] unit tests: multiple dtypes tested
    - [ ] **NEW** unit tests: MPS tested (1 MPI process, 1 GPU)
    - [ ] benchmarks: created for new functionality
    - [ ] benchmarks: performance improved or maintained
    - [x] documentation updated where needed

## Description

Reorganized README, quick_start and contributing to: a) be consistent with each other, b) remove redundancy, c) minimize statements that need to be updated for every release. 

I have removed explicit mention of all our institutions from the README, and only left in the funding projects. 

  

Issue/s resolved: #2067 

## Changes proposed:

- removed explicit mention of all our institutions from the README
- added HeatHub funding
- reorganized Quick Start and contributing
- added a LLM and AI usage section to contributing.md
-

#### Does this change modify the behaviour of other functions? If so, which?
no

PR #2216, opened 2026-03-23
Remove checksum ignores from .talismanrc

brownbaerchen:
In #2208, we accidentally added a bunch of ignores to the talismanrc. These are supposed to protect us from merging secrets to the repository, but they are not secrets, they are only interpreted that way by talisman and incorrectly classified.

I suggest rather than adding every individual instance of this to talismanrc, which will make it huge and confusing over time, we should simply ignore talisman during the commit. So I think we should be doing:
```bash
git commit
<ruff and talisman complain>
git commit
<only talisman complains with this checksum thing>
git commit --no-verify
```
this has the same outcome as explicitly adding all potential secrets to `.talismanrc`, but it's a lot less annoying.

Having a check that you didn't accidentally upload some secret is not a bad idea. I am not proposing to get rid of talisman. I just suggest to deal differently when it tells us we might be uploading a secret.


JuanPedroGHM:
I think the bigger issues seems to be that the output of the jupiter notebooks was not properly stripped out by one of our pre-commits. Either that, some of the images appear as password looking strings, in which case leaving the ignore statements is fine. Do you know which one is which?

brownbaerchen:
No, I don't know what exactly is the password-looking thing. But this PR is only about how we handle the ignoring of this. I don't see a reason to explicitly ignore this as pre-commit is happy if we don't. I suppose that if we add an explicit ignore for every such instance, the config will just grow to be huge over time.

If you disagree and you'd rather explicitly ignore these things, feel free to just close this PR.

PR #2246, opened 2026-04-14
[WIP] MPI-Free Communicator

### LINKED ISSUE #1978: Lazy MPI initialization
## Feature functionality
We have had some issues when trying to run heat on a single rank if ```mpirun/mpiexec``` is not used. This is because MPI initialization causes a crash if there is no actual MPI context to find. 

This problem could be handled better if we refactor our communication layer to initialize MPI "lazily".  This would require some refactoring involved. 

**Needed**

*  ```mpi4py``` is only imported in the communication module. ```MPI``` is made available through the communication module or object.
* ```mpi4py.rc.initialize = False``` is set before importing ```MPI```
* ```MPICommunicator``` is only created when it is needed.
* ```MPI.Init()``` is called on the ```MPICommunicator.__init___``` method.

This alone does not solve the problem, as it will still crash if heat was called without ```mpirun/mpiexec```. What we require, is a way to detect if the application was called with MPI, and I'm not sure if it exists. 
 
* One possible solution is to make ```mpi4py``` an optional dependency, and to create a dummy communicator that does nothing. 
* A possibly easier solution is to check the command that was used to start python (would not work on notebooks, probably) and check if ```srun/mpirun/mpiexec``` was used.

## Related 

* #1971 : To verify the correct installation of heat, the command line utility "heat -i" could be used. This is being blocked at the moment because MPI is initialized, so the sanity check needs to be run with ```mpirun/mpiexec```

mtar:
## Due Diligence

- General:
- Implementation:

## Description

Allows to run heat on a single process without MPI Communication for very simple testing purposes. It supports the core module exclusively. 

Issue/s resolved: #1978

## Changes proposed:

- introduce boolean HAVE_MPI 
- add parameter op_type in __reduce_op for activating arg and bool specifigc code instead of looking at the the MPI.Op


brownbaerchen:
To be honest, I still don't understand the problem that this is solving. Can you maybe post an example in the associated issue? 
I am not sure that adding this complexity is really necessary considering MPI is really integral to Heat in practice, Heat doesn't make a lot of sense without MPI, and MPI is easy enough to obtain with package managers or the module system on an HPC machine.

mtar:
On our HPC machines, we don't have MPI on the login nodes. So it is an option, if you want to check if the installation is basically working without spending compute time.
A concrete issue is the installation through easybuild. There we had to skip the sanity checks on the login nodes because heat cannot be used there.

brownbaerchen:
I just tried again and I can run the whole test suite in serial on a login node on Jureca after loading an MPI module.

mtar:
Yes, you're right, it works again on the login nodes. Thanks for checking.

> I am not sure that adding this complexity is really necessary...

What you mean by added complexity? I mostly had to skip some MPI related calls and refine some if conditions. I only added a very small dummy Communicator and a helper Enum class due to an issue in __reduce_op.

As the import of heat is working again for now, I'm also fine with removing some of the changes regarding not having MPI available, but I would rather keep the additional checks skipping some of our own MPI function wrappers and/or add an early exit in these functions to avoid some overhead on a single process.

brownbaerchen:
You're right. You didn't add that much complexity. I was just confused by the large diff.. Sorry about that!
Actually, this enum is not a bad idea considering we may want to add a NCCL communication backend at some point. In the past, I have wrapped that such that input MPI ops are mapped to NCCL ops, but it's cleaner to put a generic parallel op, maybe. 👍

mtar:
Yeah, the diff sometimes  gets confused :dizzy_face:

This might be good idea. When I made my enum closer to the MPI ID to save me a mapping, I realized that it works well with the predefined MPI Ops, but it is not guaranteed that a custom MPI op stays at the same number if code gets changed or the custom op is freed. A more general one would be better if we like to support other communications.

brownbaerchen:
I am not sure how easy it is to add sth like NCCL because it is a bit different. Most notably, there is no non-blocking communication in NCCL.
When in MPI, to exchange data between two tasks you do
```
req = comm.isend((comm.rank+1) % 2)
comm.recv((comm.rank+1)%2)
req.wait()
```
in NCCL you do
```
comm.NCCL_group_start()
comm.send((comm.rank+1)%2)
comm.recv((comm.rank+1)%2)
comm.NCCL_group_end()
```
(up to the actual syntax, please read this as pseudo code)

The point is that in NCCL, you start and end a group and within that group any communication can overlap. But this makes it quite different to port code where the send and receive operations are not back-to-back, but scattered across different functions. Do we have scattering of communication in some places?

Another problem with NCCL is that it doesn't support a lot of the collectives that MPI supports. You have to build them yourself with groups. This is not too much work, but can be annoying.

We should probably have a think about how large of an investment this is and whether it's really needed. MPI requires synchronization between CPU and GPU before sending, which kills strong scaling at some point and NCCL doesn't have this issue. On the other hand, Heat is focusing on weak scaling where the sync overhead is small compared to computation time, so... My feeling is that this would be a lot of work with little real-world reward.

brownbaerchen:
Wait, why are you doing this again? I thought this is a preparation for a NCCL / whatever backend, but the changes I see now just change behavior when there is no MPI installation present. This doesn't get us any closer to other communication backends, right?

mtar:
Yes, you're right. I scaled the changes back down to a minimum level. MPI is an integral part of Heat and, while bypassing these functions works fine for the non-communication, I'd like to discuss the changes needed for the support of other backends first before setting something up in that direction.

brownbaerchen:
Adding other backends is a lot of work indeed and of course you don't need to do it now. But why/when would we want to bypass the MPI functions?

mtar:
The changes will complicate the code for a special use case that is out of scope from the main goals of Heat. Some parts, however, will be reused.

PR #2250, opened 2026-04-20
Bugs/2093  bug errors in edge cases in parallel matrix multiplication

### LINKED ISSUE #2093: [Bug]: Errors in edge cases in parallel matrix multiplication
### What happened?

In trying to debug something, I encountered this weird edge case where the result of parallel matrix multiplication is wrong. The error is detailed below and appears only when run with two tasks. I am not sure what makes this edge case special.
I have been running into this issue first with QR decomposition. You can try:
```
import heat as ht

comm = ht.comm
A = ht.random.randn(comm.size, comm.size, dtype=ht.float32, split=0)
QR = ht.linalg.qr(A)

matmul_success_loc = ht.allclose(QR.Q @ QR.R, A)
matmul_success_glob = ht.allclose((QR.Q.resplit(None) @ QR.R.resplit(None)).resplit(QR.Q.split), A)
print(matmul_success_loc, matmul_success_glob)
```
which will print `False, True`.

I am a bit lost here. In particular, the matrix multiplication function is very long and hard to understand. So I thought I just report this bug because it seems pretty bad and maybe somebody can give me some pointers. 

```shell
import heat as ht

split = 0
shape = (4, 3)

A = ht.ones(shape, split=split)
B = ht.ones(shape[::-1], split=split)

C = A @ B
C_glob = (A.resplit(None) @ B.resplit(None)).resplit(C.split)

print(C)
print(C_glob)
```

```shell
Output when run with two tasks: 

DNDarray([[3., 3., 3., 3.],
          [6., 6., 6., 6.],
          [3., 3., 3., 3.],
          [6., 6., 6., 6.]], dtype=ht.float32, device=cpu:0, split=0)
DNDarray([[3., 3., 3., 3.],
          [3., 3., 3., 3.],
          [3., 3., 3., 3.],
          [3., 3., 3., 3.]], dtype=ht.float32, device=cpu:0, split=0)
```

### Version

main (development branch)

### Python version

3.13

### PyTorch version

2.9

### MPI version

```shell
OpenMPI 5.0.8
```

brownbaerchen:
As documented in #2093, we have some edge cases in matrix multiplication that fail when there are too few data per task. Given that matrix multiplication is essential and that we often use small matrix multiplication in testing, we should fix that at some point. This PR adds tests for all edge cases that we find such that we can work on fixing them.


brownbaerchen:
@copilot, please fix the bugs regarding edge cases in matrix multiplication that are discussed in #2093 and pointed out in the tests in this PR. Please create a PR with your suggested fix onto the branch this PR wants to merge to main.

brownbaerchen:
@copilot, please fix the bugs regarding edge cases in matrix multiplication that are discussed in https://github.com/helmholtz-analytics/heat/issues/2093 and pointed out in the tests in this PR. Please create a PR with your suggested fix onto the branch this PR wants to merge to main.

brownbaerchen:
Scaling at this point comparing this branch to the current main:

matmul_scaling_copilot

We see that performance is unchanged for large matrices but better for small matrices.

In case you are confused that split 10 performs worse that other combinations here but not in similar plots that I posted previously, that is because I was being very silly. The previous plots show split 00 for all lines by accident..

brownbaerchen:
Ok, I did some more refactoring, and I think this is now ready. Below is the latest performance plot:
matmul_scaling_copilot

The AI noticed that a lot of stuff can be removed and I removed even more stuff. I added some more tests and this is certainly less buggy than the current main, faster for small matrices, and much fewer lines. 
This integrates the changes from #2304 and #2288, so we can close those PRs.

The implementation of the split 10 case is totally different from other cases and much slower. We should maybe refactor that in the future, but this PR is large enough as is and I just moved all the stuff that is specific to that case there.

brownbaerchen:
Actually, simply doing a resplit of b and then using the implementation for split 11 in the split 10 case is both faster and much easier. Here is the performance:
matmul_scaling_copilot
So it's not a massive gain, but it's not negligible and much easier to maintain.
We can revert this last commit and wait for a proper refactor of the split 10 case if you prefer.

mtar:
Interesting, so moving the data to another split axis is still better on larger matrices than going with the worst combination. How many MPI processes did you use? Have you also looked at the performance on multinode or on more processes than GPUs to see how well it runs on smaller local arrays?
Edit: I see that you show some strong scaling in results in one of the other PRs.

brownbaerchen:
> Interesting, so moving the data to another split axis is still better on larger matrices than going with the worst combination. How many MPI processes did you use? Have you also looked at the performance on multinode or on more processes than GPUs to see how well it runs on smaller local arrays? Edit: I see that you show some strong scaling in results in one of the other PRs.

Here is the same scaling but with 16 GPUs (four nodes):
matmul_scaling_copilot
Split 10 is really abysmally slow on the current main. I reran twice to make sure this was repeatable. I stopped this scaling early because it takes super long.

Also, I forgot to mention that this is better in terms of memory consumption: The current main allocates a temporary result array that is the whole size and non-distributed. So even though the resplit is not nice, it seems preferable to what is here at the moment. But we can split this apart and merge the bugfixes and the refactor for split 10 separately.

I always use one MPI task per GPU.

heat/core/linalg/basics.py
```diff
@@ -800,473 +798,85 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
-        # split la dims 10
-        elif a.split == ndim - 1 and b.split == ndim - 2:
-            # todo: this may create the full matrix on evey process, issue #360
-            # for this case, only a sum is needed at the end
-            a_rem_locs1 = torch.nonzero(rem_map[:, 0, 1] == 1, as_tuple=False)
-            # locations of the remainders in b
-            b_rem_locs0 = torch.nonzero(rem_map[:, 1, 0] == 1, as_tuple=False)
-            res = torch.zeros(
-                (*batch_shape, a.gshape[-2], b.gshape[-1]), dtype=c_type.torch_type(), device=tdev
-            )
```
brownbaerchen, line 1257: This is the problematic line allocating the huge, global, unsplit, temporary result tensor in split case 10

tests/core/linalg/test_basics.py
mtar, line 1: We still want to support the unittest library.

heat/core/linalg/basics.py
```diff
@@ -680,13 +680,11 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
 
-    # single-process setup, torch matmul
+    # early out with torch matmul if input is not distributed
     if a.comm.size == 1:
         c = factories.array(torch.matmul(a.larray, b.larray), dtype=c_type, device=dev)
 
-    # early out for vector vector multiplication
-    # is this even covered in the tests?
-    # seems to be used in test_qr
+    # early out for vector vector multiplication TODO: output should be split like input
```
mtar: As far as I remember, the different split was kept because it would safe us some additional data transfer on all processes.

heat/core/linalg/basics.py
```diff
@@ -800,473 +798,85 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
-            )  # shouldnt this always be kB and be the same as for split 11?
-
-        if (kB == 1 and a.lshape[-1] != 1) or a.lshape[
-            -1
-        ] % kB != 0:  # does kb == 1 imply a.lshape[-1] > 1?
-            rem_a = 1
-        if (kB == 1 and b.lshape[-2] != 1) or b.lshape[-2] % kB != 0:
-            rem_b = 1
-
+    else:  # both matrices are split in la dims
```
mtar: What does 'la dims' mean?

heat/core/linalg/basics.py
```diff
@@ -800,473 +798,85 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
-            else:
-                r = None
-                r_loc = None
-
-            req = {}
-            b_lp_data = {}
+        # split dims 00 and 01
+        if (a.split == ndim - 2 and b.split == ndim - 2) or (
+            a.split == ndim - 2 and b.split == ndim - 1
+        ):
```
mtar, line 835: The style doesn't look consistent here.

heat/core/linalg/basics.py
```diff
@@ -800,473 +798,85 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
+            reqs, b_chunks = {}, {}
+            # broadcast chunks of b
             for pr in range(comm.size):
```
mtar: I think, `proc` would be clearer name than `pr`

heat/core/linalg/basics.py
```diff
@@ -800,473 +798,85 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
-                r = None
-                r_loc = None
-
-            req = {}
-            b_lp_data = {}
+        # split dims 00 and 01
+        if (a.split == ndim - 2 and b.split == ndim - 2) or (
+            a.split == ndim - 2 and b.split == ndim - 1
+        ):
+            reqs, b_chunks = {}, {}
```
mtar, line 837: What's the reason to use dictionaries here?

heat/core/linalg/basics.py
```diff
@@ -800,473 +798,85 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
+            # broadcast chunks of b
             for pr in range(comm.size):
-                # ibcast data on node first
                 if comm.rank == pr:
-                    b_lp_data[pr] = b.larray.clone()
+                    b_chunks[pr] = b.larray.clone()
                 else:
-                    b_lp_data[pr] = torch.zeros(
-                        (*batch_shape, lshape_map[pr, 1, -2].item(), lshape_map[pr, 1, -1].item()),
+                    b_chunks[pr] = torch.empty(
```
mtar: Is it fine to use torch.empty here? It could contain arbitrary values.

mtar (commented):
I would be nice if we could reduce the complexity of this function 😄 
Does the change in 10-split has an affect on the CPU?

heat/core/linalg/basics.py
brownbaerchen: To be honest, the `allow_resplit` option is kinda strange to me because it returns the input with different split after it's done with the function. I guess I phrased this TODO poorly and I'll remove it.

heat/core/linalg/basics.py
brownbaerchen: This is used a few times throughout the code. I am not really sure myself. I'll rename `la dim`->`non batch dim`

heat/core/linalg/basics.py
brownbaerchen: well, why not? :D Would you prefer something else?

heat/core/linalg/basics.py
brownbaerchen: pre-commit did that. I don't think we have control over this..

heat/core/linalg/basics.py
brownbaerchen: Yes it's fine. This is just a buffer for the `Ibcast` bellow.

tests/core/linalg/test_basics.py
brownbaerchen: We decided in the meeting last Wednesday when you were at the Easybuild thing that we embrace pytest.

brownbaerchen:
@mtar, here are scaling results on CPU:
matmul_scaling_copilot
Interestingly, here, the 00 case is slower than the others.. this is within a single node, maybe multi-node would be different.
Anyways, you can't clearly see it, but on main, 10 is as slow as 00, whereas on this branch, 10 is as fast as 11 and 01. So it's also faster than on main. I guess the communication cost is negligible here.

tests/core/linalg/test_basics.py
mtar: Okay :+1:

heat/core/linalg/basics.py
mtar: The items are created and accessed inside for loops, I thought a preallocated list may be faster here on large process numbers, but we will most likely not reach that number to have a significant impact on the performance.

brownbaerchen:
FYI, here is CPU scaling on 128 cores with 32 tasks per node (four nodes in total):
matmul_scaling_copilot

JuanPedroGHM:
# Benchmarks results - Sponsored by [perun](https://github.com/Helmholtz-AI-Energy/perun)

## FTPx86-AMD-Milan-Mi250 - CPU - N4 - RUNTIME

| Function                                 |   Value [s] |   Change % |   30 day avg. [s] |   Last entrie [s] |
|------------------------------------------|-------------|------------|-------------------|-------------------|
| lanczos                                  | 0.371604    |   -19.5608 |        0.461969   |        0.464249   |
| randomized_svd_split0                    | 0.0347957   |    66.626  |        0.0208825  |        0.0223394  |
| randomized_svd_split1                    | 0.0406784   |   200.712  |        0.0135273  |        0.0163467  |
| randomized_eigh_split0                   | 0.0145305   |    43.5901 |        0.0101194  |        0.0139501  |
| randomized_eigh_split1                   | 0.0251167   |    79.7139 |        0.0139759  |        0.0195706  |
| kmedoids                                 | 0.357558    |    11.8609 |        0.319645   |        0.322634   |
| reshape                                  | 0.194488    |   -12.5554 |        0.222413   |        0.236636   |
| apply_inplace_min_max_scaler_and_inverse | 0.00683279  |    78.3737 |        0.0038306  |        0.00497179 |
| apply_inplace_max_abs_scaler_and_inverse | 0.000619221 |   -64.2798 |        0.00173353 |        0.00184145 |
| apply_inplace_normalizer                 | 0.00638704  |    10.9089 |        0.00575882 |        0.00696373 |
| lasso_predict                            | 0.00672913  |   -14.3407 |        0.00785568 |        0.00730591 |
---

## FTPx86-AMD-Milan-Mi250 - GPU - N4 - RUNTIME

| Function                                  |   Value [s] |   Change % |   30 day avg. [s] |   Last entrie [s] |
|-------------------------------------------|-------------|------------|-------------------|-------------------|
| matmul_split_0                            |  0.0879359  |   -18.4942 |        0.107889   |        0.104038   |
| matmul_split_1                            |  0.0270488  |   -44.6389 |        0.0488588  |        0.0511468  |
| lanczos                                   |  1.83512    |   -30.6103 |        2.64466    |        2.5851     |
| randomized_svd_split0                     |  0.0400271  |   -14.4965 |        0.0468134  |        0.047465   |
| randomized_eigh_split0                    |  0.0385161  |   -37.7225 |        0.061846   |        0.0634295  |
| randomized_eigh_split1                    |  0.0613647  |   -24.5851 |        0.0813694  |        0.0838963  |
| kmedoids                                  |  1.70747    |    10.0897 |        1.55098    |        1.52023    |
| concatenate                               |  0.128119   |    12.8386 |        0.113542   |        0.116033   |
| apply_inplace_standard_scaler_and_inverse |  0.0877387  |    33.4879 |        0.0657278  |        0.0663693  |
| apply_inplace_min_max_scaler_and_inverse  |  0.00271235 |   -15.9391 |        0.00322664 |        0.00240974 |
| apply_inplace_max_abs_scaler_and_inverse  |  0.00134163 |   -31.5898 |        0.00196115 |        0.00115337 |
| apply_inplace_normalizer                  |  0.0113905  |    19.8152 |        0.00950672 |        0.00908794 |
| incremental_pca_split0                    |  7.8252     |   -12.3638 |        8.92918    |        8.98278    |
| convolution_batch_processing              |  0.00961189 |   139.238  |        0.00401772 |        0.00307064 |
| convolution_batch_processing_stride       |  0.0264587  |    14.7909 |        0.0230495  |        0.0219687  |
| convolution2d_batch_processing            |  0.00135345 |   -32.7674 |        0.00201309 |        0.00138206 |
| lasso_predict                             |  0.0043745  |    43.822  |        0.00304161 |        0.00279045 |
---

[Grafana Dashboard](https://930000e0-e69a-4939-912e-89a92316b420.ka.bw-cloud-instance.org)
Last updated: 2026-06-15T09:54:34Z

heat/core/linalg/basics.py
```diff
@@ -800,475 +798,90 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
+    else:  # both matrices are split in non-batch dims
         # get the lshape map to determine what needs to be sent where as well as M and N
         # lshape map dims -> {node, a=0 | b=1, lshape}
-        lshape_map = torch.zeros((comm.size, 2, ndim), dtype=int, device=tdev)
-        lshape_map[comm.rank, 0, :] = torch.tensor(a.lshape, device=tdev)
-        lshape_map[comm.rank, 1, :] = torch.tensor(b.lshape, device=tdev)
+        lshape_map = np.zeros((comm.size, 2, ndim), dtype=int)
+        lshape_map[comm.rank, 0, :] = a.lshape
+        lshape_map[comm.rank, 1, :] = b.lshape
         comm.Allreduce(MPI.IN_PLACE, lshape_map, MPI.SUM)
```
JuanPedroGHM: Can be made async, as lshape_map is used much later.

heat/core/linalg/basics.py
```diff
@@ -800,475 +798,90 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
+                c.larray += a_chunks[proc] @ b.larray[..., a_start:a_stop, :]
+                del a_chunks[proc]
 
-    if vector_flag:  # squeeze only in the la dimensions
+        # split dims 10
+        elif a.split == ndim - 1 and b.split == ndim - 2:
+            _b = b.copy()
+            _b.resplit_(a.split)
+            c[...] = matmul(a, _b)
+            del _b
```
JuanPedroGHM: Already looks much simpler than the previous version, and much more readable. 

I have two issues with the new form:

1. You split everything into two loops, first all the buffers are created and broadcasted, and then all the computation is performed. This has the major disadvantage that of high memory usage, as all the buffers are allocated at the same time, and ultimately uses more operations than a `.resplit_(None)`. This might ultimately limit the size of the matrixes that can be sent. One way to solve this is to simply merge the two loops, in a way that only two buffers are allocated, the one that is need for the current computation, and the one being waited on.

2. For split 10, you are respiting to 11, and then applying the full matmul function, which has some overhead. I would move these lines to the lines 825, and then directly apply the 11 split operation there.

JuanPedroGHM (changes requested):
I think the changes I suggested might lead to a smaller memory footprint, and should keep the performance identical. If it does not work, we can revert to the previous state and merge.

brownbaerchen:
@JuanPedroGHM, I had experimented with non-blocking sends in the loop because that would reduce complexity even more and, indeed, memory footprint, like you mentioned. Below is the result with non-blocking sends in a single loop, where `copilot` labeled data is with the first part of your suggestions and `JuanPedro` is with the memory-reducing changes you suggested:

matmul_scaling_JuanPedro

There is a meaningful slowdown when doing this. Let me try a hybrid with some prefetching.

brownbaerchen:
Here is the result with prefetching, which means the batch for the next computation iteration is already sent before the computation part in the current iteration. This is a compromise in that two memory buffers are allocated at any given time rather than one as you suggest, but this is much less than allocating all of them before doing computations. The code is slightly less clean, but I think it's fine. What do you think about this, @JuanPedroGHM?

matmul_scaling_prefetching

JuanPedroGHM:
Code wise, I think it looks great. We can discuss if we prioritize memory over runtime. It does not look amazing on this last plot, but I'm assuming it is still a significant speed-up when compared to main.

brownbaerchen:
> Code wise, I think it looks great. We can discuss if we prioritize memory over runtime. It does not look amazing on this last plot, but I'm assuming it is still a significant speed-up when compared to main.

Seems that prefetching was only slower for the largest value with split 00 than allocating all buffers beforehand. Therefore, I re-recorded this value and now it has the same performance. Here is the comparison with main for completeness:
matmul_scaling_prefetching

To me, this seems like a reasonable compromise between performance and memory consumption.

heat/core/linalg/basics.py
```diff
@@ -800,475 +798,97 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
...
+        lshape_map_req.Wait()
+        index_map_req.Wait()
+
+        # split dims 10 - This is the worst possible case and requires a resplit
+        if a.split == ndim - 1 and b.split == ndim - 2:
+            _b = b.copy()
+            _b.resplit_(a.split)
+            c[...] = matmul(a, _b)
+            del _b
+
```
JuanPedroGHM: Is it worth it to remove the `matmul` operation here? I would try to split this if statement from the ones below, do the resplit, and then hand it immediately to the other matmul cases below. It should at the very least save the communication cost of the lmap and index map, as they have been calculated already.

JuanPedroGHM (approved):
Looks good to me!

PR #2253, opened 2026-04-21
Return indices as ht.int64 in ht.unique(..., return_inverse=True)

brownbaerchen:
In the current main, the indices of unique elements are returned in the same data type as the input. This doesn't make sense and is not the behavior of numpy and torch. That is to say, in numpy and torch, we get:
```
>>> import numpy as np
>>> me = np.array([1., 2.])
>>> np.unique(me, return_inverse=True)
(array([1., 2.]), array([0, 1]))
```
meaning the index array is of integer type as appropriate for an index array, regardless of the type of the data.

This PR adapts the Heat API to the numpy one and returns the indices as integers by simply using the heat version of the torch type of each torch output.

The current behavior became an issue in #2199. There, I need to either force integer input or use any input type with this fix. I prefer this fix, but note that this needs to be merged before #2199.

### DIFF
--- a/heat/core/manipulations.py
+++ b/heat/core/manipulations.py
@@ -3375,7 +3375,13 @@ def unique(
         )
         if isinstance(torch_output, tuple):
             heat_output = tuple(
-                factories.array(i, dtype=a.dtype, split=None, device=a.device, comm=a.comm)
+                factories.array(
+                    i,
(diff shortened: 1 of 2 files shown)


ClaudiaComito (approved):
Absolutely, thanks @brownbaerchen

PR #2257, opened 2026-04-21
Expand indexing tests

brownbaerchen:
The advanced indexing that is developed in #938 needs to be merged! This is a preparatory PR that reduces the diff in that PR by merging some of tests that pass on the current main. This will hopefully speed up the review process a tiny bit.

PR #2259, opened 2026-04-22
Running ReceivePR and ci_full on changes to tests

brownbaerchen:
We noticed that PRs that change tests but no other functionality such as #2257 don't trigger the codebase CI or the full CI.
However, we want them to! We just forgot to add the respective triggers to the workflows when moving the tests to their own directory in #2172. This PR adds the missing path-based triggers.


ClaudiaComito (approved):

ClaudiaComito:
Thanks @brownbaerchen !

PR #2262, opened 2026-04-22
Expand `dev` optional dependencies to include everything required to run the entire test suite

brownbaerchen:
In this PR, all additional dependencies are added to the `dev` optional dependencies that are needed to run any tests.
For now this is just IO stuff, but in #2199, we will then add scikit-learn and use it as reference in the tests. This is actually the reason why we do this PR. Heat doesn't need scikit-learn because it should be used as a kind of replacement for it. So we need a streamlined set of optional dependencies for the tests. Since developers are supposed to run the tests, @JuanPedroGHM suggested to just put this in the `dev` optional dependencies.

This PR is only half of it. We still need to adapt the codebase CI. I am not sure on which branch exactly I should edit what exactly. Can you take care of it, @mtar?


mtar:
I updated the optional dependencies at codebase.

mtar (approved):

Issue #2272, opened 2026-05-04
[Bug]: IO tests failing due to newer zarr versions


marjin05:
### What happened?

The ci failed for my PR even though it did not touch any code

### Code snippet triggering the error

```shell
if ht.MPI_WORLD.rank == 0:
            root = zarr.open_group(self.ZARR_NESTED_PATH, mode="w")
            main_0 = root.create_group(nested_group_name)
>           main_0.create_dataset(
            ^^^^^^^^^^^^^^^^^^^^^
                array_name,
                shape=original_data.shape,
                dtype=original_data.dtype,
                data=original_data,
            )
E           AttributeError: 'Group' object has no attribute 'create_dataset'
```

### Error message or erroneous outcome

```shell
FAILED tests/core/test_io.py::TestIO::test_load_zarr_group - AttributeError: 'Group' object has no attribute 'create_dataset'
```

### Version

main (development branch)

### Python version

3.14

### PyTorch version

2.11

### MPI version

```shell

```

PR #2274, opened 2026-05-04
Restricting zarr version to <3.2

marjin05:
Restricted zarr Version to <3.2, temporary fix for #2272


marjin05:
Maybe merge without closing the issue because this is more of a quickfix

brownbaerchen (approved):

PR #2276, opened 2026-05-04
Fix for double counting error from #2093

brownbaerchen:
This PR fixes one of the issues in #2093, namely the bug with running code:
```python
import heat as ht

split = 0
shape = (4, 3)

A = ht.ones(shape, split=split)
B = ht.ones(shape[::-1], split=split)

C = A @ B
C_glob = (A.resplit(None) @ B.resplit(None)).resplit(C.split)

print(C)
print(C_glob)
```
which gives an output when run with two tasks:
```
DNDarray([[3., 3., 3., 3.],
          [6., 6., 6., 6.],
          [3., 3., 3., 3.],
          [6., 6., 6., 6.]], dtype=ht.float32, device=cpu:0, split=0)
DNDarray([[3., 3., 3., 3.],
          [3., 3., 3., 3.],
          [3., 3., 3., 3.],
          [3., 3., 3., 3.]], dtype=ht.float32, device=cpu:0, split=0)
```

This PR implements an analogous test with four tasks. I first pushed the test without the fix to record that the test fails on the current main. Then, I pushed the fix, which is just two small typos.

This does not close the original issue because it doesn't resolve all the issues, but I thought this is something we can merge quickly and independently of a larger refactor and fix for the other problems, which will come later.

### DIFF
--- a/heat/core/linalg/basics.py
+++ b/heat/core/linalg/basics.py
@@ -841,9 +841,9 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
 
         # check for remaining dims in the outside dimensions
         rem_a_out, rem_b_out = 0, 0
-        if a.lshape[-2] % mB != 0 or (kB == 1 and a.lshape[-2] != 1):
+        if a.lshape[-2] % mB != 0 or (mB == 1 and a.lshape[-2] != 1):
             rem_a_out = 1
-        if b.lshape[-1] % nB != 0 or (kB == 1 and b.lshape[-1] != 1):
+        if b.lshape[-1] % nB != 0 or (nB == 1 and b.lshape[-1] != 1):
             rem_b_out = 1
 
         # get the flags from all processes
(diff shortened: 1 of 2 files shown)


mtar:
Is the original matmul unit test faulty because it doesn't catch the bug? Does it only happen on smaller shapes?

mtar (commented):
The fix looks good and works as expected.

brownbaerchen:
> Is the original matmul unit test faulty because it doesn't catch the bug? Does it only happen on smaller shapes?

Yes, the tests in the current main are incomplete. Another failing edge case is pointed out in #2250. We should probably check the coverage report to make sure everything is covered.
Here, the test fails because the size of data is smaller than the number of ranks in one dimension. It's a bit subtle and niche so I understand that it wasn't considered when writing the tests.

mtar (dismissed):

brownbaerchen:
@mtar, I added a failing test for split 11 by simply transposing the matrices in the split 00 case.
I reverted the fix for split 11 when pushing the test to show that it actually fails. Then I added the fix back in and pushed this. So the latest version has two tests for split cases 00 and 11 which both fail on main and their respective fixes.

mtar (approved):

PR #2280, opened 2026-05-11
Bug report: Isolate edge case `heat.array([[5]])` failing on AMD

brownbaerchen:
This is a report of the bug in [this CI run](https://codebase.helmholtz.cloud/helmholtz-analytics/ci/-/jobs/3024934), where
```python
TypeError: invalid data of type <class 'list'>
```
is raised in `heat.factories.array` during the simple call `heat.array([[5]])`.
This appears to happen only on AMD GPUs. And can't be reproduced on CPUs or NVIDIA GPUs.

It's not marked as draft in order to trigger the codebase CI.

### DIFF
--- a/tests/core/test_factories.py
+++ b/tests/core/test_factories.py
@@ -400,6 +400,11 @@ class TestFactories(TestCase):
                 dim = self.get_rank() + 1
                 ht.array([[0] * dim] * dim, is_split=0)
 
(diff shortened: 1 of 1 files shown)


brownbaerchen:
Wait, this is kind of mysterious.. Why does this not fail by itself but it does fail in the convolution test??

PR #2288, opened 2026-05-13
Refactor matmul to avoid index tensors on GPU

brownbaerchen:
## Due Diligence

- General:
- Implementation:

## Description

GPUs are fast in computation, but not in memory allocation. Allocating small tensors for indices on GPU doesn't yield any performance benefits. In fact, quite the opposite: CPUs tend to be much better optimized for fine-grained memory access than GPUs.

Also, numpy allows for more flexible memory allocation than torch. Consider the following case:
```python
a = np.zeros(4)
a[:2] = (1, 1)
```
This works fast and easy with numpy. On the other hand, with we torch need to do:
```python
b = torch.zeros(4, device='cuda')
b[:2] = torch.tensor((1, 1), device=b.device)
```
that is to say, we need to cast to torch tensor before we can assign. This incurs overhead because we need to allocate a tiny torch tensor on GPU before the assignment which the GPU is not optimized for. Even when staying on CPU, we have some overhead from allocating the tiny torch tensor. 

This PR, therefore, changes the instances of things like the previous example in matmul to:
```python
c = np.zeros(4)
c[:2] = (1, 1)
c = torch.from_numpy(c)
```
Two things are different: We stay on CPU, and we use the more efficient assignment from Numpy. Also, I find it easier to read  what is assigned here.

In order to determine the performance impact, I recorded some weak scaling on 4 A100 on Juwels booster with all split combinations:

matmul_scaling_refactor

Solid lines are the current main and dashed lines are this branch. The black speedup line compares the split 00 combination.
For small matrix sizes, this has a large impact and yields a speedup of about 2, while for larger matrices this fine-grained optimization doesn't matter.
Performance-wise, this is not crucial to Heat because we are more focussed on weak scaling than strong scaling, but I think it's nice anyways.

## Changes proposed:

- Move tiny index tensors to CPU in matmul
- Use more efficient numpy assignment in matmul

## Type of change

Refactoring

## Memory requirements

Changes are only to small index arrays, shouldn't impact the overall memory footprint meaningfully

## Performance

Improved performance for small matrix sizes

### DIFF
--- a/heat/core/linalg/basics.py
+++ b/heat/core/linalg/basics.py
@@ -881,16 +881,14 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
         c_index_map_comm = comm.Iallreduce(MPI.IN_PLACE, c_index_map, MPI.SUM)
 
         if a.split == ndim - 2:
-            a_block_map = torch.zeros(
+            a_block_map = np.zeros(
                 (comm.size, a.shape[-2] // mB // comm.size, a.shape[-1] // kB, 2),
                 dtype=torch.int,
-                device=tdev,
             )
         elif a.split == ndim - 1:  # else should be equivalent at this point
-            a_block_map = torch.zeros(
+            a_block_map = np.zeros(
                 (comm.size, a.shape[-2] // mB, a.shape[-1] // kB // comm.size, 2),
                 dtype=torch.int,
-                device=tdev,
             )
         # units-> [process, dim0 block number, dim1 block number, start coord] **indices are local
 
@@ -902,7 +900,6 @@ def matmul(a: DNDarray, b: DNDarray, allow_resplit: bool = False) -> DNDarray:
             a_d1_1s_flag = True
 
         index_map_comm.Wait()
-        a_block_map = a_block_map.cpu().numpy()
         for pr in range(comm.size):
             start0 = index_map[pr, 0, 0, 0].item()
             stop0 = index_map[pr, 0, 0, 1].item()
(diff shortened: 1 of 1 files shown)


brownbaerchen:
After implementing the suggestions from copilot, performance is even better for small matrix sizes, though not by a huge amount:

matmul_scaling_avoid_tensors

mtar (commented):
Thank you, using the GPU for indexing was bothering me on some functions. I didn't know that numpy is quite more efficient than torch on the cpu. Does it make sense to switch back and forth on the map variables then or should we stick with numpy here?

brownbaerchen:
> Thank you, using the GPU for indexing was bothering me on some functions. I didn't know that numpy is quite more efficient than torch on the cpu. Does it make sense to switch back and forth on the map variables then or should we stick with numpy here?

I ran into issues with MPI communication with numpy arrays. That's why I cast to torch when the arrays are prepared. This casting is zero-copy and should be fine performance-wise. Though if we could do everything in numpy it would be cleaner and preferred. Do you know how to communicate numpy arrays in-place with Heat communicators?

brownbaerchen:
@ClaudiaComito, we were wondering in today's meeting how performance compares when staying on CPU with torch instead of switching to numpy for assignment and then back to torch for communication.
The resulting scaling is documented below:

matmul_scaling_tensors

I think these results are really curious. Apparently, there is slowdown for large matrices and similar speedup for small matrices.. No idea why, tbh. You can check out the code [here](https://github.com/brownbaerchen/heat/pull/new/matmul_refactor_keep_tensors) and tell me what's going on if you want. Or we just stick with numpy for performance and readability.

brownbaerchen:
Regarding the use of torch tensors on CPU instead of numpy arrays, it seems everything has gotten a bit slower somehow. I re-recorded the scaling on main and now there is no slowdown for large matrices, but the speedup for smaller matrices is less than with pure numpy:

matmul_scaling_tensors

Actually, it turns out the communication does work with numpy and I started skipping the conversion to tensor for some of the index array. In this case, there is even greater speedup of now 3 for small matrices.
I'll keep going with removing torch tensors here and then record some strong scaling plots to see the impact there.

matmul_scaling_avoid_tensors

brownbaerchen:
Turns out there is no strong scaling benefit here. The modifications are tiny fixed benefit on every GPU that only matters when communication and computation take negligibly long. However, when pushing scaling, communication will take longer than the tiny savings here.

For your reference, there is a plot on both CPU and GPU with split 00 with different number of tasks below.
The lack of scaling benefit is observed by the convergence of the solid and dashed lines, which are the different branches, with matrix size.

matmul_strong_scaling_avoid_tensors

I am done with this part of the refactor now and waiting for reviews.

PR #2294, opened 2026-05-18
Free MPI communicators in garbage collection

brownbaerchen:
MPI communicators should be [freed](https://github.com/helmholtz-analytics/heat/pull/new/free_communicators) after use in order to be cleaned up. Implementations such as MPICH have limits on the number of communicators that can be created simultaneously, see e.g. [here](https://lists.mcs.anl.gov/pipermail/mpich-discuss/2007-March/002000.html).

However, the wrapper around mpi4py communicators in Heat does not free the communicators after they are garbage collected. When using OpenMPI, this is not an issue, but with MPICH, we cannot run the CI because it creates too many communicators.
While this will not typically be a problem in practice, I think it's good practice to free the communicators in the wrapper destructor. That's all this PR is proposing.


JuanPedroGHM (dismissed):
Seems sensible, and generally good practice. I'll approve but I have some questions:

- Is the check for handle strictly necessary? All attributes should still be available while destroying an object.
- You mention that we create many communicators, but technically we should only have one instance of the actual mpi4py communicator, with a bunch of copies of the heat communication object using the same handle. This should not create any issues as far as I know. If it is the case, we should probably create an issue. (Unless there are special cases for the tests)

brownbaerchen:
Since we switched to MPICH and everything seems fine, let's leave this be for now.

JuanPedroGHM:
@brownbaerchen In the end is necessary. I added a check if MPI is still clear, and added the 4 rank tests to the PR UPDATE workflow to make sure MPICH gets tested with more than 1 rank.

brownbaerchen (commented):
LGTM. I can't approve, but I do ;) So please merge at will

JuanPedroGHM (approved):

Issue #1848, opened 2025-04-04
[Bug]: mpi4py/setup-mpi action causes failure in the CI


mrfh92:
### What happened?

When using the mpi4py/setup-mpi action, our CI fails. The exact cause is unclear so far. 
Happens in the ReceivePR and the CI-matrix workflows. 

see #1838 (which is closed to avoid unintended merging)

brownbaerchen:
Is this still relevant? The CI seems to run fine.

mrfh92:
Its one of our oldest and most mysterious issues.

mrfh92:
Its not on the "usual" CI on codebase, its about the CI-matrix (different PyTorch versions, Python versions, with and without hdf5 etc.) run on Github once you approve a PR. And there, I guess, it would still be failing if we would change to the newer version of the setup-mpi action.

PR #2295, opened 2026-05-18
Switch to MPICH in setup-MPI action

brownbaerchen:
As discussed in #1848, the CI fails when updating to the latest version of the setup-MPI action. We have no idea why. It seems that switching from OpenMPI to MPICH does the trick. (After freeing communicators, see #2294)
Do we have a reason to stick with OpenMPI? It must be possible to debug, but it's really difficult since none of us is having this issue locally with OpenMPI.. So, why not switch to MPICH?

### DIFF
--- a/.github/workflows/CIBase.yml
+++ b/.github/workflows/CIBase.yml
@@ -21,7 +21,7 @@ jobs:
           fi
 
       - name: Harden Runner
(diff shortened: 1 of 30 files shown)


JuanPedroGHM (approved):
If this works it would be awesome, both fixing the action and covering another MPI version with the tests.

brownbaerchen:
We already switched to MPICH in the CX reorganization

Issue #2273, opened 2026-05-04
[Bug]: Flaky `test_partial_h5_dataset`


JuanPedroGHM:
### What happened?

Sometimes CI tests fail on GPU because of these tests. But not always: here is error message.
```
=================================== FAILURES ===================================
__________________ TestPartialDataset.test_partial_h5_dataset __________________
tests/utils/data/test_partial_dataset.py:70: in test_partial_h5_dataset
    self.assertFalse(torch.allclose(last_batch, batch))
E   AssertionError: True is not false
=============================== warnings summary ===============================
```

### Code snippet triggering the error

```shell

```

### Error message or erroneous outcome

```shell

```

### Version

main (development branch)

### Python version

3.14

### PyTorch version

2.11

### MPI version

```shell

```

franko-c:
nice repro — i think both `last_batch = batch` assignments (lines 84 and 168) need a `.clone()`. without it, if the loader reuses the underlying tensor buffer, `last_batch` and `batch` point at the same memory and `allclose` is trivially true.

PR #2301, opened 2026-05-19
Force flaky test to retry.

JuanPedroGHM:
## Changes proposed:

More of a way of working change than a code change. I would say, if a test:

1. Fails only some of the times ( less than 50% )
2. Cannot be reproduced ( I'm looking at you #2273 )
3. And the whole team agrees

A test can be marked with the `@flaky` decorator. It will force the test to run multiple times, and we can set a minimum number of times when it needs to pass. See changes for example.

### DIFF
--- a/.github/workflows/ReceivePR.yml
+++ b/.github/workflows/ReceivePR.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Test
         run: |
           pip install .[dev]
-          pip install pytest
+          pip install pytest flaky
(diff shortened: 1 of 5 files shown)


brownbaerchen (dismissed):
YES! Thanks a lot!

brownbaerchen (dismissed):

brownbaerchen (dismissed):

brownbaerchen (approved):

PR #2304, opened 2026-05-20
Deleted unused `c_index_map` in `matmul`

brownbaerchen:
In the current main, an index array for the result matrix `c` in  `a @ b = c`, called `c_index_map`, is communicated on all tasks, yet never used. Deleting this will improve performance a tiny bit and declutter the code a little.


JuanPedroGHM (approved):
Good catch!

brownbaerchen:
Superseded by #2250

Issue #2327, opened 2026-05-27
[Bug]: `linalg.eigh` doesn't work for complex-valued matrices


GioPede:
### What happened?

`linalg.eigh` should work for Hermitian matrices (from which the h...). Currently all checks done in the function assert whether it is symmetric and real. By relaxing those checks from the few tests I've run everything seems to work just fine (torch.eigh does support complex matrices for example).

I have a small PR ready with the relevant changes

### Code snippet triggering the error

```shell
import heat as ht
mat = ht.eye(6, dtype=ht.complex128)
mat += ht.diag(ht.array([1,1,1,1,1]), offset=1) * 1j
mat += ht.diag(ht.array([1,1,1,1,1]), offset=-1) * -1j
S, V = ht.linalg.eigh(mat)

print(S)
```

### Error message or erroneous outcome

```shell
TypeError: Argument A needs to be a DNDarray with datatype float32 or float64, but data type is <class 'heat.core.types.complex128'>.
```

### Version

main (development branch)

### Python version

3.14

### PyTorch version

2.11

### MPI version

```shell

```

PR #2328, opened 2026-05-27
fix eigh to accept complex-valued DNDarrays

GioPede:
## Description
Small relaxation of the checks in `linalg.eigh` to allow complex-valued matrices

Issue/s resolved: #
#2327 

## Type of change

no


brownbaerchen:
@GioPede, thanks for getting this going! A function that implies Hermitian input in its name but doesn't allow it is indeed a bug! :D I started a [PR with some tests](https://github.com/GioPede/heat/pull/1) on your fork. Since you probably don't watch your own fork, I write a note about this here.

Also, just to make sure we have this discussion also here: This fix only works for non-distributed input, since the `_eigh` function for distributed input requires `percentile`, which requires `sort`, which doesn't allow complex input.

ClaudiaComito:
> @GioPede, thanks for getting this going! A function that implies Hermitian input in its name but doesn't allow it is indeed a bug! :D I started a [PR with some tests](https://github.com/GioPede/heat/pull/1) on your fork. Since you probably don't watch your own fork, I write a note about this here.
> 
> Also, just to make sure we have this discussion also here: This fix only works for non-distributed input, since the `_eigh` function for distributed input requires `percentile`, which requires `sort`, which doesn't allow complex input.

Addendum after discussing this in our meeting on Jun 8th.

`ht.sort` does not support complex dtypes because `torch.sort` doesn't. We could of course provide at least the CPU functionality via numpy.

brownbaerchen:
@Berkant03 is currently implementing vectorized sort, which is a kind of lexicographic sort in #2431. This should then allow to implement a complex sort.

brownbaerchen:
@Berkant03 is adding support for complex sorting in #2482, so maybe this will just work in the future.

brownbaerchen:
Sorry, I accidentally messed up merging main.

Anyways, even after #2482 is merged, this is still not working because it's relying on `polar`, which doesn't support complex matrices.

PR #2332, opened 2026-06-02
[BREAKING CHANGE] Adapt nonzero and where APIs to match NumPy

ClaudiaComito:
This PR makes `heat.nonzero` and `heat.where` behave like the Numpy counterparts.

This is an API-breaking, non-backwards-compatible change.

See this example for the effect of this PR:
```python
import heat as ht
import torch
import numpy as np

a = ht.arange(7)

print(torch.nonzero(a.larray > 3))
# tensor([[4],
#         [5],
#         [6]])

print(np.nonzero(a.numpy() > 3))
# (array([4, 5, 6]),)

print(ht.nonzero(a > 3))
# on main: DNDarray([4, 5, 6], dtype=ht.int64, device=cpu:0, split=None)
# on features/nonzero-updates: (DNDarray(MPI-rank: 0, Shape: (3,), Split: None, Local Shape: (3,), Device: cpu:0, Dtype: int64, Data:
#                                        [4, 5, 6]),)

# new on features/nonzero-updates:
print(ht.nonzero(a > 3, as_tuple=False))
# DNDarray([[4],
#           [5],
#           [6]], dtype=ht.int64, device=cpu:0, split=None)
```


heat/core/indexing.py
```diff
@@ -3,39 +3,39 @@
...
 
 
-def nonzero(x: DNDarray) -> DNDarray:
+def nonzero(x: DNDarray, as_tuple: bool = True) -> tuple[DNDarray, ...] | DNDarray:
     """
-    Return a :class:`~heat.core.dndarray.DNDarray` containing the indices of the elements that are non-zero (using ``torch.nonzero``).
-    If ``x`` is split then the result is split in the first dimension. However, this :class:`~heat.core.dndarray.DNDarray`
+    Return a Tuple of :class:`~heat.core.dndarray.DNDarray`s, one for each dimension of ``x``,
+    containing the indices of the non-zero elements in that dimension. If ``x`` is split then
+    the result is split in the 0th dimension. However, this :class:`~heat.core.dndarray.DNDarray`
```
brownbaerchen:
```suggestion
    the result is split in the first dimension. However, this :class:`~heat.core.dndarray.DNDarray`
```

heat/core/indexing.py
```diff
@@ -3,39 +3,39 @@
 """
 
 import torch
-from typing import List, Dict, Any, TypeVar, Union, Tuple, Sequence
 
 from .communication import MPI
 from .dndarray import DNDarray
-from . import sanitation
+from . import factories
```
brownbaerchen, line 9:
```suggestion
from . import factories
from . import sanitation
```
I think we should still use the sanitation module here

heat/core/indexing.py
```diff
@@ -48,48 +48,108 @@ def nonzero(x: DNDarray) -> DNDarray:
...
-    if x.split is not None:
-        _, _, slices = x.comm.chunk(x.shape, x.split)
-        lcl_nonzero[..., x.split] += slices[x.split].start
-
-    if x.ndim == 1:
-        lcl_nonzero = lcl_nonzero.squeeze(dim=1)
+    try:
+        local_x = x.larray
+    except AttributeError:
+        raise TypeError("Input must be a DNDarray, is {}".format(type(x)))
```
brownbaerchen:
```suggestion
    sanitation.santize_in(x)
    
    local_x = x.larray
```

heat/core/indexing.py
```diff
@@ -48,48 +48,108 @@ def nonzero(x: DNDarray) -> DNDarray:
...
+        # nonzero indices as tuple
+        nonzero = torch.nonzero(input=local_x, as_tuple=as_tuple)
+        # bookkeeping for final DNDarray construct
+        if as_tuple:
+            nonzero = list(nonzero)
+            for i, nz_tensor in enumerate(nonzero):
+                nonzero[i] = factories.array(nz_tensor, device=x.device, comm=x.comm)
+            return tuple(nonzero)
+        # nonzero indices as single 2D DNDarray
+        return factories.array(nonzero, device=x.device, comm=x.comm)
```
brownbaerchen:
```suggestion
        else:
            # nonzero indices as single 2D DNDarray
            return factories.array(nonzero, device=x.device, comm=x.comm)
```
This is a bit easier to read in my opinion

heat/core/indexing.py
```diff
@@ -48,48 +48,108 @@ def nonzero(x: DNDarray) -> DNDarray:
...
+
+    # distributed case
+    lcl_nonzero = torch.nonzero(input=local_x, as_tuple=False)
+    nonzero_size = torch.tensor(lcl_nonzero.shape[0], dtype=torch.int64, device=lcl_nonzero.device)
```
brownbaerchen:
```suggestion
    nonzero_size = torch.tensor(lcl_nonzero.shape[0], dtype=torch.int64)
```
No need to allocate this on GPU

heat/core/indexing.py
```diff
@@ -48,48 +48,108 @@ def nonzero(x: DNDarray) -> DNDarray:
...
+            gshape=shape_2d,
+            dtype=nonzero_dtype,
+            split=0,
+            device=x.device,
+            comm=x.comm,
+            balanced=False,
+        )
+        # vectorized sorting of nz indices along axis 0
+        global_nonzero.balance_()
+        global_nonzero = manipulations.unique(global_nonzero, axis=0)
```
brownbaerchen: Why is this needed? Seems like duplicate entries would be a bug at this point. Or are there some side effects of `unique` here?

tests/core/test_indexing.py
```diff
@@ -9,40 +10,56 @@ def test_nonzero(self):
         a = ht.array([[1, 2, 3], [4, 5, 2], [7, 8, 9]], split=None)
         cond = a > 3
         nz = ht.nonzero(cond)
-        self.assertEqual(nz.gshape, (5, 2))
-        self.assertEqual(nz.dtype, ht.int64)
-        self.assertEqual(nz.split, None)
+        self.assertEqual(len(nz), 2)
+        self.assertEqual(len(nz[0]), 5)
+        self.assertEqual(nz[0].dtype, ht.int64)
```
brownbaerchen: I think this test could be more comprehensive with a comparison to numpy and torch. Maybe compare to numpy if `as_tuple` is false and to torch otherwise?

tests/core/test_indexing.py
```diff
@@ -9,40 +10,56 @@ def test_nonzero(self):
...
         a = ht.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], split=None)
         cond = a > 3
         wh = ht.where(cond)
-        self.assertEqual(wh.gshape, (6, 2))
-        self.assertEqual(wh.dtype, ht.int64)
-        self.assertEqual(wh.split, None)
+        self.assertEqual(len(wh), 2)
+        self.assertEqual(wh[0].gshape[0], 6)
+        self.assertEqual(wh[0].dtype, ht.int64)
+        self.assertEqual(wh[0].split, None)
```
brownbaerchen: Maybe this could be compared to numpy and torch too?

brownbaerchen (changes requested):

brownbaerchen:
The issues arise because the API of nonzero is totally changed in this PR. See the following example:

```python
import heat as ht
import torch
import numpy as np

a = ht.arange(7)

print(torch.nonzero(a.larray > 3))
# tensor([[4],
#         [5],
#         [6]])

print(np.nonzero(a.numpy() > 3))
# (array([4, 5, 6]),)

print(ht.nonzero(a > 3))
# on main: DNDarray([4, 5, 6], dtype=ht.int64, device=cpu:0, split=None)
# on features/nonzero-updates: (DNDarray(MPI-rank: 0, Shape: (3,), Split: None, Local Shape: (3,), Device: cpu:0, Dtype: int64, Data:
#                                        [4, 5, 6]),)

# new on features/nonzero-updates:
print(ht.nonzero(a > 3, as_tuple=False))
# DNDarray([[4],
#           [5],
#           [6]], dtype=ht.int64, device=cpu:0, split=None)
```
That is to say the previous API was neither numpy, nor torch and this PR changes it to default to numpy but with the option to get the torch behavior.

This is a good thing in principle, but I don't like to silently change the API. So, yes it's annoying that the current main relies on the previous API in a bunch of places and doesn't work with only these changes, but also we should clearly indicate this as a breaking, API-changing change.

We have two options:
 - Adapt all uses of nonzero in the current main to use the new and improved API, then merge this PR
 - Don't merge this PR and merge the changes together with the huge advanced indexing PR #938

Either way, we need to mention the API changes in the release notes. What do you think, @JuanPedroGHM, @mtar, @ClaudiaComito?

heat/core/indexing.py
```diff
@@ -48,48 +49,106 @@ def nonzero(x: DNDarray) -> DNDarray:
...
 
 def where(
     cond: DNDarray,
-    x: Union[None, int, float, DNDarray] = None,
-    y: Union[None, int, float, DNDarray] = None,
+    x: None | int | float | DNDarray = None,
+    y: None | int | float | DNDarray = None,
 ) -> DNDarray:
     """
     Return a :class:`~heat.core.dndarray.DNDarray` containing elements chosen from ``x`` or ``y`` depending on condition.
```
brownbaerchen, line 152: The API has changed in this PR and the docstring needs to be adapted

heat/core/indexing.py
```diff
@@ -128,20 +187,52 @@ def where(
...
+        # 1) Non-distributed: behave exactly like ht.nonzero(cond)
+        if cond.split is None:
+            return nz
+
+        # 2) Distributed along axis 0: keep the legacy tuple-of-indices API.
+        #    This is relied upon in several parts of the code base (e.g. KMeans).
+        if cond.split == 0:
+            return nz
+
+        # 3) Distributed along a non-zero axis (split > 0)
```
brownbaerchen: @ClaudiaComito, what is different about split > 0 that it warrants this special case? To me, this is a very confusing API that, if we go for this, we have to justify and document.

brownbaerchen: I just removed this because it didn't seem necessary. The tests in #938 also pass without this.

heat/core/indexing.py
brownbaerchen: I don't like that this is needed because it's not clean, but the tests don't pass without, so..

brownbaerchen (dismissed):

heat/core/indexing.py
```diff
@@ -48,48 +49,104 @@ def nonzero(x: DNDarray) -> DNDarray:
...
-    sanitation.sanitize_in(x)
-
+    sanitize_in(x)
+
+    if not x.is_distributed():
+        # nonzero indices as tuple
+        nonzero = torch.nonzero(input=x.larray, as_tuple=as_tuple)
+        # bookkeeping for final DNDarray construct
+        if as_tuple:
+            nonzero = list(nonzero)
```
mtar, line 64:
```suggestion
            nonzero = nonzero.tolist()
```

brownbaerchen:
```python
>>> (1, 2, 3).tolist()
  File "<python-input-0>", line 1, in <module>
    (1, 2, 3).tolist()
    ^^^^^^^^^^^^^^^^
AttributeError: 'tuple' object has no attribute 'tolist'
```
?

brownbaerchen: Oh, I see. Heat DNDarray has `.tolist`. However, the type of `nonzero` here is `tuple`, not `DNDarray` and we need to use `list(...)`.

mtar: Ah sorry, it came automatically when I saw what seems to be a torch tensor at first glance 😅

heat/core/indexing.py
```diff
@@ -114,34 +171,47 @@ def where(
...
+    elif x is None and y is None:  # delegate to nonzero(cond)
+        return nonzero(cond)  # tuple of DNDarrays, one per dimension
+
+    # ---- invalid combinations ----------------------------------------------
     else:
         raise TypeError(
-            f"either both or neither x and y must be given and both must be DNDarrays or numerical scalars({type(x)}, {type(y)})"
+            "either both or neither x and y must be given and both must be "
+            f"DNDarrays or numerical scalars (got {type(x)}, {type(y)})"
         )
```
mtar, line 212: The comments are a little verbose.

tests/core/test_indexing.py
```diff
@@ -1,84 +1,149 @@
...
     def test_nonzero(self):
-        # cases to test:
-        # not split
-        a = ht.array([[1, 2, 3], [4, 5, 2], [7, 8, 9]], split=None)
-        cond = a > 3
-        nz = ht.nonzero(cond)
-        self.assertEqual(nz.gshape, (5, 2))
-        self.assertEqual(nz.dtype, ht.int64)
-        self.assertEqual(nz.split, None)
+        for split in [None, 0, 1]:
```
mtar: Can we parameterize the tests now that we are more using pytest?

heat/core/indexing.py
```diff
@@ -114,34 +171,47 @@ def where(
...
+    # ---- binary where(cond, x, y) branch ------------------------------------
     if cond.split is not None and (isinstance(x, DNDarray) or isinstance(y, DNDarray)):
         if (isinstance(x, DNDarray) and cond.split != x.split) or (
             isinstance(y, DNDarray) and cond.split != y.split
         ):
-            if len(y.shape) >= 1 and y.shape[0] > 1:
+            # Only raise if the "other" array has a meaningful first dimension.
+            if isinstance(y, DNDarray) and len(y.shape) >= 1 and y.shape[0] > 1:
                 raise NotImplementedError("binary op not implemented for different split axes")
+
```
mtar: I think the if conditions can be reduced, like putting the innermost condition to the top reduces the repeated isinstance calls for example.

```suggestion
    if cond.split is not None and isinstance(y, DNDarray) and len(y.shape) >= 1 and y.shape[0] > 1:
        if (isinstance(x, DNDarray) and cond.split != x.split) or cond.split != y.split):
            raise NotImplementedError("binary op not implemented for different split axes")

```

tests/core/test_indexing.py
```diff
@@ -1,84 +1,149 @@
...
+                self.assertEqual(len(ht_res), len(np_res))
+                for _ht_res, _np_res in zip(ht_res, np_res):
+                    self.assertTrue(ht.equal(_ht_res, ht.array(_np_res)))
+
+                self.assertEqual(ht_res[0].dtype, ht.int64)
+
+            else:
+                self.assertTrue(np.allclose(ht_res.shape, np_res.shape))
+                self.assertTrue(ht.equal(ht_res, ht.array(np_res)))
+                self.assertEqual(ht_res[0].dtype, ht.types.canonical_heat_type(np_res.dtype))
```
mtar: Nice. Such helper functions would be good for some other tests as well.

mtar (commented):
Everything works as expected. I only have some small comments.

heat/core/indexing.py
JuanPedroGHM: @brownbaerchen @ClaudiaComito I think this should be addressed before we proceed with the merge. Or has there been any updates. I would wait until this is resolved first before any merge.

ClaudiaComito: I use `unique()` here because I need the global vectorized sorting

JuanPedroGHM (commented):
In like the changes, but I would not merge this until we find the reason why the global nonzero tensor needs to be pruned of duplicates. 

Is this somehow addressed in the other PR's that affect this two functions?

heat/core/indexing.py
ClaudiaComito: @JuanPedroGHM to me this is resolved. The indices need to be sorted by coordinate axis. i.e.

```
[[0, 0, 1],
 [0, 1, 3],
 [0, 1, 5],
 [1, 0, 0],
 [2, 0, 1]]
```

`unique(axis)` does exactly that. I don't know what it means that it's not clean.

ClaudiaComito: Expanding with more details. Suppose the global key is 

```
key = ht.array([[3, 3, 3],
       [5, 4, 9],
       [5, 6, 0],
       [6, 3, 8],
       [1, 6, 6]], split=0)
```       


We need to align the key with the actual local positions of the elements of the indexed array.  The first step is the vectorized sorting (please if there is a more appropriate term for this let me know). I.e. we want  to sort `key` by the rows:

```
key = ht.array([[1, 6, 6],
       [3, 3, 3],
       [5, 4, 9],
       [5, 6, 0],
       [6, 3, 8]], split=0)
```

`unique(axis=0)` sorts by the rows keeping elements of the same original row together.   

`sort(axis)` will sort the elements within each axis (row or column in this case) so the connection between elements of the same row will be lost.

I agree `unique` is overkill, but I'm pretty sure Heat doesn't have a dedicated vectorized sorting functionality that works in distributed mode (we have an old issue that we could revive though #363 )



```
>>>  np.sort(a, axis=0)
array([[1, 3, 0],
       [3, 3, 3],
       [5, 4, 6],
       [5, 6, 8],
       [6, 6, 9]])

>>> np.sort(a, axis=1)
array([[3, 3, 3],
       [4, 5, 9],
       [0, 5, 6],
       [3, 6, 8],
       [1, 6, 6]])

```

brownbaerchen: What's not clean about this is:
 - This is using a non-obvious side effect of unique
 - This is wasting compute eliminating non-existent non-unique values

What we need is essentially:
```python
>>> import numpy as np
>>> a = np.array([[0, 0, 1],
...  [0, 1, 3],
...  [0, 1, 5],
...  [1, 0, 0],
...  [2, 0, 1]])
>>> a[np.argsort(a[:, 0])]
array([[0, 0, 1],
       [0, 1, 3],
       [0, 1, 5],
       [1, 0, 0],
       [2, 0, 1]])
>>> np.unique(a, axis=0)
array([[0, 0, 1],
       [0, 1, 3],
       [0, 1, 5],
       [1, 0, 0],
       [2, 0, 1]])
```
Using `argsort` is clean because the name of the function implies what it's doing and all that it's doing. The reader need not wonder why this function is called because it's clear what it does.

Problem: Heat doesn't have argsort. It has `sort` which returns the indices, so we could do
```python
>>> import heat as ht
>>> b = ht.array(a)
>>> b[ht.sort(b[:, 0])[1]]
DNDarray(MPI-rank: 0, Shape: (5, 3), Split: None, Local Shape: (5, 3), Device: cpu:0, Dtype: int64, Data:
         [[0, 0, 1],
          [0, 1, 3],
          [0, 1, 5],
          [1, 0, 0],
          [2, 0, 1]])
```
But doesn't work in parallel..

What to do? :D

brownbaerchen (approved):

PR #2335, opened 2026-06-03
First small cleanup

brownbaerchen:
This is a tiny cleanup with some `distr_mask_fast_path` stuff. It doesn't get rid of it, just of a section that cannot be reached.


ClaudiaComito (approved):
Looks good, thanks

PR #2350, opened 2026-06-09
early out for non distributed arrays

### LINKED ISSUE #2326: [Commmunication] Early out for non-distributed arrays
Heat's MPI commmunication functions also run on non-distributed arrays. This may lead to some unnecessary buffer creations.

mtar:
## Due Diligence

- General:
- Implementation:

## Description

The collective MPI communication functions check if the communicator size is one and copies the send buffer to the receive buffer if available. Additionally, this PR removes direct access to mpi variables outside of the communication module.

Issue/s resolved: #2326 

## Changes proposed:

- access MPI variables through own communicator, not directly
- make general communicator abstract
- add early out in collective mpi communication functions
- use newer type hints
- remove unused imports

## Type of change

Refactor

yes / no


heat/core/communication.py
```diff
@@ -73,36 +79,50 @@ def __getattr__(self, name: str) -> Callable:
...
-        raise NotImplementedError()
-
-    def chunk(self, shape, split) -> NotImplementedError:
+    @abstractmethod
+    def chunk(
+        self,
+        shape: tuple[int, ...],
+        split: int,
+        rank: int = None,
+        w_size: int = None,
```
brownbaerchen: I don't know what the type hint policy is but the default should probably be of the allowed type? So either: No default, default int, or type hint `int | None`?

heat/core/communication.py
```diff
@@ -73,36 +79,50 @@ def __getattr__(self, name: str) -> Callable:
...
         Parameters
         ----------
-        shape : Tuple[int,...]
+        shape : tuple[int,...]
             The global shape of the data to be split
         split : int
             The axis along which to chunk the data
-
+        rank : int, optional
+            Process for which the chunking is calculated for, defaults to ``self.rank``.
```
brownbaerchen:
```suggestion
            Process for which the chunking is calculated for.
```
We can't tell what it defaults to here, right?

heat/core/communication.py
```diff
@@ -73,36 +79,50 @@ def __getattr__(self, name: str) -> Callable:
...
+            Intended for creating chunk maps without communication
+        w_size : int, optional
+            The MPI world size, defaults to ``self.size``.
```
brownbaerchen:
```suggestion
            The MPI world size.
```

heat/core/communication.py
```diff
@@ -968,22 +984,23 @@ def op(sendbuf: MPI.memory, recvbuf: MPI.memory, datatype):
...
 
         Parameters
         ----------
         func: Callable
             The respective MPI reduction operation
-        sendbuf: Union[DNDarray, torch.Tensor, Any]
-            Buffer address of the send message
-        recvbuf: Union[DNDarray, torch.Tensor, Any]
+        sendbuf: DNDarray or torch.Tensor or np.ndarray
+            Buffer address of the send message. If MPI.IN_PLACE is set,
```
brownbaerchen:
```suggestion
            Buffer for the send message. If MPI.IN_PLACE is set,
```

heat/core/communication.py
```diff
@@ -968,22 +984,23 @@ def op(sendbuf: MPI.memory, recvbuf: MPI.memory, datatype):
...
+            recvbuf is also used as send buffer.
+        recvbuf: DNDarray or torch.Tensor or np.ndarray
             Buffer address where to store the result of the reduction
```
brownbaerchen:
```suggestion
            Buffer for storing the result of the reduction
```

heat/core/communication.py
```diff
@@ -1003,6 +1020,16 @@ def __reduce_like(
         if isinstance(recvbuf, DNDarray):
             recvbuf = recvbuf.larray
 
+        if not self.is_distributed():
+            if sendbuf is not MPI.IN_PLACE:
+                # set the contiguousness like in multiprocess
+                sendbuf.set_(sendbuf.contiguous())
```
brownbaerchen: Why is this needed?

heat/core/communication.py
```diff
@@ -1509,6 +1552,16 @@ def __alltoall_like(
         if not isinstance(recvbuf, torch.Tensor) and send_axis != 0:
             raise TypeError(f"recvbuf of type {type(recvbuf)} does not support send_axis != 0")
 
+        if not self.is_distributed():
+            if recv_axis > 1 and send_axis > 1 and recv_axis == send_axis:
```
brownbaerchen: Why check if the axes are >1?

heat/core/communication.py
```diff
@@ -2213,6 +2282,11 @@ def __scatter_like(
         if not isinstance(recvbuf, torch.Tensor) and recv_axis != 0:
             raise TypeError(f"recvbuf of type {type(recvbuf)} does not support recv_axis != 0")
 
+        if not self.is_distributed():
+            if sendbuf is not MPI.IN_PLACE:
```
brownbaerchen: The type hints are a bit confusing here, maybe because it is unexpected that `MPI.IN_PLACE` is of type `Array`. How about renaming the `Array` type to `BufferType`?

brownbaerchen (changes requested):
Thanks for the large cleanup! I have only a few minor comments.

heat/core/communication.py
mtar: That slipped through. Thanks for spotting that.

heat/core/communication.py
mtar: Yes, not at this point. It was copy-pasted from MPICommunication with its additional parameters that some callers passes arguments to.

heat/core/communication.py
mtar: Do you mean making the send buffer contiguous? Our unit test checks for that and it follows the output of multiprocess one. However, it could be an oversight.
https://github.com/helmholtz-analytics/heat/blob/8cb76f3c585800e1c731abc4791d64c959d0d1f9/tests/core/test_communication.py#L2177

heat/core/communication.py
mtar: It is not not needed for a single process, but I wanted it to behave like multi-process again.

heat/core/communication.py
mtar: Good idea! That is a better name.

heat/core/communication.py
brownbaerchen: Two issues: 
 - The type hints say `Array`, but this will not accept numpy arrays because numpy doesn't have `set_`.
 - The reduce operation is not actually performed, right?

heat/core/communication.py
brownbaerchen: How does multi-process relate to changing the split?

heat/core/communication.py
mtar: I moved the if block a little down so I don't have to deal with the contiguousness.

heat/core/communication.py
mtar: My intention was to have the lower level functions to be consistent on all processes numbers for like testing or debugging. I will have to do some change in the unit test otherwise on of the ones expected to fail.

brownbaerchen: I just don't understand why this particular error would be raised in this particular situation. I would prefer either adapting the tests or making more clear why we need to raise this error only when the axes are > 1 in the error message.

heat/core/communication.py
```diff
@@ -968,23 +984,24 @@ def op(sendbuf: MPI.memory, recvbuf: MPI.memory, datatype):
...
 
         Parameters
         ----------
         func: Callable
             The respective MPI reduction operation
-        sendbuf: Union[DNDarray, torch.Tensor, Any]
-            Buffer address of the send message
-        recvbuf: Union[DNDarray, torch.Tensor, Any]
-            Buffer address where to store the result of the reduction
+        sendbuf: DNDarray or torch.Tensor or np.ndarray
```
brownbaerchen: Actually, there is a `sendbuf.numel()` call in line 1028 which still means this function doesn't support numpy arrays in some cases. This `numel` / `size` difference is such a mess...

mtar: Yeah, I'm not sure if all functions really supports np.ndarray properly and they need a closer look, but the new types hints are finer than `Any` :smile:

brownbaerchen: Yes, type hint `Any` is arguably not very useful :D But maybe we remove `np.ndarray` if that would be a lie? That's why I personally don't like type hints: They are not code and anything that is not code is probably a lie :D

heat/core/communication.py
mtar: I changed it into a UserWarning mentioning that it is not supported in multiprocess. Most communication functions were not touched, only patched,  and might need some overhaul.

heat/core/communication.py
```diff
@@ -996,17 +1013,19 @@ def __reduce_like(
...
         buf = None
-        # unpack the send buffer if it is a HeAT tensor
+        # unpack the send buffer if it is a Heat tensor
         if isinstance(sendbuf, DNDarray):
             sendbuf = sendbuf.larray
-        # unpack the receive buffer if it is a HeAT tensor
+        # unpack the receive buffer if it is a Heat tensor
         if isinstance(recvbuf, DNDarray):
             recvbuf = recvbuf.larray
+        if not isinstance(sendbuf, torch.Tensor) and isinstance(recvbuf, torch.Tensor):
```
mtar: I added another case that  falls back to the mpi4py version if we don't got any heat or torch tensors for better readability.

mtar: Removing it. Out of scope for this PR.

heat/core/communication.py
mtar: I reverted the type hints back as it is not part of the PR and only left simpler updates. It makes no sense to have an union with Any, so I left Any as the only type whenever it is included. Actual changes go to another PR.

heat/core/communication.py
```diff
@@ -2478,5 +2550,17 @@ def use_comm(comm: Communication = None):
...
 
+def _copyto(sendbuf, recvbuf):
+    if isinstance(recvbuf, torch.Tensor):
+        if isinstance(sendbuf, np.ndarray):
+            sendbuf = torch.from_numpy(sendbuf)
+        recvbuf.copy_(sendbuf)
+    elif isinstance(recvbuf, np.ndarray):
+        if isinstance(sendbuf, torch.Tensor):
+            sendbuf = sendbuf.cpu().numpy()
+        np.copyto(sendbuf, recvbuf)
```
brownbaerchen, line 2561:
```suggestion
        np.copyto(sendbuf, recvbuf)
    else:
        raise NotImplementedError(f'Don\'t know how to copy {type(sendbuf)} to {type(recvbuf)}')
```

brownbaerchen (changes requested):
Looks good! I'd only like to raise an error in `_copyto` if nothing is copied. Just in case someone tries to put a `DNDarray` in there or whatever. Otherwise, I think is good to go.

brownbaerchen (approved):

PR #2359, opened 2026-06-16
Changes from #938 regarding batch-parallel clustering

brownbaerchen:
These are some changes regarding batch-parallel clustering from #938. I have no idea what they do, but they don't belong in #938, so they get split off here.


mtar (approved):
The changes seem handle special values like nan and inf and where all weights or dists are zero. It looks fine to merge.

PR #2360, opened 2026-06-16
Refactoring: Updating some type hints in dndarray.py

brownbaerchen:
These are a few changes from #938 that are not related to advanced indexing, but are simply small refactorings. They shouldn't change the behavior meaningfully.


heat/core/dndarray.py
```diff
@@ -77,10 +77,10 @@ def __init__(
         self.__split = split
         self.__device = device
         self.__comm = comm
-        self.__balanced = balanced
+        self.__balanced: bool = balanced
         self.__ishalo = False
-        self.__halo_next = None
-        self.__halo_prev = None
+        self.__halo_next: torch.Tensor | None = None
+        self.__halo_prev: torch.Tensor | None = None
```
mtar: Is there a reason why exactly these fields have a type hint?

heat/core/dndarray.py
```diff
@@ -10,7 +10,7 @@
 from inspect import stack
 from mpi4py import MPI
 from pathlib import Path
-from typing import List, Union, Tuple, TypeVar, Optional
+from typing import Union, TypeVar
```
mtar, line 13: You already replaced the union type with the newer `|` operator mostly.

```suggestion
from typing import TypeVar
```

heat/core/dndarray.py
```diff
@@ -263,7 +263,7 @@ def lnumel(self) -> int:
         return np.prod(self.__array.shape)
 
     @property
-    def lloc(self) -> Union[DNDarray, None]:
+    def lloc(self) -> "DNDarray" | None:
```
mtar:
```suggestion
    def lloc(self) -> DNDarray | None:
```

heat/core/dndarray.py
```diff
@@ -623,7 +623,7 @@ def __cast(self, cast_function) -> Union[float, int]:
 
         raise TypeError("only size-1 arrays can be converted to Python scalars")
 
-    def collect_(self, target_rank: Optional[int] = 0) -> None:
+    def collect_(self, target_rank: int | None = 0) -> None:
```
mtar:
```suggestion
    def collect_(self, target_rank: int = 0) -> None:
```

heat/core/dndarray.py
```diff
@@ -676,7 +676,7 @@ def __complex__(self) -> DNDarray:
         """
         return self.__cast(complex)
 
-    def counts_displs(self) -> Tuple[Tuple[int], Tuple[int]]:
+    def counts_displs(self) -> tuple[tuple[int], tuple[int]]:
```
mtar:
```suggestion
    def counts_displs(self) -> tuple[tuple[int, ...], tuple[int, ...]]:
```

heat/core/dndarray.py
```diff
@@ -878,7 +879,7 @@ def fill_diagonal(self, value: float) -> DNDarray:
 
         return self
 
-    def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDarray:
+    def __getitem__(self, key: Union[int, tuple[int, ...], list[int, ...]]) -> DNDarray:
```
mtar:
```suggestion
    def __getitem__(self, key: int | slice[int | None] | tuple[int, ...] | list[int]) -> DNDarray:
```

heat/core/dndarray.py
```diff
@@ -1250,7 +1255,7 @@ def __repr__(self) -> str:
         """
         return printing.__repr__(self)
 
-    def ravel(self):
+    def ravel(self) -> "DNDarray":
```
mtar:
```suggestion
    def ravel(self) -> DNDarray:
```

heat/core/dndarray.py
```diff
@@ -888,7 +889,7 @@ def __getitem__(self, key: Union[int, Tuple[int, ...], List[int, ...]]) -> DNDar
 
         Parameters
         ----------
-        key : int, slice, Tuple[int,...], List[int,...]
+        key : int, slice, tuple[int,...], list[int,...]
```
mtar:
```suggestion
        key : int or slice or tuple[int,...] or list[int]
```

heat/core/dndarray.py
```diff
@@ -1569,15 +1574,15 @@ def resplit_(self, axis: int = None):
 
     def __setitem__(
         self,
-        key: Union[int, Tuple[int, ...], List[int, ...]],
+        key: Union[int, tuple[int, ...], list[int, ...]],
```
mtar:
```suggestion
        key: int | tuple[int, ...] | list[int],
```

heat/core/dndarray.py
```diff
@@ -1569,15 +1574,15 @@ def resplit_(self, axis: int = None):
...
         value: Union[float, DNDarray, torch.Tensor],
     ):
         """
         Global item setter
 
         Parameters
         ----------
-        key : Union[int, Tuple[int,...], List[int,...]]
+        key : Union[int, tuple[int,...], list[int,...]]
```
mtar:
```suggestion
        key : int or tuple[int,...] or list[int]
```

heat/core/dndarray.py
```diff
@@ -1858,7 +1863,7 @@ def __setitem__(
 
     def __setter(
         self,
-        key: Union[int, Tuple[int, ...], List[int, ...]],
+        key: Union[int, tuple[int, ...], list[int, ...]],
```
mtar:
```suggestion
        key: int | tuple[int, ...] | list[int],
```

heat/core/dndarray.py
```diff
@@ -1890,7 +1895,7 @@ def __str__(self) -> str:
         """
         return printing.__str__(self)
 
-    def tolist(self, keepsplit: bool = False) -> List:
+    def tolist(self, keepsplit: bool = False) -> list:
```
mtar:
```suggestion
    def tolist(self, keepsplit: bool = False) -> list[int | float]:
```

tests/cluster/test_kmedians.py
```diff
@@ -41,7 +41,7 @@ def test_fit_iris_unsplit(self):
 
         # fit the clusters
         k = 3
-        kmedian = ht.cluster.KMedians(n_clusters=k)
+        kmedian = ht.cluster.KMedians(n_clusters=k, random_state=1)
```
mtar: Why do you pass an argument for random_state?

mtar (changes requested):
We don't have a consistent style for the type notation in the docstrings. My suggestions are from one style that we have throughout the library, but we can discuss which one we want to use.

heat/core/dndarray.py
brownbaerchen: I just copied this from #938, I guess because this doesn't have a non `None` default value..?

tests/cluster/test_kmedians.py
brownbaerchen: This is a random change from #938. I can remove it from here and there, if you want..

heat/core/dndarray.py
mtar: I was bewildered, because the added type hints looks so randomly added. Some parameters got one and some not. It should be consistent.

brownbaerchen:
@mtar, I have removed the type hints in the `__init__` method and the random seed from this PR. You can decide what to do with that in #938 directly.
Note that the tests require #2378...

mtar (approved):

PR #2369, opened 2026-07-07
Cast numpy to torch in `_resolve_indexing_state`

brownbaerchen:
Part one of unifying the treatment of different types of array keys.
Part two will be casting torch tensors to heat dndarray.

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -287,8 +287,11 @@ def _resolve_indexing_state(
             backwards_transpose_axes=tuple(range(arr.ndim)),
         )
 
-    # evaluate if this is a distributed fast-path mask before we modify the key
+    # cast any numpy keys to torch tensor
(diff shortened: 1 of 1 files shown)


heat/core/dndarray.py
mtar, line 419: Is this also numpy specific?

mtar (approved):
looks good to me

heat/core/dndarray.py
brownbaerchen: No. A type error is raised in the case of DNDarray
```python
>>> import heat as ht
>>> ht.ones(1).nonzero(as_tuple=False)
  File "<python-input-1>", line 1, in <module>
    ht.ones(1).nonzero(as_tuple=False)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
TypeError: <lambda>() got an unexpected keyword argument 'as_tuple'
```

PR #2370, opened 2026-07-07
Small refactor of the use of `nonzero` in `_resolve_indexing_state`

brownbaerchen:
This is just a small simplification if I understand correctly.


brownbaerchen:
Let's do a different simplification that does not use `flatten`.

PR #2371, opened 2026-07-07
Skip key normalization in `_resolve_indexing_state`

brownbaerchen:
The tests pass (locally) without this, so maybe we can remove this.
Not casting `DNDarray` to `Tensor` will be helpful in casting `Tensor` to `DNDarray`...

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -332,18 +332,6 @@ def _resolve_indexing_state(
                     backwards_transpose_axes=tuple(range(arr.ndim)),
                 )
 
-    # normalize index components
-    if isinstance(key, DNDarray):
-        if key.dtype not in (ht_bool, ht_uint8) and key.split is None:
-            key = key.larray.to(torch.int64)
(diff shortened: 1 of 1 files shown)


mtar:
The CI tests end on multiprocess with GPU support with a device mismatch. Was it already the case before the changes? Would it be better to move the input array to the CPU first?

brownbaerchen:
> The CI tests end on multiprocess with GPU support with a device mismatch. Was it already the case before the changes? Would it be better to move the input array to the CPU first?

No, the tests pass without these changes. It's very confusing that not casting int32 to int64 means some array is apparently not transferred to GPU.
Anyways, since I have no idea which side effect is responsible for this, I would be fine with just closing this PR and deferring all troubles.
What do you think, @mtar? :D

mtar:
In the removed part, the underlying local Tensor is extracted from DNDarray. This could also be reason for the discrepancy.
In some other parts we are still using int64, so I'm fine with closing it.

PR #2376, opened 2026-07-12
Fix distributed GaussianNB to match scikit-learn (per-class indexing & class_count_)

### LINKED ISSUE #771: GaussianNB distributed results deviate from the scikit-learn results
**Description**

After the indexing update #758 and subsequent `naive_bayes.GaussianNB` changes, results from the distributed tests deviate enough from the `sklearn` comparison that tests fail.  Those tests have been temporarily commented out in #758, as the problem is not due to the indexing changes. 

TODO: clean up the `ht.naive_bayes.GaussianNB` implementation, clean separation of  local (torch) operations from distributed (heat) operations needed.

**To Reproduce**
1. Which module/class/function is affected?
2. What are the circumstances under which the bug appears?
unit tests 
3. What is the exact error message / erroneous behavior?

**Expected behavior**
Distributed `ht.naive_bayes.GaussianNB` should return the same result as the non-distributed case (and same as sklearn)

**Version Info**
branch [754-getitem-indexing](https://github.com/helmholtz-analytics/heat/tree/bug/754-getitem-indexing), PR #758 

**Additional comments**

maliesen:
- save class_count with split=None so all ranks can access it
- get the index/mask for the global x with respect to the global y, not y.larray

## Due Diligence

- General:
- Implementation:

## Description

The distributed results of `ht.naive_bayes.GaussianNB` deviated from scikit-learn (and from the
non-distributed case), which is why the distributed assertions were commented out in #758.

Root cause: in the per-class fit loop, the sample rows for each class were selected with
**rank-local** indices (`torch.where(y.larray == y_i)`) but used to index the **global**,
distributed `x` — so every process except the first selected the wrong rows. In addition,
`class_count_` was stored as a per-rank `(comm.size, n_classes)` array and summed afterwards,
which entangled the count bookkeeping with the split layout.

This PR selects each class's samples with a global boolean mask on the distributed `DNDarray`,
stores `class_count_` as a replicated `(n_classes,)` array, and centralizes/cleans up the
handling of the `classes` argument.

Issue/s resolved: #771 

## Changes proposed:

- Select per-class samples with a global boolean mask (`y == y_i`) instead of rank-local
  indices, so the correct rows of the distributed `x` (and `sample_weight`) are selected on
  every process — the root-cause fix for #771.
- Store `class_count_` as a replicated `(n_classes,)` array (`split=None`) instead of a
  per-rank `(comm.size, n_classes)` split array that was summed afterwards; simplify the
  empirical-prior computation accordingly. `class_count_` now has the same shape as
  scikit-learn's.
- Centralize the `classes` replication guard in `__partial_fit` (covering both `fit` and
  `partial_fit`) using `is_distributed()` to avoid an unnecessary resplit in serial; assert the
  replicated invariant in the fit loop and drop the now-unreachable class-index fallback branch.
- Make `__check_partial_fit_first_call` a pure predicate and assign `self.classes_` explicitly
  in `__partial_fit`.
- Fix the `x` shape in the `fit` docstring (`(n_samples, n_features)`).
- Re-enable the distributed assertions in `test_fit_iris` (`theta_`, `sigma_`, predictions, and
  the sample-weighted path) that were disabled in #758, and add `test_distributed_classes`.

## Type of change

Bug fix (non-breaking change which fixes an issue).

## Memory requirements

No meaningful change. `class_count_` is now a replicated `(n_classes,)` array on every rank
instead of a `(comm.size, n_classes)` split array; both are tiny (per-class), so memory is
equal or slightly lower.

## Performance

Not formally benchmarked; the per-class statistics operate on small arrays. Boolean-mask
`DNDarray` indexing replaces the local `torch.where`, and the post-hoc cross-rank sum of
`class_count_` is removed. Expected impact is negligible; correctness is the goal.

yes — internal to `GaussianNB`: the `class_count_` attribute shape changes from `(1, n_classes)`
to `(n_classes,)` (now matching scikit-learn), and `__check_partial_fit_first_call` no longer
has the side effect of setting `classes_`.


brownbaerchen:
Hi, @maliesen! Thanks a lot for getting started on this!

The tests in this PR are failing because some part of our pipeline has recently changed and we have yet to fix that. This has nothing to do with you and I apologize for the inconvenience. We'll try to fix this up as soon as possible.

maliesen:
This pr is still a draft as there are several concerns i would like to discuss. But maybe someone can already have a look (tried to make the diff as minimal as possible) to see if its going in the intended direction.

1. In https://github.com/helmholtz-analytics/heat/blob/86056b8c8089a2c9e6fe30d19a3290df089992bd/heat/naive_bayes/gaussianNB.py#L79 it should be `Shape = (n_samples, n_features)`. While strictly speaking probably out of scope I would pragmatically just include the docstring fix since we are already touching the file
2. In the `fit` method `classes` are extracted from `y` and then we have a guard to ensure `classes` are replicated over all ranks:
https://github.com/helmholtz-analytics/heat/blob/86056b8c8089a2c9e6fe30d19a3290df089992bd/heat/naive_bayes/gaussianNB.py#L97-L98
In `partial_fit` the caller must provide `classes` on the first call and may do so on subsequent calls. Would it make sense to have the same guard here as well? We could pull it to the private __partial_fit method then. We then don't need the else branch in https://github.com/helmholtz-analytics/heat/blob/86056b8c8089a2c9e6fe30d19a3290df089992bd/heat/naive_bayes/gaussianNB.py#L348 anymore.
3. Right now we have a problem with `partial_fit` if at some call we don't have any sample data to a given class. This is due to the fact that we check for equality in https://github.com/helmholtz-analytics/heat/blob/86056b8c8089a2c9e6fe30d19a3290df089992bd/heat/naive_bayes/gaussianNB.py#L336 whereas in https://github.com/scikit-learn/scikit-learn/blob/269c84b67bd3734c132e1a5fab1ba1a8fdc18015/sklearn/naive_bayes.py#L495 they check for set inclusion aligning with the potential error message https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/naive_bayes.py#L498-L501. I think this is out of scope for this pr as this issue already comes into play for undistributed data. (On the other hand maybe we do the sort earlier out of performance reasons.)
4. I think in https://github.com/helmholtz-analytics/heat/blob/86056b8c8089a2c9e6fe30d19a3290df089992bd/heat/naive_bayes/gaussianNB.py#L129 the correct type might be `float` instead of `int` but I must spend more time to understand. This is also potentially out of scope.

The list is not necessarily exhaustive and not fully thought through but I wanted to share. 👐

heat/naive_bayes/gaussianNB.py
```diff
@@ -76,7 +76,7 @@ def fit(self, x: DNDarray, y: DNDarray, sample_weight: Optional[DNDarray] = None
         ----------
         x : DNDarray
             Training set, where n_samples is the number of samples
-            and n_features is the number of features.  Shape = (n_classes, n_features)
+            and n_features is the number of features.  Shape = (n_samples, n_features)
```
brownbaerchen, line 79: Small fixes like this are definitely welcome here 👍

heat/naive_bayes/gaussianNB.py
```diff
@@ -282,6 +280,9 @@ def __partial_fit(
         # deviation of the largest dimension.
         self.epsilon_ = self.var_smoothing * ht.var(x, axis=0).max()
 
+        if classes is not None and classes.split is not None:
+            classes = ht.resplit(classes, axis=None)
```
brownbaerchen, line 282:
```suggestion
        if classes is not None and classes.is_distributed():
            classes = ht.resplit(classes, axis=None)
```
`is_distributed()` is preferable over `split is not None` because this prevents unnecessary copy when running in serial and the code is slightly cleaner. We are slowly changing this throughout the code when we notice it.

I think it does make sense to move the resplitting of the classes..

The entire file is not very clean and while we're at it, we may do some more refactoring..

The function `__check_partial_fit_first_call` has the unexpected side effect of assigning `classes` to `self.classes_`. I suggest you move this assignment out of that function and into this one, e.g. in line 290.
Then, I think resplitting `classes` belongs in this function and the code becomes a little less confusing..

heat/naive_bayes/gaussianNB.py
brownbaerchen, line 286:
```suggestion
            self.classes_ = classes
```

brownbaerchen:
```suggestion
```

brownbaerchen:
```suggestion
        if self.__check_partial_fit_first_call(classes) or _refit:
```

heat/naive_bayes/gaussianNB.py
```diff
@@ -282,6 +280,9 @@ def __partial_fit(
...
+
         if _refit:
             self.classes_ = None
```
brownbaerchen:
```suggestion
```

heat/naive_bayes/gaussianNB.py
```diff
@@ -298,7 +299,7 @@ def __partial_fit(
             else:
                 class_count_dtype = ht.types.promote_types(x.dtype, ht.float)
             self.class_count_ = ht.zeros(
-                (x.comm.size, n_classes), dtype=class_count_dtype, device=x.device, split=0
+                (n_classes), dtype=class_count_dtype, device=x.device, split=None
```
brownbaerchen, line 298: good catch!

heat/naive_bayes/gaussianNB.py
```diff
@@ -343,16 +344,12 @@ def __partial_fit(
         # DNDarrays for distributed operations only
         for y_i in unique_y.larray:
             # assuming classes.split is None
```
brownbaerchen:
```suggestion
            assert not classes.is_distributed()
```

brownbaerchen (commented):
Hi @maliesen, thanks a lot for your work!

I added a bunch of suggestions for more refactoring. Let me know what you think.

As far as I am concerned, this PR is already useful as is, since the problematic tests are now running. If you don't want to do the additional refactoring, I would be fine with merging as is.

brownbaerchen:
Regarding points 3 and 4, I agree that it's out of scope for this PR. If you feel like addressing point 3, you can just open another PR. I agree that scipy is sometimes performing more checks on the data than is really needed and we can choose not to do that.

maliesen:
> Hi @maliesen, thanks a lot for your work!
> 
> I added a bunch of suggestions for more refactoring. Let me know what you think.
> 
> As far as I am concerned, this PR is already useful as is, since the problematic tests are now running. If you don't want to do the additional refactoring, I would be fine with merging as is.

Hi @brownbaerchen,
thank you very much for the comments and code suggestions! I agree on all of them and integrated them into the pull request. For the headers of the pr (due diligence, description, etc.) I took over AI generated suggestions as I think they are well formulated - please feel free to give feedback if we rather want to have it manually with own (human) words or anything changed. Also, I wasn't quite sure about the checkboxes.

tests/naive_bayes/test_gaussiannb.py
```diff
@@ -94,23 +94,23 @@ def test_fit_iris(self):
...
-        # X_test_split = ht.resplit(X_test, axis=0)
+        X_test_split = ht.resplit(X_test, axis=0)
         y_train_split = ht.resplit(y_train, axis=0)
-        # y_test_split = ht.resplit(y_test, axis=0)
-        # y_pred_split = gnb_heat.fit(X_train_split, y_train_split).predict(X_test_split)
+        y_test_split = ht.resplit(y_test, axis=0)
+        y_pred_split = gnb_heat.fit(X_train_split, y_train_split).predict(X_test_split)
         self.assert_array_equal(gnb_heat.class_prior_, sklearn_class_prior)
         self.assert_array_equal(gnb_heat.epsilon_, sklearn_epsilon)
         # TODO: clean up gnb Issue #771
```
brownbaerchen:
```suggestion
```

heat/naive_bayes/gaussianNB.py
```diff
@@ -113,14 +111,13 @@ def __check_partial_fit_first_call(self, classes: Optional[DNDarray] = None) ->
...
-                self.classes_ = unique_labels
                 # This is the first call to partial_fit
                 return True
             if not ht.equal(self.classes_, unique_labels):
                 raise ValueError(
                     f"`classes={classes}` is not the same as on last call to partial_fit, was: {self.classes_}"
                 )
-        # classes is None and self.classes_ has already previously been set:
+        # classes is None or self.classes_ has already previously been set:
         # nothing to do
```
brownbaerchen:
```suggestion
        # classes is None or self.classes_ has already previously been set
```

brownbaerchen (dismissed):
Thanks a lot, @maliesen! You know how to reach us if you need anything else! In case you are interested, please join our [hackathon](https://indico3-jsc.fz-juelich.de/event/327/) in November.

brownbaerchen (approved):

PR #2377, opened 2026-07-13
Print world rank rather than 0 in setup guide smoke test

### LINKED ISSUE #2373: [Bug]: Rank is taken from wrong communicator
### What happened?

In the quick_start.md the command

https://github.com/helmholtz-analytics/heat/blob/4af4f7fbef7088630830ae2d996e6b0ee27a9516/quick_start.md?plain=1#L53

always prints Rank 0. `ht.communication.MPI_SELF.rank` should be replaced by `x.comm.rank` (or  `ht.communication.MPI_WORLD.rank`)

```shell

```

```shell

```

### Version

main (development branch)

### Python version

3.12

### PyTorch version

2.3

### MPI version

```shell
openmpi-5.0.10-h67ed482_1
```

brownbaerchen:
As @maliesen for pointed out in #2373, the distributed smoke test in the setup guide prints the rank of `COMM_SELF`, which is always zero, instead of the actually relevant `COMM_WORLD`.
This PR fixes this. Note that this PR doesn't change any code but only documentation.

Issue/s resolved: #2373

### DIFF
--- a/quick_start.md
+++ b/quick_start.md
@@ -50,7 +50,7 @@ See our [docker README](https://github.com/helmholtz-analytics/heat/blob/main/do
 Confirm the installation by running a distributed smoke test to verify tensor splitting across processes:
 
 ```bash
-mpirun -n 2 python -c "import heat as ht; x = ht.arange(10, split=0); print(f'Rank {ht.communication.MPI_SELF.rank} local shape: {x.lshape}')"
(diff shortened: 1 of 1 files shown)


ClaudiaComito (approved):
Thanks @brownbaerchen and @maliesen !

Issue #2375, opened 2026-07-12
[Bug]: ht.kurtosis(x, axis=0) ignores `Fischer`


maliesen:
### What happened?

In ht.kurtosis() there is an optional parameter `Fischer` defaulting to `True`. But when an axis is given the code does not correctly carry over the argument value to pytorch.

###Fix
In
https://github.com/helmholtz-analytics/heat/blob/4af4f7fbef7088630830ae2d996e6b0ee27a9516/heat/core/statistics.py#L1307
we should check if the optional value of `Fischer` is present:
```diff
-if Fischer:
+if Fischer is not None:
```

### Code snippet triggering the error

```shell

```

### Error message or erroneous outcome

```shell

```

### Version

main (development branch)

### Python version

3.12

### PyTorch version

2.3

### MPI version

```shell

```

mtar:
Fixed in #2379

PR #2379, opened 2026-07-13
Fix for #2374: ht.var(x, axis=0) ignores ddof

### LINKED ISSUE #2374: [Bug]: ht.var(x, axis=0) ignores ddof
### What happened?

In `ht.var()` there is an optional parameter `ddof` defaulting to `0`. But when an `axis` is given the code does not correctly carry over the argument value to pytorch.

### Fix
In 
https://github.com/helmholtz-analytics/heat/blob/4af4f7fbef7088630830ae2d996e6b0ee27a9516/heat/core/statistics.py#L1305
we should check if the optional value of `unbiased` is present:
```diff
-if unbiased:
+if unbiased is not None:
```

```shell
import heat as ht

x = ht.array([0.])
print(var(x, axis=0, ddof=1)) # correctly nan
print(var(x, axis=0, ddof=0)) # nan, expected 0.
print(var(x, axis=0)) # nan, expected 0.
```

```shell

```

### Version

main (development branch)

### Python version

3.12

### PyTorch version

2.3

### MPI version

```shell

```

brownbaerchen:
See #2374 for the description of the problem reported by @maliesen.

Short recap: The function `__moment_w_axis` takes `unbiased` as input and forwards it to the respective torch function (`var` in this case). However, it only does so if `unbiased` is `True`. We need to forward `unbiased=False` as well, though.

The same happens with `Fisher` (see #2375) which may also be fixed by this PR. It remains to be checked, though.

Issue/s resolved: #2374


heat/core/statistics.py
```diff
@@ -1302,9 +1302,9 @@ def __moment_w_axis(
     """
     # helper for calculating a statistical moment with a given axis
     kwargs = {"dim": axis}
-    if unbiased:
+    if unbiased is not None:
         kwargs["unbiased"] = unbiased
-    if Fischer:
+    if Fischer is not None:
```
mtar: Parameter name should be `fischer` due to the code style. We use it in other places too.

heat/core/statistics.py
```diff
@@ -1302,9 +1302,9 @@ def __moment_w_axis(
     """
     # helper for calculating a statistical moment with a given axis
     kwargs = {"dim": axis}
-    if unbiased:
+    if unbiased is not None:
```
mtar: `unbiased` is `correction` in [torch.var](https://docs.pytorch.org/docs/2.13/generated/torch.var.html) since version 2.0 and works with integers. We will have to look what it is for the other functions.

mtar (changes requested):

maliesen:
I don't know if this is the right place for it, but upon merging the fix we can now omit the four `atol=1e-1` argument values in the assertions in tests/naive_bayes/test_gaussiannb.py. This is to reflect the change in the numerical values of `ht.naive_bayes.GaussianNB`:

https://github.com/helmholtz-analytics/heat/blob/5d9851ba0b5cbec9ccad2cc8b4fa918f8e132df0/tests/naive_bayes/test_gaussiannb.py#L79-L80
https://github.com/helmholtz-analytics/heat/blob/5d9851ba0b5cbec9ccad2cc8b4fa918f8e132df0/tests/naive_bayes/test_gaussiannb.py#L91
https://github.com/helmholtz-analytics/heat/blob/5d9851ba0b5cbec9ccad2cc8b4fa918f8e132df0/tests/naive_bayes/test_gaussiannb.py#L105

heat/core/statistics.py
brownbaerchen: It's also called `correction` in numpy. See the [docs](https://numpy.org/doc/stable/reference/generated/numpy.var.html). 
Scipy kurtosis and skew have `bias`, see [docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kurtosis.html).

I switched to `correction` internally, but left the interface of `ht.skew` and `ht.kurtosis` to keep `unbiased`. What do you think?

brownbaerchen:
@maliesen, thanks for pointing this out! You are solving all of our bugs! :)

heat/core/statistics.py
mtar: I think we should revert correction back for now. It is only in var and accepts ints. Maybe, change it to _bias_ and flip the values.
Nonetheless, we should use the same parameters as in the other APIs for our public functions and update the calls to torch.var

brownbaerchen: `correction` is the closest to the torch and numpy APIs, so why would be do something yet completely different?

mtar: Actually, I don't see it close to numpys API.

Torch/NumPy:
- correction is only used in var
- correction takes ints and floats and is a array api compatible name for ddof.
- skew and kurtosis use bias in scipy and are bools

PR:
- correction is used everywhere
- correction takes bools in var and it has a different logic underneath
- skew and kurtosis also have the correction parameter name

Right now in the PR, it is nothing alike but the name in a single place.

mtar: We will keep it for now.

mtar (approved):
The title of the pr needs a better description of the fix.

PR #2386, opened 2026-07-16
[Breaking Change] Implement argsort and make sort numpy API compliant

### LINKED ISSUE #777: Implement `argsort`
**Related**
#774

**Feature functionality**
see https://numpy.org/doc/stable/reference/generated/numpy.argsort.html

This functionality is already present within `ht.sort`, make API-compliant

**Additional context**
See [Python array API standard](https://data-apis.org/array-api/latest/index.html)

Berkant03:
## Due Diligence

- General:
- Implementation:

## Description

`argsort` and `sort` have been changed to reflect the Numpy API, while also allowing old usage.

Issue/s resolved: #777 

## Type of change
- New feature 

yes, `sort` now does not give back the indices when sorting like pytorch. Instead a flag must be given to achieve same behaviour.


tests/core/test_manipulations.py
```diff
@@ -7,6 +7,116 @@
...
+        with self.assertRaises(ValueError):
+            ht.argsort(data, axis=3)
+        with self.assertRaises(TypeError):
+            ht.argsort(data, axis="1")
+
+        rank = ht.MPI_WORLD.rank
+        ht.random.seed(1)
+        data = ht.random.randn(100, 1, split=0)
+        indices = ht.argsort(data, axis=0)
+        result = ht.resplit(data, None)[indices]
```
Berkant03: Interestingly. When resplitting after indexing, the array is not sorted and more garbled. This is a working workaround, which I think should not be needed.

heat/core/manipulations.py
```diff
@@ -65,6 +66,69 @@
...
+    order: str | list[str] | None = None,
+    *,
+    stable: bool | None = None,
+    descending: bool | None = None,
+) -> DNDarray:
+    """
+    Returns the indices that would sort an array. This is the distributed equivalent of `np.argsort`.
+    The sorting is not stable which means that equal elements in the result may have a different ordering than in the
+    original array.
+    Sorting where `axis==a.split` needs a lot of communication between the processes of MPI.
```
```suggestion
    Sorting with `axis==a.split` needs a lot of communication between the processes.
```

heat/core/manipulations.py
```diff
@@ -65,6 +66,69 @@
...
+    ----------
+    a : DNDarray
+        Input array to be sorted.
+    axis : int, optional
+        The dimension to sort along.
+        Default is the last axis.
+    kind : str, optional
+        Sorting algorithm. Gets ignored.
+    order : str, list[str], optional
+        Used in numpy for array with fields, which are not possible in HeAT. Gets ignored.
```
```suggestion
        Used in numpy for array with fields, which are not possible in Heat. Gets ignored.
```

heat/core/manipulations.py
```diff
@@ -2564,11 +2638,20 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
     axis : int, optional
         The dimension to sort along.
         Default is the last axis.
+    kind : str, optional
+        Sorting algorithm. Gets ignored.
+    order : str, list[str], optional
+        Used in numpy for array with fields, which are not possible in HeAT. Gets ignored.
+    stable : bool, optional
+        Sort stability, currenty not supported.
```
brownbaerchen: Do we really want to allow arguments that are ignored? What's the policy on this, @ClaudiaComito?
At least, I would `assert not stable` and raise a warning that the parameters are ignored rather than silently ignoring them.

heat/core/manipulations.py
```diff
@@ -2597,6 +2680,13 @@ def sort(a: DNDarray, axis: int = -1, descending: bool = False, out: Optional[DN
     )
     stride_tricks.sanitize_axis(a.shape, axis)
 
+    if axis is None:
+        a = flatten(a)
+        axis = 0
+
+    descending = descending or False
+    return_sort_indices = return_sort_indices or False
```
brownbaerchen: I presume this is to cast `None` to `False`. Why do we allow `None` as input? Seems to have the same impact as just using `False`...

brownbaerchen (changes requested):
Thanks! I think it's a good idea to be closer to Numpy and to support `argsort`. However, I don't think we should silently ignore additional arguments that Numpy users are used to.


Note that the errors in the collection step in pytest happen all over the place, including in other repositories, and have nothing to do with this PR.

heat/core/manipulations.py
brownbaerchen: We want to use the same interface as numpy but call torch functions which requires this..

tests/core/test_manipulations.py
brownbaerchen: Surely, this will be fixed in #938

brownbaerchen:
Excellent! If you apply something similar to `argsort`, this should be ready to merge.

tests/core/test_manipulations.py
```diff
@@ -7,6 +7,116 @@
 
 
 class TestManipulations(TestCase):
+    def test_argsort(self):
+        size = ht.MPI_WORLD.size
+        rank = ht.MPI_WORLD.rank
+        tensor = (
+            torch.arange(size, device=self.device.torch_device).repeat(size).reshape(size, size)
+        )
```
brownbaerchen: Can we switch to random data? Something like
```python
data = ht.random.rand(2, 3, 4)
for axis in [0, 1, 2]:
    for descending in [True, False]:
        for split in [0, 1, 2]:
            data.resplit_(split)
            result_indices = ht.argsort(data, axis=axis, descending=descending)
            exp_indices = np.argsort(data.numpy(), axis=axis, descending=descending)
            self.assertTrue(np.allclose(result_indices.numpy(), exp_indices))
```

This would make the tests more concise.

Also, note the switch to numpy, which then tests that the API is the same, not just the result.

tests/core/test_manipulations.py
brownbaerchen, line 1: Can you please add a test for `sort` similar to what I suggested for `argsort` here? Feel free to use `pytest.parametrize` and move away from the `unittest` classes if you want.

brownbaerchen (changes requested):
Thanks! I think we can clean up the tests a bit more. Otherwise, this is ready in my opinion.

Berkant03:
The current tests fail, due to an API change in numpy 2.5, which added the `descending` keyword to `argsort`, but is absent in previous versions. How should we handle the cases with a lower numpy version? @brownbaerchen

brownbaerchen:
> The current tests fail, due to an API change in numpy 2.5, which added the `descending` keyword to `argsort`, but is absent in previous versions. How should we handle the cases with a lower numpy version? @brownbaerchen

That's annoying! I think the best we can do is to split the tests into ones with `descending` and without. Then you can skip tests using `descending` if `np.lib.NumpyVersion(np.__version__) < '2.5.0'`.
You can use a decorator:
```python
class Tests(...):
    @unittest.skipif(np.lib.NumpyVersion(np.__version__) < '2.5.0', f'Numpy version {np.__version__} does not support `descending` keyword in `sort`)
    def test_sort(self):
        ...
    def test_sort_no_descending_argument(self):
        ...
```
Alternatively, you can use the `unittest.subtest` stuff within a single function and skip some subtests from there.
Neither is particularly clean.. Do use skipping infrastructure, though. This will tell you how many tests were skipped during the collection step and why and is not silent.

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
+import torch
+import pytest
+import os
+import heat as ht
+
+
+class TestSorting:
+    def __init__(self, *args, **kwargs):
+        self.device = os.environ.get("HEAT_TEST_USE_DEVICE", "cpu")
```
brownbaerchen: The place where this is set up in other tests is [here](https://github.com/helmholtz-analytics/heat/blob/a94030afc1e5c51103a50d4f15a325cbe5ec49ea/heat/testing/basic_test.py#L32).
It's annoying that this is actually not called in pytest. At least we are not certain that it's called before any tests are run without this infrastructure that don't run on GPU.
We need to find a solution to this. For now, I suggest 
```suggestion
    def __init__(self, *args, **kwargs):
        ht.testing.basic_test.TestCase.setUpClass()
```
and then you can get rid of all the `self.device` down below.

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+    @pytest.mark.parametrize("axis", [None, 0, 1, 2])
+    def test_sort(self, axis, descending, split):
+        kwargs = {"axis": axis}
+
+        if np.lib.NumpyVersion(np.__version__) < np.lib.NumpyVersion('2.5.0'):
+            pytest.skip(f"NumPy {np.__version__} does not support the 'descending' keyword.")
+        else:
+            kwargs["descending"] = descending
+
+        data = ht.random.rand(2, 3, 4, split=split, device=self.device)
```
brownbaerchen:
```suggestion
        data = ht.random.rand(2, 3, 4, split=split)
```

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+    @pytest.mark.parametrize("axis", [None, 0, 1, 2])
+    def test_argsort_random(self, axis, descending, split):
+        kwargs = {"axis": axis}
+
+        if np.lib.NumpyVersion(np.__version__) < np.lib.NumpyVersion('2.5.0'):
+            pytest.skip(f"NumPy {np.__version__} does not support the 'descending' keyword.")
+        else:
+            kwargs["descending"] = descending
+
+        data = ht.random.rand(2, 3, 4, split=split, device=self.device)
```
brownbaerchen:
```suggestion
        data = ht.random.rand(2, 3, 4, split=split)
```

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+                [[8, 4, 7], [0, 8, 9]],
+            ],
+            dtype=torch.int32,
+            device=self.device,
+        )
+
+        data = ht.array(tensor, split=0)
+        if torch.cuda.is_available() and data.device == ht.gpu and size < 4:
+            indices_axis_zero = torch.tensor(
+                [[0, 2, 2], [3, 2, 0]], dtype=torch.int32, device=self.device
```
brownbaerchen:
```suggestion
                [[0, 2, 2], [3, 2, 0]], dtype=torch.int32
```

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+            )
+        else:
+            indices_axis_zero = torch.tensor(
+                [[0, 2, 2], [3, 0, 0]], dtype=torch.int32, device=self.device
```
brownbaerchen:
```suggestion
                [[0, 2, 2], [3, 0, 0]], dtype=torch.int32
```

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+            )
+        result_indices = ht.argsort(data, axis=0)
+        first_indices = result_indices[0].larray
+        if rank == 0:
+            assert torch.equal(first_indices, indices_axis_zero)
+
+        data = ht.array(tensor, split=1)
+        indices_axis_one = torch.tensor(
+            [[0, 1, 1]], dtype=torch.int32, device=self.device
```
brownbaerchen:
```suggestion
            [[0, 1, 1]], dtype=torch.int32
```

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+        result_indices = ht.argsort(data, axis=1)
+        first_indices = result_indices[0].larray[:1]
+        if rank == 0:
+            assert torch.equal(first_indices, indices_axis_one)
+
+        data = ht.array(tensor, split=2)
+        indices_axis_two = torch.tensor(
+            [[0], [1]], dtype=torch.int32, device=self.device
```
brownbaerchen:
```suggestion
            [[0], [1]], dtype=torch.int32
```

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+
+        tensor = torch.tensor(
+            [
+                [[2, 8, 5], [7, 2, 3]],
+                [[6, 5, 2], [1, 8, 7]],
+                [[9, 3, 0], [1, 2, 4]],
+                [[8, 4, 7], [0, 8, 9]],
+            ],
+            dtype=torch.int32,
+            device=self.device,
```
brownbaerchen:
```suggestion
            dtype=torch.int32,
```

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # test exceptions
+        with pytest.raises(ValueError):
+            ht.argsort(data, axis=3)
+        with pytest.raises(TypeError):
+            ht.argsort(data, axis="1")
+
+        rank = ht.MPI_WORLD.rank
+        ht.random.seed(1)
+        data = ht.random.randn(100, 1, split=0, device=self.device)
```
brownbaerchen:
```suggestion
        data = ht.random.randn(100, 1, split=0)
```

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+    @pytest.mark.parametrize("split", [None, 0, 1, 2])
+    @pytest.mark.parametrize("descending", [None, True, False])
+    @pytest.mark.parametrize("axis", [None, 0, 1, 2])
+    def test_sort(self, axis, descending, split):
+        kwargs = {"axis": axis}
+
+        if np.lib.NumpyVersion(np.__version__) < np.lib.NumpyVersion('2.5.0'):
+            pytest.skip(f"NumPy {np.__version__} does not support the 'descending' keyword.")
+        else:
+            kwargs["descending"] = descending
```
brownbaerchen: This skips all the tests with older Numpy versions, right?
Can we do 
```suggestion
        if descending in [True, False]:
            if np.lib.NumpyVersion(np.__version__) < np.lib.NumpyVersion('2.5.0'):
                pytest.skip(f"NumPy {np.__version__} does not support the 'descending' keyword.")
            else:
                kwargs["descending"] = descending
```

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+    @pytest.mark.parametrize("split", [None, 0, 1, 2])
+    @pytest.mark.parametrize("descending", [None, True, False])
+    @pytest.mark.parametrize("axis", [None, 0, 1, 2])
+    def test_argsort_random(self, axis, descending, split):
+        kwargs = {"axis": axis}
+
+        if np.lib.NumpyVersion(np.__version__) < np.lib.NumpyVersion('2.5.0'):
+            pytest.skip(f"NumPy {np.__version__} does not support the 'descending' keyword.")
+        else:
+            kwargs["descending"] = descending
```
brownbaerchen: Same comment as above for not skipping with `descending=None`.

tests/core/test_sorting.py
```diff
@@ -0,0 +1,107 @@
...
+
+        data = ht.random.rand(2, 3, 4, split=split, device=self.device)
+        result_indices = ht.argsort(data, **kwargs)
+        exp_indices = np.argsort(data.numpy(), **kwargs)
+        assert np.allclose(result_indices.numpy(), exp_indices)
+
+    def test_argsort_specific(self):
```
brownbaerchen: What's the point of these tests? Can you add a brief comment what makes the special cases special enough to warrant these extra tests? Like `# test with integers` and so on.

brownbaerchen (changes requested):
Sorry, I have a few more small points, but should be quick to address 🙈

tests/core/test_sorting.py
Berkant03: Here I do need the device, which I can just get by using `ht.get_device().torch_device`, as the torch tensors do not know about the heat device. One could also question if the specific test is needed at all, as one could argue that the random data is enough.

tests/core/test_sorting.py
Berkant03: These were already there when I started working on it. I think the idea behind it was to have something calculated manually for safety. I would say we can remove them If you want, since the random cases should be sufficient enough.

brownbaerchen: yes, anything that is also covered by random data should go. Tests for correct assertions and other special cases can stay 👍

brownbaerchen (dismissed):
Feel free to remove any unnecessary tests, or not. Otherwise, we can finally merge :D

brownbaerchen:
The failures in previous pushes look unrelated to this.. No idea what's going on there... :(

Berkant03:
@brownbaerchen, I removed the specific test. The codebase test fails due to not enough storage. If everything is fine, you can merge it.

brownbaerchen:
> @brownbaerchen, I removed the specific test. The codebase test fails due to not enough storage. If everything is fine, you can merge it.

Let's merge! @JuanPedroGHM, do you have some bandwidth for looking into why we are suddenly running out of disk space on codebase during installation?

brownbaerchen (approved):

PR #2391, opened 2026-07-17
Allow unsafe pr checkout

brownbaerchen:
This is the same as #2378, but no longer across forks. Otherwise, the changes don't take effect because we are still running with `pull_request_target` and the tests just fail.

I now understand copilot's comments in @2378: For triggering the codebase CI, we don't need the actual code that we want to test. It's suffices to checkout the current main, which contains the action that tells codebase what to check out.
So, by not specifying any `ref` in `checkout`, we checkout the base branch, which is safe as it is part of the main repo. This launches the action on codebase and codebase does whatever the PR includes. This may be malicious, but it can't extract secrets from us.

We can still choose if we want to always manually trigger workflows from external contributors or only from first time contributors, but I think this solution is good enough either way.


JuanPedroGHM:
That is a smart idea. I would still keep setup to require approval for first time external people, but not checking their branch to launch the Codebase jobs is a good plan.

JuanPedroGHM (approved):

PR #2414, opened 2026-07-31
Add datasets to Jupyter4nfdi

brownbaerchen:
This PR adds access to the datasets required to run the tutorials to the Jupyter4NFDI config and adapts the notebooks to choose the correct path (and device) in both JSC and Jupyter4NFDI environments.
Loading the climate data didn't work because I ran out of memory, but at least the other one works..

Note that we need to merge main into the `jupyter4nfdi` branch afterwards for the changes to the tutorials to take effect.


mtar (approved):
Looks fine. Thank you!

PR #2422, opened 2026-08-03
add newaxis constant

mtar:
## Due Diligence

- General:
    - [ ]  **title** of the PR is suitable to appear in the [Release Notes](https://github.com/helmholtz-analytics/heat/releases/latest)
- Implementation:
    - [ ] unit tests: all split configurations tested
    - [ ] unit tests: multiple dtypes tested
    - [ ] **NEW** unit tests: MPS tested (1 MPI process, 1 GPU)
    - [ ] benchmarks: created for new functionality
    - [ ] benchmarks: performance improved or maintained
    - [ ] documentation updated where needed

## Description

Adds the newaxis constant to the main namespace.

Issue/s resolved: #2417 

## Changes proposed:

- add newaxis constant

## Type of change

array api

#### Does this change modify the behaviour of other functions? If so, which?
yes / no

PR #2431, opened 2026-08-04
Add vectorized_sort function

### LINKED ISSUE #363: vectorized sorting

Berkant03:
## Due Diligence
- General:
- Implementation:

## Description

Issue/s resolved: #363

## Changes proposed:

Introduce the vectorized sorting mentioned in #363, which is also a lexicographical sort
Allows to finalize #2332.

## Type of change

## Memory requirements
A complete alltoallv requires a total of 3x the memory here: the original buffer, contiguous send buffer and the receive buffer.

One could in theory implement an own chunking algorithm to reduce this, if needed. That might make the performance worse or better. Testing is the only way to know for sure.

Custom datatypes should also be considered, but would need extensive analysis on the MPI implementation itself, since the sending rank might want to send multiple packets and might need to pack them together, creating again a copy, unless the network protocol can completely handle the non-contiguous data.

## Performance
-

no


Berkant03:
I have written some pytest tests, but would wait until #777 is merged, to write them into the `TestSorting` class.

brownbaerchen:
> I have written some pytest tests, but would wait until #777 is merged, to write them into the `TestSorting` class.

#2386 was actually already merged. 
For future things like this, see the very cool new [GitHub stack feature](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests). Using this feature, you can branch out of the development branch of #2386, implement something on top, start a PR and stack it in top of #2386. Every PR only shows it's own diff and not the one from PRs below in the stack, making it easy to review. PRs below in the stack can be merged independently, or the entire stack can be merged at once. #2422 is an example where we stacked PRs.

Don't worry about this now. I just love this new feature and want to advertise it ;)

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    columns first, which acts as the primary sort key, with subsequent
+    columns acting as secondary, tertiary, etc., keys.
+
+    Parameters
+    ----------
+    a : DNDarray
+        The array to be sorted.
+    axis : int, optional
+        The axis along which to sort. If the split dimension of the array does
+        not match this axis, the array is resplit.
```
brownbaerchen, line 2927: Why do we require `split = axis`? Here we need to reshuffle a lot anyways, right? What can you do more efficiently by resplitting? Or is it just simpler to implement only this one case? (That would be very valid)

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+        Default is -1 (last axis).
+    stable : bool, optional
+        Whether the sorting algorithm should be stable. Default is True.
+    descending : bool, optional
+        Whether to sort in descending order. Default is False.
+    resplit_result : bool, optional
+        Whether to resplit the final sorted array back to the original split
+        axis of the input array after rows are distributed. Default is True.
+    return_sort_indices_only : bool, optional
```
brownbaerchen: This parameter is a bit confusingly named. If I understand correctly, if this is true, it returns the sort indices, but if it's false, it returns the sorted data. The "only" implies that if false, it returns both the data and the indices.

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+
+    Returns
+    -------
+    DNDarray
+        Either the final sorted array, or if `return_sort_indices_only` is True, the global sort indices.
+    """
+    sanitation.sanitize_in(a)
+
+    if not isinstance(axis, int):
+        raise ValueError("'axis' must be an int.")
```
```suggestion
        raise ValueError(f"'axis' must be integer, not {type(axis)}")
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    if not isinstance(stable, bool):
+        raise ValueError("'stable' must be a bool.")
```
```suggestion
        raise ValueError(f"'stable' must be bool, not {type(stable)}")
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    if not isinstance(descending, bool):
+        raise ValueError("'descending' must be a bool.")
```
```suggestion
        raise ValueError(f"'descending' must be bool, not {type(descending)}")
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    if not isinstance(resplit_result, bool):
+        raise ValueError("'resplit_result' must be a bool.")
```
```suggestion
        raise ValueError(f"'resplit_result' must be bool, not {type(resplit_result)}")
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    if not isinstance(return_sort_indices_only, bool):
+        raise ValueError("'return_sort_indices_only' must be a bool.")
```
```suggestion
        raise ValueError(f"'return_sort_indices_only' must be bool, not {(type(return_sort_indices_only)}")
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+
+    if len(a.gshape) == 0:
+        raise ValueError("dndarray must have atleast one dimension.")
```
```suggestion
        raise ValueError("dndarray must have at least one dimension.")
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+
+    if axis >= len(a.gshape) or (axis < 0 and abs(axis) > len(a.gshape)):
```
```suggestion
    if not (-a.ndim <= axis < a.ndim): 
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+        raise ValueError(f"'axis'={axis} does not exist for '{len(a.gshape)}' dimenions.")
```
```suggestion
        raise ValueError(f"{axis=} does not exist for array with {a.ndim} dimensions.")
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+
+    def _permute_indices(data, idx):
+        sort_idx = torch.argsort(data[idx], stable=stable, descending=descending)
+        return idx[sort_idx]
+
+    if len(a.gshape) == 1:
```
```suggestion
    # 1D vectorized sort is just sort
    if a.ndim == 1:
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    if not a.is_distributed():
+        data = a.larray.transpose(axis, 0)
+        shape = data.shape
+
+        data = data.reshape(shape[0], -1)
+        indices = torch.arange(0, data.shape[0])
+        for i in range(data.shape[-1] - 1, -1, -1):
+            indices = _permute_indices(data[:, i], indices)
+
+        data = data.reshape(shape)[indices]
```
brownbaerchen: How does this work for `data.ndim>2`? Reshape back and then apply 1D indices?

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+        sort_idx = torch.argsort(data[idx], stable=stable, descending=descending)
+        return idx[sort_idx]
+
+    if len(a.gshape) == 1:
+        arr, idx = sort(a, axis=axis, descending=descending, return_sort_indices=True)
+        if return_sort_indices_only:
+            return idx
+        return arr.resplit_(a.split) if resplit_result else arr
+
+    if not a.is_distributed():
```
brownbaerchen, line 2970:
```suggestion
    # early out for non-distributed input
    if not a.is_distributed():
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+        shape = data.shape
+
+        data = data.reshape(shape[0], -1)
+        indices = torch.arange(0, data.shape[0])
+        for i in range(data.shape[-1] - 1, -1, -1):
+            indices = _permute_indices(data[:, i], indices)
+
+        data = data.reshape(shape)[indices]
+        return factories.array(data.transpose(axis, 0), split=None)
+    original_split = a.split
```
```suggestion
        return factories.array(data.transpose(axis, 0), split=None)
        
    # distributed vectorized sort
    original_split = a.split
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    if axis != a.split:
+        a = resplit(a, axis)
+
+    comm = MPI.COMM_WORLD
```
```suggestion
    comm = a.comm
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    rank = comm.Get_rank()
+    size = comm.Get_size()
```
```suggestion
    rank = comm.rank
    size = comm.size
```
I think @mtar did a lot of these changes in #2350.

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+
+    data = a.larray.transpose(axis, 0)
```
brownbaerchen: maybe rename to `local_data` to make it more clear?

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    original_shape = data.shape
+    inner_shape = original_shape[1:]
+
+    local_count = data.shape[0]
+    total_rows = sum(comm.allgather(local_count))
```
```suggestion
    total_rows = np.prod(a.gshape) // a.gshape[a.split] 
```
requires no communication. Not sure if this is the exact formula we need, though.

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+    block_length = np.prod(inner_shape) if len(inner_shape) > 0 else 1
```
```suggestion
    block_length = np.prod(inner_shape)
```
```python
>>> np.prod(())
np.float64(1.0)
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+
+    mpi_type: MPI.Datatype = a.comm.mpi_type_of(data.dtype)
```
```suggestion
    mpi_type: MPI.Datatype = comm.mpi_type_of(data.dtype)
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+        send_counts = np.array(comm.gather(local_count), dtype=int)
+        send_displ = np.insert(np.cumsum(send_counts)[:-1], 0, 0)
+
+        buffer = torch.empty((total_rows,), dtype=data.dtype)
+        recv_args = [buffer, send_counts, send_displ, mpi_type]
+    else:
+        comm.gather(local_count)
+
+        buffer = None
+        recv_args = None
```
```suggestion
    local_counts = comm.gather(local_count, root=0)
    
    if rank == 0:
        send_counts = np.array(local_counts, dtype=int)
        send_displ = np.insert(np.cumsum(send_counts)[:-1], 0, 0)

        buffer = torch.empty((total_rows,), dtype=data.dtype)
        recv_args = [buffer, send_counts, send_displ, mpi_type]
    else:
        buffer = None
        recv_args = None
```
I always expect deadlock when collectives are in if clauses based on rank.

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+
+    def gather_column(flat_idx: int):
```
brownbaerchen: This function doesn't start with `_` like `_permute_indices` does even though both are the same amount of private. Would be nice to keep this consistent.

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+        idx = np.unravel_index(flat_idx, inner_shape)
+        slice_tuple = (slice(None),) + tuple(idx)
+
+        local_col = data[slice_tuple].contiguous()
+        comm.Gatherv(local_col, recv_args)
```
brownbaerchen: I hope this works with a heat communicator as well and not just with `mpi4py.COMM_WORLD`. We shouldn't use the world communicator, because that can lead to deadlocks, though.

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+        return buffer
+
+    if rank == 0:
+        indices = torch.arange(0, total_rows, dtype=torch.int64)
+    else:
+        indices = torch.empty(total_rows, dtype=torch.int64)
```
```suggestion
    indices = torch.arange(0, total_rows, dtype=torch.int64)
```
I don't really see a reason to allocate differently on different ranks here.

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+
+    for i in range(block_length - 1, -1, -1):
+        if rank == 0:
+            indices = _permute_indices(gather_column(i), indices)
+        else:
+            gather_column(i)
```
```suggestion
        gather_column(i)
        if rank == 0:
            indices = _permute_indices(buffer, indices)
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+
+    comm.Bcast(indices)
```
```suggestion
    comm.Bcast(indices, root=0)
```

brownbaerchen (changes requested):
It's a bit too hot for me to understand everything. I marked a bunch of small things that could be a bit cleaner. I suggest you add some tests and then I'll have another look.
Since we want to extract the side effect of sorting from `unique`, a comparison with `unique` would be a useful test.

brownbaerchen:
I failed at creating a stack of this and #2332, which depends on this. Getting the commit histories in order after the fact is not so easy :D I'll take care of fixing the merge conflicts that will arise when we merge #2332 after merging this. But on the bright side, the tests seem to pass :) Sorry about the mess.

heat/core/manipulations.py
Berkant03: It had to do with allowing a simpler data handling. When the split axis is the same as the sorting axis, I can just do a simple transpose, as the data is already there. But if it is not the same, I need to gather the sort axis data. Therefore I just used a resplit to make it simpler.

heat/core/manipulations.py
Berkant03: I now changed it up to `return_sort_indices_instead`

heat/core/manipulations.py
Berkant03: We reshape back to the shape, where the sorting axis is the first dimension. where we then apply the sorting indices. After which we place the sorting axis back on the correct dimension.

heat/core/manipulations.py
Berkant03: To only use the heat communicator here, I first need to change the allgather here: `total_rows = sum(comm.allgather(local_count))`, since I only send around an integer, which gets pickled. 

Both behave differently here. I can change it up though.

heat/core/manipulations.py
Berkant03: I unnecessarily used a communication here, instead of just using `total_rows = a.gshape[axis]`. This gets me the total elements on the sorting axis faster and simpler.

heat/core/manipulations.py
Berkant03: I will test it and transition to the heat communicator.

heat/core/manipulations.py
brownbaerchen: We just cannot rely on the world communicator because we don't know if the communicator of `a` is the world communicator. Maybe there is a way to access the unwrapped communicator? The line you mention here, I would anyways cut, so this specific problem is not really an issue.

Berkant03: I changed it up now to use the heat communicator.

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,204 @@ def sort(
...
+        local_data = a.larray.transpose(axis, 0)
+        shape = local_data.shape
+
+        local_data = local_data.reshape(shape[0], -1)
+        indices = torch.arange(0, local_data.shape[0])
+        for i in range(local_data.shape[-1] - 1, -1, -1):
+            indices = _permute_indices(local_data[:, i], indices)
+
+        local_data = local_data.reshape(shape)[indices].transpose(axis, 0)
+        return factories.array(local_data, split=None)
```
brownbaerchen, line 2983:
```suggestion
        if return_sort_indices_instead:
            return indices
        else:
            local_data = local_data.reshape(shape)[indices].transpose(axis, 0)
            return factories.array(local_data, split=None)

```

brownbaerchen (changes requested):

Berkant03:
There was the Idea to also implement a `ht.lexsort` using `vectorized_sort` to mimic similar behaviour as to numpy's `np.lexsort`. Unfortunetely, this is not possible for arrays with more than two dimensions, since numpy does not reduce the keys a two dimensional array like `vectorized_sort`.

Berkant03:
The other Idea was to use `vectorized_sort` for sorting complex types, which also would not work for arrays with more than one dimensions, since the sorting should happen on the specified axis, while ignoring other axis, which is not possible using `vectorized_sort`. One could use it for one dimensional data, but performance might suffer.

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,207 @@ def sort(
...
+    send_buf = torch.tensor([local_count], dtype=torch.int64)
+    local_counts = torch.empty(size, dtype=torch.int64)
+    comm.Gather(send_buf, local_counts, root=0)
+
+    if rank == 0:
+        send_counts = local_counts.numpy()
+        send_displ = np.insert(np.cumsum(send_counts)[:-1], 0, 0)
+
+        buffer = torch.empty((total_rows,), dtype=local_data.dtype)
+        recv_args = buffer, send_counts, send_displ  # , mpi_type]
```
```suggestion
        recv_args = (buffer, send_counts, send_displ)
```

heat/core/manipulations.py
```diff
@@ -2899,6 +2900,207 @@ def sort(
...
+    else:
+        buffer = None
+        recv_args = torch.empty(0, dtype=torch.int64), None, None
```
```suggestion
        recv_args = (torch.empty(0, dtype=torch.int64), None, None)
```
I find the brackets slightly more readably, although that is up for debate, I guess.

brownbaerchen (changes requested):
tiny aesthetic nitpicks

brownbaerchen (approved):

brownbaerchen (approved):

PR #2445, opened 2026-08-07
add dlpack support

### LINKED ISSUE #2436: support DLPack
DLPack itself doesn't support distributed arrays.

See also:
https://data-apis.org/array-api/latest/API_specification/generated/array_api.from_dlpack.html#from-dlpack
https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html#dlpack
https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack_device__.html#array_api.array.__dlpack_device__

mtar:
## Due Diligence

- General:
- Implementation:

## Description

This PR adds support for DLPack for undistributed arrays by using the functionality of the underlying torch tensor.

Issue/s resolved: #2436 

## Changes proposed:

- add `factories.from_dlpack`
- add `DNDarray.__dlpack__`
- add `DNDarray.__dlpack_device__`

## Type of change

array api

yes / no


heat/core/dndarray.py
```diff
@@ -829,6 +830,47 @@ def create_partition_interface(self):
...
+        max_version: tuple[int, int], optional
+            the maximum DLPack version that the consumer supports, in the form of a 2-tuple ``(major, minor)``.
+        dl_device: tuple[enum.Enum, int], optional
+            the DLPack device type. If ``None``, use same device as self. Default: None
+        copy: bool, optional
+            If ``True``, the input is copied. If ``False``, the input will not be copied, and a ``BufferError`` is
+            raised in the case a copy would be necessary. If ``None``, the existing memory buffer is used if possible,
+            and is copied otherwise.
+        """
+        if self.is_distributed():
```
brownbaerchen, line 861: Would it make sense to `resplit(None)` and raise a warning instead of an error here?

tests/core/test_factories.py
```diff
@@ -599,6 +599,22 @@ def get_offset(tensor_array):
         self.assertEqual(eye.shape, shape)
         self.assertEqual(eye.split, 1)
 
+    def test_from_dlpack(self):
+        a = ht.ones([4,4])
+        b = ht.from_dlpack(a)
```
brownbaerchen: I think it would make sense to test if this works with Numpy array and torch tensor as well, since that is the actual use case.

mtar: The DLPack support is completely handled by PyTorch. Basically, we are testing it with torch tensors :smiley:

brownbaerchen: Yes, but it is somewhat implicit.

heat/core/dndarray.py
mtar: Not sure if we should do it implicitly. I think the user should be responsible for the distribution. 
We can make the docstring clearer that it only works for undistributed arrays as DLPack doesn't support them.

tests/core/test_factories.py
brownbaerchen: Maybe the tests could be streamlined by doing
```python
a_ht = ht.ones([4,4])
for a in [a_ht, a_ht.numpy(), a_ht.larray]:
    self.assertIsInstance(b, ht.DNDarray)
    self.assertEqual(b.dtype, ht.core.types.canonical_heat_type(a.dtype))
    self.assertEqual(b.shape, a.shape)
    self.assertIsNone(b.split)

    if isinstance(a, np.ndarray):
        self.assertTrue(np.all(np.equal(b.larray.numpy(), a)))
        self.assertEqual(b.device, ht.cpu)
    elif ...
```

heat/core/dndarray.py
```diff
@@ -829,6 +830,52 @@ def create_partition_interface(self):
...
+        Raises
+        ------
+        BufferError
+            if the DNDarray is distributed as it is not supported by DLPack.
+        """
+        if self.is_distributed():
+            raise BufferError("DLPack export works for undistributed arrays only.")
+
+        return self.larray.__dlpack__(
+            stream=stream, max_version=max_version, dl_device=dl_device, copy=copy
```
brownbaerchen:
```suggestion
            stream=stream, dl_device=dl_device, copy=copy
```
The tests say torch doesn't accept `max_version`.

heat/core/dndarray.py
```diff
@@ -829,6 +830,52 @@ def create_partition_interface(self):
 
         return partition_dict
 
+    def __dlpack__(
+        self,
+        *,
+        stream: int | Any | None = None,
+        max_version: tuple[int, int] | None = None,
```
brownbaerchen:
```suggestion
```
See below

heat/core/dndarray.py
```diff
@@ -829,6 +830,52 @@ def create_partition_interface(self):
...
+
+        Parameters
+        ----------
+        self: DNDarray
+            array instance.
+        stream : int or Any, optional
+            For CUDA and ROCm, a Python integer representing a pointer to a stream,
+            on devices that support streams.
+        max_version: tuple[int, int], optional
+            the maximum DLPack version that the consumer supports, in the form of a 2-tuple ``(major, minor)``.
```
brownbaerchen:
```suggestion
```
see below

brownbaerchen (changes requested):
I'd be happy with this, except the tests fail :D
I marked what torch complains about.

mtar:
This is for the older version using an older standard. Suprisingly, with the latest pytorch version, the test fail for percentile on 4 processes :worried: dlpack gets called somehow. EDIT: fixed

heat/core/dndarray.py
```diff
@@ -829,6 +830,55 @@ def create_partition_interface(self):
...
+        Raises
+        ------
+        BufferError
+            if the DNDarray is distributed as it is not supported by DLPack.
+        """
+        if self.is_distributed():
+            raise BufferError("DLPack export works for undistributed arrays only.")
+
+        if torch.__version__ < "2.13.0":
+            return self.larray.__dlpack__(stream=stream, copy=copy)
```
brownbaerchen: Why don't you just use `**kwargs` and pass them on to torch? This dunder function is not very user-directed anyways, right?

heat/core/factories.py
```diff
@@ -878,6 +879,25 @@ def __factory_like(
...
+    Parameters
+    ----------
+    x : object
+        Input (array) object.
+    device: Device, optional
+        device on which to place the created array. Default: None.
+    copy: bool, optional
+        boolean indicating whether or not to copy the input. Default: None.
+    """
+    return array(torch.from_dlpack(x))
```
brownbaerchen: Arguments `Device` and `copy` are ignored here.

brownbaerchen (approved):

PR #2446, opened 2026-08-10
Refactor `__dlpack__`

brownbaerchen:
This worked on my laptop just now. We'll see if the entire pipeline goes through. If not, just close this PR, and carry on, @mtar.

### DIFF
--- a/heat/core/dndarray.py
+++ b/heat/core/dndarray.py
@@ -837,11 +837,16 @@ class DNDarray:
     ) -> Any:
         """
         Exports the undistributed array for consumption by ``from_dlpack()`` as a DLPack capsule.
+        Any positional arguments ``*args`` and keyword arguments ``**kwargs`` are directly forwarded to torch ``__dlpack__``.
+
+        Note
+        ----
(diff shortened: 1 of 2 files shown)


heat/core/dndarray.py
```diff
@@ -832,31 +832,12 @@ def create_partition_interface(self):
 
     def __dlpack__(
         self,
-        *,
-        stream: int | Any | None = None,
-        max_version: tuple[int, int] | None = None,
-        dl_device: tuple[Enum, int] | None = None,
-        copy: bool | None = None,
+        *args,
+        **kwargs,
```
mtar, line 836: The Array API standard defines the function signatures and the type annotations. We should adhere to them.

mtar (changes requested):

heat/core/dndarray.py
brownbaerchen: Well, this is adhering to the standard, assuming that torch adheres to it. The documentation just doesn't specify. Given that this method is not user facing, I think that's fine. We could also reference the torch function for valid arguments and say we only pass through here, if you want.

mtar: Yes, some more description would be good if we want to keep it this way.

mtar (approved):

PR #2452, opened 2026-08-10
add unique_values

### LINKED ISSUE #2451: implement unique_values
Move unique_values to the main namespace.

mtar:
## Due Diligence

- General:
- Implementation:

## Description

This PR moves unique_values to the main namespace. Also adds a very simple test as it solely calls unique with default arguments.

Issue/s resolved: #2451 

## Changes proposed:

- unique_values

## Type of change

array api

yes / no

### DIFF
--- a/tests/core/test_manipulations.py
+++ b/tests/core/test_manipulations.py
@@ -3731,6 +3731,11 @@ class TestManipulations(TestCase):
             (inv == ht.array(exp_inv.to(dtype=inv.larray.dtype), split=inv.split)).all()
         )
 
+    def test_unique_values(self):
+        # unique_values calls unique internally
+        array = ht.array([1, 1, 2])
+        self.assertTrue(ht.equal(ht.unique_values(array), ht.unique(array)))
+
     def test_vsplit(self):
         # for further testing, see test_split
         data_ht = ht.arange(24).reshape((4, 3, 2))
--- a/tests/core/test_manipulations.py
+++ b/tests/core/test_manipulations.py
@@ -3732,9 +3732,25 @@ class TestManipulations(TestCase):
         )
 
     def test_unique_values(self):
-        # unique_values calls unique internally
-        array = ht.array([1, 1, 2])
-        self.assertTrue(ht.equal(ht.unique_values(array), ht.unique(array)))
+        # 1D array
+        array_ht = ht.array([1, 1, 2])
+        array_np = np.array([1, 1, 2])
+
+        unique_ht = ht.unique_values(array_ht)
+        unique_np = np.unique_values(array_np)
+
+        self.assertEqual(unique_ht.dtype, array_ht.dtype)
+        self.assertTrue(np.all(np.equal(ht.sort(array_ht).numpy(), np.sort(array_np))))
+
+        # 2D array
+        array_ht = ht.array([[1., 1], [2, 3]], split=0)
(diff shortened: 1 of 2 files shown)


tests/core/test_manipulations.py
```diff
@@ -3731,6 +3731,11 @@ def test_unique(self):
             (inv == ht.array(exp_inv.to(dtype=inv.larray.dtype), split=inv.split)).all()
         )
 
+    def test_unique_values(self):
+        # unique_values calls unique internally
+        array = ht.array([1, 1, 2])
+        self.assertTrue(ht.equal(ht.unique_values(array), ht.unique(array)))
```
brownbaerchen:
```suggestion
        self.assertTrue(np.allclose(np.unique_values(array.numpy()), ht.unique(array).numpy()))
```
I think the comparison with the API we want to mirror makes this test more meaningful.

tests/core/test_manipulations.py
```diff
@@ -3731,6 +3731,11 @@ def test_unique(self):
             (inv == ht.array(exp_inv.to(dtype=inv.larray.dtype), split=inv.split)).all()
         )
 
+    def test_unique_values(self):
+        # unique_values calls unique internally
+        array = ht.array([1, 1, 2])
```
brownbaerchen: Maybe test for both 1d as well as 2d that the result matches with what we expect from numpy

brownbaerchen (changes requested):

tests/core/test_manipulations.py
```diff
@@ -3731,6 +3731,27 @@ def test_unique(self):
...
+
+        # 2D array
+        array_ht = ht.array([[1., 1], [2, 3]], split=0)
+        array_np = np.array([[1., 1], [2, 3]])
+
+        unique_ht = ht.unique_values(array_ht)
+        unique_np = np.unique_values(array_np)
+
+        self.assertEqual(unique_ht.dtype, array_ht.dtype)
+        self.assertTrue(np.all(np.equal(ht.sort(array_ht).numpy(), np.sort(array_np))))
```
brownbaerchen:
```suggestion
        for array in [ht.array([1, 1, 2]), ht.array([[1., 1], [2, 3]], split=0)]:
            unique_ht = ht.unique_values(array)
            unique_np = np.unique_values(array.numpy())
            
            self.assertEqual(unique_ht.dtype, array.dtype)
            self.assertEqual(unique_ht.device, array.device)
            self.assertTrue(np.allclose(unique_ht.numpy(), unique_np))
```
you forgot to actually check that the unique values match. You only checked if the arrays match before `unique_values`

tests/core/test_manipulations.py
```diff
@@ -3731,6 +3731,27 @@ def test_unique(self):
             (inv == ht.array(exp_inv.to(dtype=inv.larray.dtype), split=inv.split)).all()
         )
 
+    def test_unique_values(self):
+        # 1D array
```
brownbaerchen:
```suggestion
```
This should have been part of the below comment..

tests/core/test_manipulations.py
mtar: Oh :sweat_smile:, but the sorting was intended as the order of the elements is unspecified in unique_values

brownbaerchen (approved):

PR #2461, opened 2026-08-11
Refactor `unique_inverse`

brownbaerchen:
Puts definition of `UniqueInverseResult` into the `unique_inverse` function.


mtar (approved):
I don't see much a difference using the functional syntax, but as the NamedTuple is small and only used a single time, we can use it here.

brownbaerchen:
> I don't see much a difference using the functional syntax, but as the NamedTuple is small and only used a single time, we can use it here.

Yes, the code is doing exactly the same as what you wrote. I just preferred to move the definition of the result class into the function, so as not to clutter the surrounding scope.

PR #2482, opened 2026-08-24
Implementation for complex sorting of dndarrays and `take`

### LINKED ISSUE #2372: ht.sort should support complex input
**Related**
see https://github.com/helmholtz-analytics/heat/pull/2328#issuecomment-4647634276
#2327 #2328 

**Feature functionality**
`ht.sort` currently does not support complex input because torch.sort doesn't. We should a way to bypass this limitation.

**Additional context**

Tagging @GioPede as this is related to complex input support for `eigh`  (#2327)

Berkant03:
## Due Diligence

- General:
- Implementation:

## Description

Use `vectorized_sort` to implement a complex sort similar to `np.sort_complex`.

Issue/s resolved: #2372

## Changes proposed:

- Create a `ht.sort_complex` similar to `np.sort_complex`.
- Reuse row exchange fuctionality of `vectorized_sort`.

## Type of change

- `vectorized_sort`, extraction of the communication logic into a seperate function.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


brownbaerchen:
I marked the PR as draft because I couldn't find any actual implementation or tests for the new feature.

heat/core/manipulations.py
```diff
@@ -3038,70 +3180,124 @@ def _gather_column(flat_idx: int):
...
+    local_data = a.larray.transpose(axis, 0)
 
-        local_indices = recv_indices[mask] - offset
+    original_shape = local_data.shape
+    inner_shape = original_shape[1:]
 
-        send_counts[recv_rank] += mask.sum()
-        send_indices.append(local_indices)
+    total_rows = a.gshape[axis]
+    block_length = math.prod(inner_shape)
```
```suggestion
    block_length = np.prod(inner_shape)
```
No need to import math here.

heat/core/manipulations.py
```diff
@@ -6,6 +6,7 @@
 
 import numpy as np
 import torch
+import math
```
```suggestion
```
We can just use `np.prod` and skip `math` import.

heat/core/manipulations.py
```diff
@@ -2901,16 +2912,151 @@ def sort(
...
+    resplit_result: bool = True,
+    return_sort_indices: bool = False,
+) -> DNDarray | tuple[DNDarray, DNDarray]:
+    """
+    Stable complex sorting for DNDarrays.
+
+    Parameters
+    ----------
+    a : DNDarray
+        THe array to be sorted.
```
```suggestion
        The array to be sorted.
```

heat/core/manipulations.py
```diff
@@ -2901,16 +2912,151 @@ def sort(
...
+            axis=0,
+            descending=descending,
+            resplit_result=resplit_result,
+            return_sort_indices_instead=True,
+        )
+
+        res = reorder(a, idx.larray, resplit_result=resplit_result)
+        if return_sort_indices:
+            return res, resplit(idx, res.split)
+        return res
```
brownbaerchen, line 2992: Is this not a special case of the following code? What would be the impact if we removed this?

heat/core/manipulations.py
```diff
@@ -2901,16 +2912,151 @@ def sort(
...
+
+    larr_2d = larr.reshape(larr.shape[0], -1)
+
+    if return_sort_indices:
+        larr_2d_idxs = torch.empty(larr_2d.shape, dtype=torch.int64, device=a.device.torch_device)
+
+    for i in range(larr_2d.shape[1]):
+        col = torch.view_as_real(larr_2d[:, i])
+
+        temp = factories.array(col, split=None, device=a.device)
```
```suggestion
        temp = factories.array(col, split=None, device=a.device, comm=MPI.COMM_SELF)
```
Not sure if this is needed and I don't think this syntax works. `MPI` is probably not imported and `comm` probably needs to be Heat-wrapped. However, I am scared of deadlocks when not passing the self communicator..

heat/core/manipulations.py
```diff
@@ -3016,7 +3158,7 @@ def _permute_indices(data, idx):
         recv_args = (buffer, send_counts, send_displ)
     else:
         buffer = None
-        recv_args = (torch.empty(0, dtype=torch.int64), None, None)
+        recv_args = torch.empty(0, dtype=torch.int64), None, None
```
```suggestion
        recv_args = (torch.empty(0, dtype=torch.int64), None, None)
```

heat/core/manipulations.py
```diff
@@ -3038,70 +3180,124 @@ def _gather_column(flat_idx: int):
...
+
+def reorder(
+    a: DNDarray,
+    indices: torch.Tensor,
+    axis: int = -1,
+    resplit_result: bool = True,
+    original_split: int | None = None,
+) -> DNDarray:
+    """
+    Redistributes the dndarray along the specified axis using a global indice tensor.
```
```suggestion
    Redistributes the dndarray along the specified axis using a global index tensor.
```

heat/core/manipulations.py
```diff
@@ -3038,70 +3180,124 @@ def _gather_column(flat_idx: int):
...
+    a : DNDarray
+        The array whose slices along `axis` are to be rearranged.
+    indices : torch.Tensor
+        A 1D tensor of length `a.gshape[axis]` defining the new global order.
+    axis : int, optional
+        The axis along which to permute. Default is -1.
+    resplit_result : bool, optional
+        Whether to resplit the result back to the original split axis of `a`. Default is True.
+    original_split: int, optional
+        Overrides the split dimension gathered from the input `a` dndarray.
```
brownbaerchen: This parameter is confusing. Why would one override the original split?

heat/core/manipulations.py
```diff
@@ -3038,70 +3180,124 @@ def _gather_column(flat_idx: int):
...
+        raise ValueError(f"{axis=} does not exist for array with {a.ndim} dimensions.")
 
-    rank_slices = [comm.chunk((total_rows,), split=0, rank=i)[-1][0] for i in range(size)]
+    if axis < 0:
+        axis += a.ndim
 
-    local_slice = rank_slices[rank]
+    if not a.is_distributed():
+        local_data = a.larray.transpose(axis, 0)
+        local_data = local_data[indices].transpose(axis, 0)
```
```suggestion
        local_data = a.larray[(slice(None),) * (axis-1) + (indices,)]
```
should be faster

heat/core/manipulations.py
```diff
@@ -3038,70 +3180,124 @@ def _gather_column(flat_idx: int):
...
+        return factories.array(local_data, is_split=a.split)
 
-    assert all([s.step is None for s in rank_slices])  # Sanity check
+    if original_split is None:
+        original_split = a.split
```
brownbaerchen: I guess this controls the output split? So maybe rename to `out_split` or something like that?

heat/core/manipulations.py
```diff
@@ -3038,70 +3180,124 @@ def _gather_column(flat_idx: int):
...
+        send_indices_list.append(r_wants[mask] - local_start)
+
+    send_indices_tensor = torch.cat(send_indices_list)
+
+    needed_indices = indices[local_slice]
+    src_ranks = torch.bucketize(needed_indices, boundaries_tensor, right=True) - 1
+    recv_counts_tensor = torch.bincount(src_ranks, minlength=size)
+
+    send_counts = (send_counts_tensor * block_length).cpu().numpy()
+    recv_counts = (recv_counts_tensor * block_length).cpu().numpy()
```
```suggestion
    send_counts = (send_counts_tensor * block_length).numpy()
    recv_counts = (recv_counts_tensor * block_length).numpy()
```
that's redundant

tests/core/test_sorting.py
```diff
@@ -68,33 +68,100 @@ def test_vectorized_sort_multi_dim(self, orig_shape, split, axis, stable, descen
...
+
+                    comm.Bcast(permutation)
+
+                    yield shape, axis, permutation, split
+
+    @pytest.mark.parametrize("resplit_result", [False, True])
+    @pytest.mark.parametrize("shape, axis, permutation, split", list(_generate_reorder_params()))
+    def test_reorder(self, shape, axis, permutation, split, resplit_result):
+        a = ht.random.randn(*shape, split=split)
+        arr = torch.from_numpy(a.numpy())
```
brownbaerchen: Can't numpy do this?
```suggestion
        arr = a.numpy()
```

tests/core/test_sorting.py
```diff
@@ -68,33 +68,100 @@ def test_vectorized_sort_multi_dim(self, orig_shape, split, axis, stable, descen
...
+
+        res = ht.reorder(a, indices=permutation, axis=axis, resplit_result=resplit_result)
+        exp_res = arr.transpose(0, axis)[permutation].transpose(0, axis).numpy()
```
brownbaerchen:
```suggestion
        exp_res = arr[(slice(None),) * (axis-1) + (permutation,)]
```

brownbaerchen (changes requested):
I have a few minor comments, but look's good overall. Thanks!

tests/core/test_sorting.py
Berkant03: numpy's transpose works differently to pytorch's. Pytorch swaps 2 axis, while in numpy you define a tuple of the new order of all axis.

Berkant03: but given your suggestion for the faster variant of the double transpose we can use numpy only there.

heat/core/manipulations.py
Berkant03: Yes. It might probably work by removing it. It is from when I looked at individual axis and dimensions before generalizing. Will test it out

heat/core/manipulations.py
Berkant03: Heat has a MPI_SELF. I changed it up.

heat/core/manipulations.py
Berkant03: Interestingly, the tests fail when removing this part. Will look into this more.

Berkant03: Removing it would not work, since the 1d array has no split axis other than 0, which would not allow to find an orthogonal axis.

heat/core/manipulations.py
```diff
@@ -2901,16 +2911,151 @@ def sort(
...
+            resplit_result=resplit_result,
+            return_sort_indices_instead=True,
+        )
+
+        res = reorder(a, idx.larray, resplit_result=resplit_result)
+        if return_sort_indices:
+            return res, resplit(idx, res.split)
+        return res
+
+    if needs_resplit := (a.split == axis):
```
```suggestion
    # for nD arrays with n > 1, we must not split along the axis we sort along.
    needs_resplit = a.split == axis and a.is_distributed()
    if needs_resplit:
```
right? If `split is None`, we don't need to resplit.

heat/core/manipulations.py
Berkant03: In `vectorized_sort` a resplit of the data might happen, which the reorder function does not know about. and would thus try to resplit to the resplit split axis, and not to the original split axis of the input array. This of course could be mitigated by not overwriting the `a` input parameter. 

But it helps with preventing a second resplit. If we give `reorder` the original `a`, then it would resplit again, if it was resplit in `vectorized_sort`.

brownbaerchen: I still don't get it. Seems very much non-foolproof and therefore somewhat error prone.. Maybe it's just an issue of renaming things. Maybe a `prevent_resplit` or `skip_resplit` or whatever parameter could make this more comprehensible?

Berkant03: In `vectorized_sort`, if i give it a dndarray called `arr` with `split=0` and then let it sort across `axis=1`, it needs to do a `resplit` of `arr` onto `split=1` and saves this again onto `arr`. Now after the function found the correct indices, it will give the `arr` on `split=1` together with the indices into reorder, which sees that the split axis and the reorder axis match, therefore does no `resplit`. After the axis has been reordered, it checks if the input dndarray, in this case `arr` has been `resplit` (which is not the case here), if yes, the resulting array gets resplit onto the original split axis from `arr`, which in our case here is `split=1`, which matches our input split. Now the problem here is, if the user wants the original split axis for the resulting array of `vectorized_sort`, it would in this example be not the case, as the result is split across `axis=1` and not `axis=0`, which is why i added the `out_split` parameter, to correct this in `vectorized_sort`. 

Apart from the example, `vectorized_sort` also has the `resplit_result` parameter, which can prevent any resplit onto the input array split axis from happening on the resulting array. Saving communication, when reorder needs to do a resplit, due to `axis` and `split` mismatch, and the resulting array has a different `split`, as the final resplit gets ignored.

brownbaerchen: Ah, I see. I think this mixes concepts from both functions.

This `original_split` thing has nothing to do with `reorder`, it's entirely related to `vectorized_sort`, which may change the split in the first place and then maybe needs to revert this, depending on `resplit_results`, right?
Therefore, the second resplit back to the original split should happen in `vectorized_sort`, not in `reorder`.
Otherwise, `reorder` becomes coupled to `vectorized_sort`, making it difficult to understand/maintain because functions should be somewhat self-contained. Proof by example of me not understanding :D

If you want to read more, [pragmatic programmer](https://pragprog.com/titles/tpp20/the-pragmatic-programmer-20th-anniversary-edition/) is a 320 page version of "keep coupling of code parts to a minimum".

Berkant03: Changed it now to be seperate.

heat/core/manipulations.py
```diff
@@ -3038,70 +3181,119 @@ def _gather_column(flat_idx: int):
...
+        raise ValueError(f"{axis=} does not exist for array with {a.ndim} dimensions.")
 
-    local_slice = rank_slices[rank]
+    if axis < 0:
+        axis += a.ndim
 
-    assert all([s.step is None for s in rank_slices])  # Sanity check
+    if not a.is_distributed():
+        local_data = torch.index_select(a.larray, axis, indices)
+        return factories.array(local_data, is_split=a.split)
```
```suggestion
    if not a.is_distributed() or axis != a.split:
        local_data = torch.index_select(a.larray, axis, indices)
        return factories.array(local_data, is_split=a.split)
```
This may be a stupid question, but can't we just do this and not communicate anything if we reorder along a non-distributed axis?

Berkant03: Actually, it should work like this. I used this principle for the complex sort already, by giving always a non distributed array into vectorized sort and subsequent reorder.

brownbaerchen: Great, this should simplify `reorder` a good bit because you don't actually need to do any resplitting, right? This will also make `vectorized_sort` faster then.

Berkant03: Testing it, did not throw any errors. The logic was a "relic" from the first `vectorized_sort` implementation when it and `reorder` where still fused together.

brownbaerchen: This is a great example of when refactoring exposes a lot of implications and performance improvements :D

heat/core/manipulations.py
```diff
@@ -3038,70 +3182,119 @@ def _gather_column(flat_idx: int):
...
+    if res.split != original_split and resplit_result:
+        return resplit(res, original_split)
+    return res
+
+
+def reorder(
+    a: DNDarray,
+    indices: torch.Tensor,
+    axis: int = -1,
+    resplit_result: bool = True,
```
```suggestion
```
this is not longer, needed because `reorder` will never resplit.

heat/core/manipulations.py
```diff
@@ -3038,70 +3182,119 @@ def _gather_column(flat_idx: int):
...
+    Parameters
+    ----------
+    a : DNDarray
+        The array whose slices along `axis` are to be rearranged.
+    indices : torch.Tensor
+        A 1D tensor of length `a.gshape[axis]` defining the new global order.
+    axis : int, optional
+        The axis along which to permute. Default is -1.
+    resplit_result : bool, optional
+        Whether to resplit the result back to the original split axis of `a`. Default is True.
```
```suggestion
```

heat/core/manipulations.py
```diff
@@ -3038,70 +3182,119 @@ def _gather_column(flat_idx: int):
...
-    assert all([s.step is None for s in rank_slices])  # Sanity check
+    if not a.is_distributed() or axis != a.split:
+        local_data = torch.index_select(a.larray, axis, indices)
+        return factories.array(local_data, is_split=a.split)
 
-    send_counts = np.zeros(size, dtype=np.int64)
-    send_indices = []
+    original_split = a.split
+    if axis != a.split:
+        a = resplit(a, axis)
```
```suggestion
    assert axis == a.split  # any other cases should have been handled earlier
```

heat/core/manipulations.py
```diff
@@ -3038,70 +3182,119 @@ def _gather_column(flat_idx: int):
...
-    sorted_array = factories.array(recv_buf.transpose(0, axis), is_split=a.split, device=a.device)
+    reordered_array = factories.array(
+        recv_buf.transpose(0, axis), is_split=a.split, device=a.device
+    )
 
     if original_split != a.split and resplit_result:
-        return resplit(sorted_array, original_split)
-    return sorted_array
+        return resplit(reordered_array, original_split)
+
```
brownbaerchen: Hm, I can't suggest to remove this because line 3293 is not part of the diff. But I suggest to remove these lines :D

heat/core/manipulations.py
```diff
@@ -3038,70 +3182,119 @@ def _gather_column(flat_idx: int):
     if return_sort_indices_instead:
         return factories.array(indices, split=None, device=a.device)
 
-    offset, _, _ = comm.chunk((total_rows,), split=0, rank=rank)
+    res = reorder(original_a, indices, axis=axis, resplit_result=resplit_result)
```
```suggestion
    res = reorder(original_a, indices, axis=axis)
```

brownbaerchen (changes requested):
I pointed out a few places where outdated code needs to be removed, mostly as a note to myself...

heat/core/manipulations.py
```diff
@@ -2901,16 +2911,153 @@ def sort(
         return tensor
 
 
+def sort_complex(
+    a: DNDarray,
+    axis: int = -1,
+    descending: bool = False,
+    resplit_result: bool = True,
```
brownbaerchen, line 2918: This parameter is not properly implemented, right?

heat/core/manipulations.py
```diff
@@ -2901,16 +2911,153 @@ def sort(
...
+        larr.transpose(0, axis),
+        gshape=a.gshape,
+        dtype=a.dtype,
+        split=a.split,
+        device=a.device,
+        comm=a.comm,
+        balanced=a.balanced,
+    )
+
+    if needs_resplit:
```
```suggestion
    if needs_resplit and resplit_result:
```
Is this enough to implement `resplit_result` correctly?

brownbaerchen (changes requested):
`resplit_result` in `sort_complex` not implemented

heat/core/manipulations.py
Berkant03: Yes.

heat/core/manipulations.py
Berkant03: should be.

heat/core/manipulations.py
```diff
@@ -2901,16 +2911,153 @@ def sort(
...
+    sanitation.sanitize_in(a)
+
+    if not isinstance(axis, int):
+        raise ValueError(f"'axis' must be integer, not {type(axis)}.")
+    if not isinstance(descending, bool):
+        raise ValueError(f"'descending' must be bool, not {type(descending)}.")
+    if not isinstance(resplit_result, bool):
+        raise ValueError(f"'resplit_result' must be bool, not {type(resplit_result)}.")
+    if not isinstance(return_sort_indices, bool):
+        raise ValueError(f"'return_indices_instead' must be bool, not {type(return_sort_indices)}.")
```
```suggestion
        raise ValueError(f"'return_sort_indices' must be bool, not {type(return_sort_indices)}.")
```

heat/core/manipulations.py
```diff
@@ -2901,16 +2911,153 @@ def sort(
...
+    if axis < 0:
+        axis += a.ndim
+
+    if a.ndim == 1:
+        view = torch.view_as_real(a.larray)
+        shape = a.gshape + (2,)
+        temp = DNDarray(
+            view,
+            gshape=shape,
+            dtype=a.dtype,
```
```suggestion
            dtype=types.canonical_heat_type(view.dtype),
```
the dtype is not complex anymore

heat/core/manipulations.py
```diff
@@ -2901,16 +2911,153 @@ def sort(
...
+
+        idx = vectorized_sort(
+            temp,
+            axis=0,
+            descending=descending,
+            resplit_result=resplit_result,
+            return_sort_indices_instead=True,
+        )
+
+        res = reorder(a, idx.larray)
```
brownbaerchen: Wait: `a` can be split along 0 and `idx.larray` would then be only local and the dimension would be too small, no? Don't we need the global index array?

heat/core/manipulations.py
```diff
@@ -3038,70 +3182,110 @@ def _gather_column(flat_idx: int):
...
+        The reordered array.
+    """
+    sanitation.sanitize_in(a)
 
-    for recv_rank, s in enumerate(rank_slices):
-        recv_indices = indices[s]
+    if not isinstance(axis, int):
+        raise ValueError(f"'axis' must be integer, not {type(axis)}.")
+    if not (-a.ndim <= axis < a.ndim):
+        raise ValueError(f"{axis=} does not exist for array with {a.ndim} dimensions.")
```
```suggestion
    if not (-a.ndim <= axis < a.ndim):
        raise ValueError(f"{axis=} does not exist for array with {a.ndim} dimensions.")
    if not indices.shape == (a.gshape[axis],):
        raise ValueError(f'indices of shape {indices.shape} cannot be used to index array of shape {a.shape} along axis {axis}')
```
best to assert this.

heat/core/manipulations.py
```diff
@@ -2901,16 +2911,153 @@ def sort(
...
+            return res, resplit(idx, res.split)
+        return res
+
+    # for nD arrays with n > 1, we must not split along the axis we sort along.
+    needs_resplit = a.split == axis and a.is_distributed()
+    if needs_resplit:
+        orthogonal_axis = (axis + 1) % a.ndim
+        a = resplit(a, orthogonal_axis)
+
+    larr = a.larray.transpose(axis, 0).clone()
```
```suggestion
    larr = a.larray.transpose(axis, 0)
```
is this needed?

heat/core/manipulations.py
```diff
@@ -3038,70 +3182,110 @@ def _gather_column(flat_idx: int):
...
+    if not (-a.ndim <= axis < a.ndim):
+        raise ValueError(f"{axis=} does not exist for array with {a.ndim} dimensions.")
 
-        mask = (recv_indices >= offset) & (recv_indices < rank_slices[rank].stop)
+    if axis < 0:
+        axis += a.ndim
 
-        local_indices = recv_indices[mask] - offset
+    if not a.is_distributed() or axis != a.split:
+        local_data = torch.index_select(a.larray, axis, indices)
```
brownbaerchen, line 3238: would it make sense to call this function `index_select` as well? Does it do the same as this torch function? In that case, we should not assert that the length of the index array is the same as the length of the axis..
Is there maybe even an array API function for this?

heat/core/manipulations.py
```diff
@@ -3038,70 +3182,110 @@ def _gather_column(flat_idx: int):
...
+        return factories.array(local_data, is_split=a.split)
```
```suggestion
        return factories.array(local_data, is_split=a.split, comm=a.comm, device=a.device)
```

tests/core/test_sorting.py
```diff
@@ -68,33 +68,100 @@ def test_vectorized_sort_multi_dim(self, orig_shape, split, axis, stable, descen
...
 
-        res = ht.vectorized_sort(a, axis=axis, stable=stable, descending=descending)
-        res_idxs = ht.vectorized_sort(a, axis=axis, stable=stable, descending=descending, return_sort_indices_instead=True)
+        res = ht.vectorized_sort(a, axis=axis, descending=descending)
+        res_idxs = ht.vectorized_sort(a, axis=axis, descending=descending, return_sort_indices_instead=True)
 
         assert np.isclose(res.numpy(), expected_res).all()
         assert a.device == res.device
         assert np.equal(sort_idx, res_idxs.numpy()).all()
         assert a.device == res_idxs.device
```
brownbaerchen, line 78: Can we add tests that the split is what we expect? Since we are touching this above, and since we were missing something with the devices last time, I think this would be useful.

tests/core/test_sorting.py
```diff
@@ -68,33 +68,100 @@ def test_vectorized_sort_multi_dim(self, orig_shape, split, axis, stable, descen
...
+
+        res, res_idx = ht.sort(a, axis=axis, return_sort_indices=True, **kwargs)
+        exp_res = np.sort(arr, axis=axis, stable=True, **kwargs)
+        exp_res_idx = np.argsort(arr, axis=axis, stable=True, **kwargs)
+
+        assert np.isclose(res.numpy(), exp_res).all()
+        assert (res_idx.numpy() == exp_res_idx).all()
+
+        assert a.device == res.device
+        assert res.device == res_idx.device
```
brownbaerchen, line 147:
```suggestion
        assert res.device == res_idx.device
        assert res.split == a.split
```

brownbaerchen (changes requested):
A few more small things... Sorry about that :D

heat/core/manipulations.py
Berkant03: In a sense, `reorder` works like `index_select` that automatically balances the array too. We could call it that or we could also name it `take`, since `np.take` has the same functionality as `torch.index_select`

heat/core/manipulations.py
Berkant03: `larr_2d` gets modified in the for loop. The reshape to 2d, can be done without a copy, depending on the current layout, but most of the time, it is just some stride modification. Which here could mean that if the `transpose` does not do a copy, it would override the input array. The problem now is, that a transpose can copy but may also not, depending on the necessity, decided by pytorch. So to be sure, I added the clone.

brownbaerchen: Ah ok. I think it would be cleaner to add the clone in line 3001 to put it closer to where it should be, but I guess it's fine.

heat/core/manipulations.py
Berkant03: the idx heat array returned from `vectorized_sort` is always the global index array.

brownbaerchen: Yes, but if it is split, then `idx.larray` is not global. If we rely on the result being unsplit, we should assert this here, just to make sure we don't run into issues when someone changes `vectorized_sort`.

heat/core/manipulations.py
brownbaerchen: I like the idea of calling this `take`. That would help other people with finding this function, I think.

heat/core/manipulations.py
Berkant03: moved it to 3001

heat/core/manipulations.py
Berkant03: removed this, to allow functionality similar to `np.take`

brownbaerchen: [`np.take`](https://numpy.org/doc/2.4/reference/generated/numpy.take.html) allows for `axis=None`, which results in regular indexing. I am not sure how to handle this here. I don't think it's necessary to allow `ht.take(a, b)`, resulting in `a[b]`, but it may be a good idea to tell people this, especially since `axis=None` is the default in `np.take`

I would suggest to set `axis=None` the default here, too, and raise
```python
if axis is None:
    ValueError(f'{axis=} is not supported by `ht.take`, index the array directly instead.')
```

with how it is now, I am afraid that `import heat as np` could lead to some very difficult to debug problems.

what do you think?

Berkant03: We can do that, but we can also implement the axis=None case, since that is just a flattening and subsequent indexing.

brownbaerchen: Ah, ok. So `ht.take(a, b, axis=None)` is equal to `a.flatten()[b.flatten()].reshape(b.shape)`, yes? If so, sure, let's go for it.

Berkant03: its rather `a.flatten()[b]`.

brownbaerchen: Ah, interesting. Not sure if this relies on #938, though..

Berkant03: It should not, since I only used PyTorch Indexing.

brownbaerchen: But `a` is a DNDarray, right? so `DNDarray.__getitem__` will be called..

Berkant03: yes, but i am not calling a.flatten()[b] directly. Rather I just check first if the axis is None, if yes just do `a = a.flatten(); axis = 0` and let the rest be done normally by the function as before, which does not use any DNDarray indexing.

brownbaerchen: Ok yes, but does your implementation work if b is not flat?

Berkant03: no, it does not, which should work if we want to make it numpy like. But I also have found another bug, where in the non distributed case, i can have arbitrary flat index tensors, as long as the index is bound, but this is not the case for the distributed case. I will fix it up and try to find a solution for multi dim indexing.

tests/core/test_sorting.py
Berkant03: After going through the testing code again, do we really need separate split tests, when the split already gets tested in the existing functions (there are still some that do not do this, but I will add those)? Adding separate tests, would just add code duplication, since the cases would stay the same. Or do you have specifc cases in mind?

heat/core/manipulations.py
brownbaerchen: We can also just raise `NotImplementedError` for now. A simple rename of the function was suitable for the scope of this PR, but further implementation is maybe not. (Could go one layer up in the stack, though :D)

tests/core/test_sorting.py
brownbaerchen: Ah, I was somehow missing the below lines.
```python
assert res.split == a.split
assert res_idxs.split is None
```
is exactly what I meant.

Sorry about that!

heat/core/manipulations.py
Berkant03: I will add the error for now in regards to multi dimensional indices. Though the next person working on this can stack it even further :D.

brownbaerchen (approved):
Looks great! Thanks a lot, especially for your patience :D

brownbaerchen (changes requested):
Wait, the tests for `take` are failing :D

Berkant03:
Now it works again. At least everything except codebase which has not run through.

brownbaerchen (approved):

PR #2494, opened 2026-08-27
[Breaking Change] Deprecate `DNDarray.lloc`

### LINKED ISSUE #2492: Remove LocalIndex and DNDarray.lloc
The LocalIndex class and the property DNDarray.lloc defined in `dndarray.py` can be deleted. DNDarray.larray has superseded it.

brownbaerchen:
## Due Diligence

- General:
- Implementation:

## Description
`DNDarray.lloc` is used for local indexing. However, the function is confusingly named and indexing via `DNDarray.larray` is more intuitive, cleaner, and simpler.

Issue/s resolved: #2492

## Changes proposed:

- Remove `DNDarray.lloc`
- Remove `LocalIndex` class

## Type of change

Breaking change: Change API


heat/core/dndarray.py
```diff
@@ -263,40 +247,6 @@ def lnumel(self) -> int:
...
-        (2/2) tensor([0., 0., 0., 0.])
-        >>> a.lloc[1, 0:4] = torch.arange(1, 5)
-        >>> a
-        DNDarray([[0., 0., 0., 0., 0.],
-                  [1., 2., 3., 4., 0.],
-                  [0., 0., 0., 0., 0.],
-                  [1., 2., 3., 4., 0.]], dtype=ht.float32, device=cpu:0, split=0)
-        """
-        return LocalIndex(self.__array)
-
```
mtar, line 299: For the immediate release we could throw a deprecation warning here and in the docstring and delete the function afterwards.

mtar (dismissed):
Changes look fine. Suggestion for the unlikely case that somebody is still using that method.

heat/core/dndarray.py
brownbaerchen: Instead, I suggest to raise an error since we are doing an API changing release. But it's a good idea to tell people how to migrate to the new API in the error message.

mtar (approved):

PR #2495, opened 2026-08-27
Fix bug leading to mean and var being NaN if calculated distributed

maliesen:
## Due Diligence

- General:
    - [ X]  **title** of the PR is suitable to appear in the [Release Notes](https://github.com/helmholtz-analytics/heat/releases/latest)
- Implementation:

## Description

If the data is distributed so that the first two leading ranks are empty a divison by zero in `merge_moments` leads to mean and variance being calculated as `nan`. This can be circumvented by early returning what has been merged so far if there is no new actual data coming in.

Issue/s resolved: #

## Changes proposed:

-
-
-
-

## Type of change

yes / no


tests/core/test_statistics.py
```diff
@@ -1599,3 +1599,11 @@ def test_var(self):
 
         # edge case from #2374
         self.assertEqual(ht.var(ht.array([0.], split=None), axis=0, ddof=0), 0)
+
+    # meant to be run on four parallel processes (mpirun -n 4)
```
brownbaerchen:
```suggestion
@unittest.skipUnless(ht.communication.MPI_WORLD.size >= 3, "Test requires at least 3 tasks")
```
This comment can be turned into code. But also, we just need more than two tasks with the comments below.

heat/core/statistics.py
brownbaerchen: The type hints are wrong... might as well fix while we're at it.
```suggestion
    m1: Tuple, m2: Tuple, correction: bool = True
```

heat/core/statistics.py
```diff
@@ -1120,6 +1120,8 @@ def __merge_moments(
     if len(m1) != len(m2):
         raise ValueError(f"m1 and m2 must be same length, currently {len(m1)} and {len(m2)}")
     n1, n2 = m1[-1], m2[-1]
+    if (n2 == 0).all():
```
brownbaerchen:
```suggestion
    if n2.sum() == 0:
```
I think this would be more performant because this applies a reduce operation and then compares the resulting int versus comparing elementwise and then doing a reduce.

tests/core/test_statistics.py
```diff
@@ -1599,3 +1599,11 @@ def test_var(self):
...
+    def test_first_two_leading_ranks_empty(self):
+        data_all = ht.arange(100., split=0)
+        data = data_all[data_all >= 50.]
+
+        self.assertEqual(ht.mean(data), 74.5)
+        self.assertEqual(ht.var(data), 208.25)
```
brownbaerchen:
```suggestion
        self.assertEqual(ht.mean(data), np.mean(data.numpy()))
        self.assertEqual(ht.var(data), np.var(data.numpy()))

```

tests/core/test_statistics.py
```diff
@@ -1599,3 +1599,11 @@ def test_var(self):
 
         # edge case from #2374
         self.assertEqual(ht.var(ht.array([0.], split=None), axis=0, ddof=0), 0)
+
+    # meant to be run on four parallel processes (mpirun -n 4)
+    def test_first_two_leading_ranks_empty(self):
+        data_all = ht.arange(100., split=0)
+        data = data_all[data_all >= 50.]
```
brownbaerchen:
```suggestion
        comm = self.comm
        local_data = torch.tensor([], dtype=torch.float32) if comm.rank < 2 else torch.tensor([comm.rank], dtype=torch.float32)
        data = ht.DNDarray(local_data, gshape=(comm.size-2,), dtype=ht.float32, split=0, device=ht.devices.cpu, comm=comm, balanced=False)
```
This can be made more explicit.

brownbaerchen (changes requested):
Thanks a lot for spotting this! I have some minor comments but the overall fix is good.

maliesen:
Thanks for the review!

maliesen:
I put the pr in draft made as I realized the fix as is would introduce the following regression:

```
    @unittest.skipUnless(ht.communication.MPI_WORLD.size == 2, "Test for two tasks")
    def test_corrected_var_single_element(self):
        comm = self.comm
        ltensor_empty = torch.tensor([], dtype=torch.float32)
        ltensor_data = torch.tensor([0.], dtype=torch.float32)

        ldata_on_first_rank = ltensor_data if comm.rank == 0 else ltensor_empty
        data_on_first_rank = ht.DNDarray(ldata_on_first_rank, gshape=(2,), dtype=ht.float32, split=0, device=ht.devices.cpu, comm=comm, balanced=False)

        self.assertTrue(ht.isnan(ht.var(data_on_first_rank, ddof=1)))
```

This is because if the data (single element) is on the first rank, torch will give nan in https://github.com/helmholtz-analytics/heat/blob/main/heat/core/statistics.py#L2226 which will then be normalized to 0 in https://github.com/helmholtz-analytics/heat/blob/main/heat/core/statistics.py#L2228-L2229 so the `merge_moments` function is responsible for turning this in nan at https://github.com/helmholtz-analytics/heat/blob/main/heat/core/statistics.py#L1132 which would not happen anymore by the early return in this fix.

The distributed calculation of the corrected variance has a nan bug already before this fix (in the `__moment_w_axis` code branch) and I planned on fixing this as a separate pr but due to the regression I think it is better to do it at once. Therefore I merged the fix of this pr into https://github.com/helmholtz-analytics/heat/pull/2507 and if that pr is successful this pr may be closed.

brownbaerchen:
Closing as superseded by #2507.

PR #2506, opened 2026-08-28
[core] Implement DNDarray.__abs__

### LINKED ISSUE #2502: Implement DNDarray.__abs__()
**Related**
Array API

**Feature functionality**
dunder function for ht.abs()

[**Additional context**
Add any other context or screenshots about the feature request here.](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__abs__.html#array_api.array.__abs__)

mrprohack:
## Due Diligence
- General:
    - [x] **title** of the PR is suitable to appear in the Release Notes
- Implementation:
    - [x] unit tests: all applicable split configurations tested (`split=None`, `split=0`)
    - [x] unit tests: multiple dtypes tested (`int32`, `float32`)
    - [ ] **NEW** unit tests: MPS tested (1 MPI process, 1 GPU) — no MPS device is available on the development VPS; CI can cover hardware-specific checks
    - [x] benchmarks: not applicable; this exposes an existing operation through Python’s `abs()` protocol and adds no new kernel
    - [x] benchmarks: performance is unchanged; the dunder delegates to the existing `ht.abs()` implementation
    - [x] documentation updated where needed; `__abs__` reuses the existing `abs` docstring and adds no new public function

## Description

Add Python built-in `abs()` support for `DNDarray` by wiring `DNDarray.__abs__` to the existing `heat.core.rounding.abs` implementation.

Closes #2502

## Changes proposed:

- Bind `DNDarray.__abs__` to the existing absolute-value implementation.
- Reuse the existing `abs` docstring.
- Add regression coverage for `split=None` and `split=0`.
- Cover `int32` and `float32` and verify dtype/split metadata are preserved.

## Type of change

New feature (non-breaking change which adds Python protocol support for existing functionality).

## Testing

- TDD regression confirmed: without the binding, `abs(DNDarray)` fails with `TypeError`.
- `tests/core/test_rounding.py`: 10 tests passed with MPI process counts 1, 2, 3, and 4 on CPU.
- `pre-commit run --files heat/core/rounding.py tests/core/test_rounding.py`: passed.

## Memory requirements

No additional persistent memory requirements. The dunder delegates to the existing `ht.abs()` operation.

## Performance

No new numerical kernel or communication path is introduced. `abs(dndarray)` delegates directly to the existing absolute-value implementation.

No. It adds `abs(DNDarray)` support; existing `ht.abs()`, `DNDarray.abs()`, and `ht.absolute()` behavior is unchanged.

AI Support 🦾


tests/core/test_rounding.py
```diff
@@ -97,6 +97,18 @@ def test_abs(self):
...
+        for dtype in (ht.int32, ht.float32):
+            for split in (None, 0):
+                values = ht.array([-2, -1, 0, 1, 2], dtype=dtype, split=split)
+                result = abs(values)
+                expected_dtype = np.int32 if dtype == ht.int32 else np.float32
+                expected = np.array([2, 1, 0, 1, 2], dtype=expected_dtype)
+
+                self.assert_array_equal(result, expected)
+                self.assertEqual(result.dtype, dtype)
+                self.assertEqual(result.split, split)
```
brownbaerchen:
```suggestion
                values_numpy = values.numpy()
                expected = abs(values_numpy)

                self.assert_array_equal(result.numpy(), expected)
                self.assertEqual(result.dtype, dtype)
                self.assertEqual(result.split, split)
```

tests/core/test_rounding.py
```diff
@@ -97,6 +97,18 @@ def test_abs(self):
         self.assertEqual(float32_unsplit_absolute_values_fabs.sum(), 110.5)
         self.assertEqual(float32_unsplit_absolute_values_fabs.dtype, ht.float32)
 
+    def test_abs_dunder(self):
+        for dtype in (ht.int32, ht.float32):
+            for split in (None, 0):
+                values = ht.array([-2, -1, 0, 1, 2], dtype=dtype, split=split)
+                result = abs(values)
+                expected_dtype = np.int32 if dtype == ht.int32 else np.float32
```
brownbaerchen:
```suggestion
```
this is not used

heat/core/rounding.py
```diff
@@ -65,6 +65,9 @@ def abs(
 )
 DNDarray.abs.__doc__ = abs.__doc__
 
+DNDarray.__abs__: Callable[[DNDarray], DNDarray] = lambda self: abs(self)
```
brownbaerchen:
```suggestion
DNDarray.__abs__ = lambda self: abs(self)
```
`abs` doesn't take any further arguments. I don't think a type hint is helpful here.

brownbaerchen (changes requested):
Thanks! I have only a few minor comments.

tests/core/test_rounding.py
mrprohack: Updated to derive the expected values from NumPy as suggested. I kept `self.assert_array_equal(result, expected)` instead of converting `result` to NumPy because HeAT’s assertion helper expects a DNDarray as its first argument; `result.numpy()` caused an AttributeError on `.comm` during the focused test.

tests/core/test_rounding.py
mrprohack: Removed the unused `expected_dtype` helper.

heat/core/rounding.py
mrprohack: Removed the type hint. The binding is now `DNDarray.__abs__ = lambda self: abs(self)`.

brownbaerchen (approved):
Great, thanks!

PR #2507, opened 2026-08-30
Fix NaN bugs in the distributed calculation of mean and variance

maliesen:
## Due Diligence

- General:
- Implementation:

## Description

Issue/s resolved: There was a bug producing nan if the first two ranks are empty as described in https://github.com/helmholtz-analytics/heat/pull/2495. Also, there is bug leading to nan if there is one rank with only one single element if one calls `ht.var(..., axis=0, ddof=1)`.

## Changes proposed:

- Make `__merge_moments` not produce any `nan`s by not dividing by (n-ddof) and only divide by n if its safe (which is guaranteed by the early return if n2 == 0)
- Divide by (n-ddof) only once at the end of the calculation in the spirit of 
  [1] J. Bennett, R. Grout, P. Pebay, D. Roe, D. Thompson, Numerically stable, single-pass, parallel statistics algorithms, IEEE International Conference on Cluster Computing and Workshops, 2009, Oct 2009, New Orleans, LA, USA.

## Type of change

yes / no


maliesen:
`mpirun -n 1 pytest -vv` passes locally so this might be a CI pipeline failure (exit code 15). Maybe we can just retry the job?

heat/core/statistics.py
```diff
@@ -2222,27 +2221,26 @@ def reduce_vars_elementwise(output_shape_i: torch.Tensor) -> DNDarray:
             )
 
         else:  # case for full matrix calculation (axis is None)
+            n = x.lnumel
```
brownbaerchen, line 2224:
```suggestion
            n = x.larray.numel
```
I would like to deprecate the `lnumel` function at some point because I think it's cleaner to operate on the local array directly when doing anything local.

brownbaerchen (commented):
Looks really good, thanks!

The problem I have is that I am not an expert on this code and that the tests for var are a bit too confusing for me to just accept that surely everything works. That's not your fault, of course.
Before merging this, I would like to replace the tests that make sure the variance of ones is zero with tests that var on random data is the same between heat and numpy. In particular, with all this correction stuff which is very confusing.
If you have time to add that, that would be lovely. Otherwise, I can also do that in a separate PR and then we can merge this once all those tests pass on this branch.

tests/core/test_statistics.py
```diff
@@ -1,6 +1,9 @@
+import unittest
+
```
brownbaerchen, line 2:
```suggestion
```

maliesen:
Sorry for the late response! 🙈 got caught up. I noticed several problems with this pr varying in severity so i put it in draft mode. I should manage to come back to it tonight.

brownbaerchen:
> Sorry for the late response! 🙈 got caught up. I noticed several problems with this pr varying in severity so i put it in draft mode. I should manage to come back to it tonight.

No worries! I updated the tests, which apparently leads to merge conflicts.. But now I am a lot more confident in the tests :)

maliesen:
Good morning, I was quite busy lately so I still haven't found enough time yet but I'm still on it and hope to have updates shortly.

maliesen:
Out of Scope:

**1. `ht.mean` accumulates in float32, so the returned dtype depends on the number of processes.**
`reduce_means_elementwise` (`heat/core/statistics.py:987`, `:991`, `:992`, the `axis == split` path)
and the `axis=None` distributed branch (`:1032`, `:1033`) build their accumulators with
`factories.zeros(..., device=x.device)` and no `dtype=`, which defaults to float32. `var` passes
`dtype=x.dtype` in the corresponding places and is unaffected.

```python
x = ht.array(np.random.default_rng(0).random(120) + 1e3, dtype=ht.float64, split=0)
print(x.comm.size, ht.mean(x).dtype, float(ht.mean(x)))
```

```
1 float64 1000.5397506284788
2 float32 1000.5397338867188
3 float32 1000.5397338867188
```

The same happens for `ht.mean(x, axis=0)` with `split=0`. `n_tot` being float32 also caps exact
element counts at 2**24. The fix is `dtype=x.dtype` on those five calls, but it changes a returned
dtype, so it likely wants its own PR and a decision on whether that counts as a breaking change.

**2. `ht.var(x)` with `axis=None` downcasts to float32 when the array is *not* distributed.**
`heat/core/statistics.py:2272`: `ret = torch.var(x.larray.float(), correction=correction)`.
This mirrors 1: for float64 input the result is float32 on one process and float64 on two or more,
because the distributed branch uses `dtype=x.dtype`. Probably the same issue as 1, i.e. "the dtype
of a moment depends on the number of processes", and fixable in one go.

**3. Off-by-one bounds check in `__moment_w_axis`, currently unreachable.**
`heat/core/statistics.py:1397` reads `if any(d > len(x.shape) for d in axis)`, while the int branch
at `:1351` correctly uses `>=`. No caller can reach it today: `skew` and `kurtosis` reject tuple
axes outright at `:791` and `:1992`, and `mean`/`var`/`std` sanitize every entry before the call
(`ValueError: axis 3 is out of bounds for 3-dimensional array` for `axis=(2, 3)` on a 3-D array,
checked for all five functions). So this is a latent trap for whoever next routes a tuple through
`__moment_w_axis` rather than a live bug — a one-character cleanup.

maliesen:
Something else which came into my mind: Since we use random test data to verify the property that heat yields the same as numpy (the reference), I think we actually do property-based testing. Would it make sense to use a testing library like https://hypothesis.readthedocs.io/en/latest/ (which I don't have experience with)?

PR #2513, opened 2026-08-31
Remove array api namespace

### LINKED ISSUE #2420: remove array_api namespace
Delete the namespace. This should be done after all other issues have been completed.

mtar:
## Due Diligence

- General:
- Implementation:

## Description

This PR removes the Array API Namespace and moves the skips.txt and xfails.txt files for testing.

Issue/s resolved: #2420 

## Changes proposed:

- Delete array_api
- move skips.txt and xfails.txt to .github
- update namespace in test

## Type of change

Breaking Change

yes / no


brownbaerchen (dismissed):
Great, thanks! What's the merge process for this? Can you trigger the array API tests on this branch and post the result?

brownbaerchen (dismissed):

mtar:
I removed the skips.txt and xfails.txt files as we are still improving support for Array API at the moment and don't have to edit them every time we're adding a function.

brownbaerchen:
> I removed the skips.txt and xfails.txt files as we are still improving support for Array API at the moment and don't have to edit them every time we're adding a function.

That means the array api tests will fail, though, right? I think having a list for what needs to be added is a good idea. I guess xfail makes the most sense. Then the tests would pass if we didn't change compatibility and fail if we extended or restricted compatibility.

PR #2519, opened 2026-09-02
Raise error if GPU is selected for test via environment variable but no GPU is found

brownbaerchen:
In [this CI run on codebase](https://codebase.helmholtz.cloud/helmholtz-analytics/Cx/-/jobs/3390027), the AMD tests are failing because local variable `ht_device` is not found in [this line](https://github.com/helmholtz-analytics/heat/blob/9dfe62185474b2878d188d3beb6c2277dc535c1a/heat/testing/basic_test.py#L72) in the setup of the basic test class.
I think the problem is probably that the AMD runners no longer have GPUs. This PR doesn't fix that problem, but it raises an error if GPU is selected for the tests via `HEAT_TEST_USE_DEVICE=GPU pytest ...` but no GPU is found.

### DIFF
--- a/heat/testing/basic_test.py
+++ b/heat/testing/basic_test.py
@@ -67,7 +67,7 @@ class TestCase(unittest.TestCase):
                 is_mps = True
             else:
                 raise RuntimeError(
-                    f'Device {envar} from environment variable "HEAT_TEST_USE_DEVICE" not found! Neither CUDA, nor MPS is available'
+                    f'Device {envar} from environment variable "HEAT_TEST_USE_DEVICE" not found! Neither CUDA, ROCm, nor MPS is available'
                 )
         else:
             raise RuntimeError(
--- a/heat/testing/basic_test.py
+++ b/heat/testing/basic_test.py
@@ -47,12 +47,13 @@ class TestCase(unittest.TestCase):
         if envar == "cpu":
             ht.use_device("cpu")
             ht_device = ht.cpu
-            other_device = ht.cpu
             if torch.cuda.is_available():
                 torch.cuda.set_device(torch.device(ht.gpu.torch_device))
                 other_device = ht.gpu
             elif torch.backends.mps.is_built() and torch.backends.mps.is_available():
                 other_device = ht.gpu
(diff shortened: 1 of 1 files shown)


heat/testing/basic_test.py
```diff
@@ -64,6 +65,10 @@ def setUpClass(cls) -> None:
                 ht_device = ht.gpu
                 other_device = ht.cpu
                 is_mps = True
+            else:
+                raise RuntimeError(
+                    f'Device {envar} from environment variable "HEAT_TEST_USE_DEVICE" not found! Neither CUDA, nor MPS is available'
```
mtar:
```suggestion
                    f'Device {envar} from environment variable "HEAT_TEST_USE_DEVICE" not found! Neither CUDA, ROCm, nor MPS is available'
```

mtar (changes requested):
The runner has this issue once in a blue moon. If it doesn't get fixed, we have to report it again. Nevertheless, some better error message doesn't hurt.

mtar (approved):

brownbaerchen:
I sent an email to HIFIS support about the missing ROCm on AMD runners. Maybe they have an idea what changed. If not, we'll have to go hunting..

mtar:
The issue is a dependency version mismatch where pytorch for the wrong platform gets installed. This is fixed with one of the PRs:
- #2526 increases the dependency to the latest version
- #2523 sets the unused variable TORCH_PACKAGES to the previous version

brownbaerchen:
> The issue is a dependency version mismatch where pytorch for the wrong platform gets installed. This is fixed with one of the PRs:
> 
> * [Support PyTorch 2.14.0 #2526](https://github.com/helmholtz-analytics/heat/pull/2526) increases the dependency to the latest version
> * [add alias bitwise_inverse #2523](https://github.com/helmholtz-analytics/heat/pull/2523) sets the unused variable TORCH_PACKAGES to the previous version

The latest run in #2523 fails with the same error `local variable ht_device not found`... What do you suggest to do? Merge #2526 first?

PR #2520, opened 2026-09-02
Refactor tests for `ht.var`

brownbaerchen:
The tests for `ht.var` are very confusing. This PR cleans them up substantially by simply asserting that result is the same as numpy on random data with all sorts of examples. I suggest to merge this before #2507 in order to inspire some confidence.

### DIFF
--- a/tests/core/test_statistics.py
+++ b/tests/core/test_statistics.py
@@ -1532,23 +1532,25 @@ class TestStatistics(TestCase):
 
         size = self.comm.size
         shapes = [(2*size,), (2*size, 3*size), (2*size, 3*size, 4*size)]
+        correction_parameter_aliases = ['ddof', 'bessel']
         for shape in shapes:
             splits = [None,] + [i for i in range(len(shape))]
             for split in splits:
                 axes = [None,] + [i for i in range(len(shape))]
                 for axis in axes:
                     for correction in [0, 1]:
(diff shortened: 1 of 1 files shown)


tests/core/test_statistics.py
```diff
@@ -1534,68 +1530,25 @@ def test_var(self):
...
-                    if res.split is not None:
-                        if any([split >= x for x in lp_split]):
-                            self.assertEqual(res.split, len(target_dims) - 1)
-                        else:
-                            self.assertEqual(res.split, z.split)
-
-        # values for the iris dataset var measured by libreoffice calc
-        for sp in [None, 0, 1]:
-            iris = ht.load(get_dataset_path('iris.csv'), sep=";", split=sp)
-            self.assertTrue(ht.allclose(ht.var(iris, bessel=True), 3.90318519755147))
```
mtar, line 1598: The new tests does not test the `bessel` parameter.

mtar (changes requested):
Thanks. It is easier to read now. One optional parameter was lost during the refactoring.

mtar (approved):

brownbaerchen:
Codebase fails with a weird error. #2519 raises the proper error, but seems that AMD runners not having a GPU on codebase is no longer the exception..

PR #2471, opened 2026-08-14
skip large resplit test on GHActions

mtar:
## Due Diligence

- General:
    - [ ]  **title** of the PR is suitable to appear in the [Release Notes](https://github.com/helmholtz-analytics/heat/releases/latest)
- Implementation:
    - [ ] unit tests: all split configurations tested
    - [ ] unit tests: multiple dtypes tested
    - [ ] **NEW** unit tests: MPS tested (1 MPI process, 1 GPU)
    - [ ] benchmarks: created for new functionality
    - [ ] benchmarks: performance improved or maintained
    - [ ] documentation updated where needed

## Description

This PR skips a test on GitHub Actions that runs roughly an hour there.

Issue/s resolved: #2469 

## Changes proposed:

- skip test_resplit_large_count_limit on GitHub Actions

## Type of change

CI

#### Does this change modify the behaviour of other functions? If so, which?
yes / no

PR #2527, opened 2026-09-04
Tmp disable CB, remove concurrency limit for main CI, remove release drafter

JuanPedroGHM:
## CB

Due to the migration to bwcloud 2 to 3, CB is disabled until the new instance can be acquired and setup.

## Disabling concurrency limits for CI in main.

CI in main, both Github and Codebase keeps failing. On codebase AMD broke, on Github, it is unclear, because most of the errors come from the CI being interrupted due to a new commit. I think this feature should be disabled, otherwise we can miss some errors.

## Release draft

Removing this. It was ok while it lasted.


.github/workflows/push_main.yml
```diff
@@ -9,9 +9,9 @@ on:
       - 'tests/**'
       - 'pyproject.toml'
 
-concurrency:
-  group: main-ci
-  cancel-in-progress: true
+# concurrency:
+#   group: main-ci
+#   cancel-in-progress: true
```
brownbaerchen:
```suggestion
```
Let's delete. This was useful before @mtar skipped the tests for resplitting huge arrays in #2471 on GitHub. Now, I don't think we need this anymore.

brownbaerchen (commented):
I never really got the point of the release drafter. I don't feel comfortable with evaluating it, though. Maybe @ClaudiaComito  or @mtar can give their ok before we remove it.

mtar:
We haven't used the release drafter the last two releases because of the unreliable output. It is fine to remove it.

brownbaerchen (approved):

PR #2560, opened 2026-09-22
Limit concurrent CI workload to prevent Quick Tests failure

ClaudiaComito:
## Due Diligence

- General:
- Implementation:
    - [ ] unit tests: tests for correct device behavior (CPU, GPU, MPS)
    - [ ] benchmarks: created for new functionality or improved/maintained for existing functionality
    - [ ] non-general-Heat licensed material specified in REUSE.toml

## Description

I've  run a few tests to try and figure out what causes the super annoying `Error: Process completed with exit code 15.` in random Quick Tests. I found two causes:

1) CPU / threads oversubscription; 
2) interruption signal (for example because of a new push before the previous tests have finished); you can identify this one because in this case, the `docs/readthedocs` action will also fail with `interrupted by user`. 

## Changes proposed:

This PR modifies the GitHub Actions configuration to  limit concurrent CI workload:

  - added `MATRIX_MAX_PARALLEL` [repository variable](https://github.com/helmholtz-analytics/heat/settings/variables/actions) controlling the maximum number of matrix jobs running concurrently;
  -   added `max-parallel` variable to the test matrix jobs in the `PR_update` and `push_...` workflows, reading the repository variable `MATRIX_MAX_PARALLEL`;
  - limited threading in serial and MPI test runs (gh_tests/action.yml) by setting:
```bash
        OMP_NUM_THREADS=1
        MKL_NUM_THREADS=1
        OPENBLAS_NUM_THREADS=1
        NUMEXPR_NUM_THREADS=1
```      
- only for `pr_update`, a new push should no longer cancel the previous tests

Issue/s resolved: # every PR in the last ~6 months

-
-
-
-

yes / no


ClaudiaComito:
@Berkant03 @LeonKaem I tested this here but actually we will only know how much of a difference it makes when we merge it, as pr_update by default runs from `main`

Berkant03 (approved):

LeonKaem (approved):
In my limited understanding that looks plausible

ClaudiaComito:
Merging without codebase tests as nothing buy GH actions was changed.

